PR #3, opened 2015-10-06
Removed __EV3_MODULE_connected

antonvh:
I think we agree on these changes, at least

### DIFF
--- a/ev3dev/ev3dev.py
+++ b/ev3dev/ev3dev.py
@@ -493,10 +493,28 @@ class Motor(Device):
 
     time_sp = property( __get_time_sp, __set_time_sp, None, __doc_time_sp )
 
-    def run_timed(self,time_sp, duty_cycle_sp):
+    def run_timed(self, time_sp, *args, **kwargs):
+        """
+        Runs the motor for for a certain time in milliseconds at a certain power.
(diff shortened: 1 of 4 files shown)


antonvh:
Is this helpful? It works on my brickpi....

rhempel:
I will look at this tonight and maybe tomorrow - super busy the next couple of days

rhempel:
Anton, did you also make the change in the autogen templates to generate the changes in the bindings? I'm going to do the merge now but I'll also fix up the bindings, OK?

rhempel:
i just pushed a first draft of `autogen`ed `ev3dev.py` with support for helper functions that take kwargs. I still need to clean up how string vs int arguments are handled, but I'm pretty happy with how this worked out - comments on naming etc are welcome!

antonvh:
Nice. You can't have slashes in parameters, though. ("hold_pid/Ki") python thinks you want to divide it then.





I would also make two lists of properties: _int_properties and _str_properties so you can sort them out and cast them in the helper function.

rhempel:
Yes, I caught that - check out what I just pushed - it works pretty well now

rhempel:
I have merged with @ddemidov's more recent pull request - closing this one.

PR #4, opened 2015-10-13
Some cleanup

ddemidov:
This gets rid of trailing spaces in both source and generated files, formats docstrings as multiline python strings, and, most importantly, cleans up helper motor functions.

There is no need to create `_properties` dict and `_helper()` function.  `setarg(self, key, kwargs[key])` will access the correct property and deal with property type (string, int, etc) automatically.


rhempel:
Thanks @ddemidov - I like the changes. I'm going to remove the helper section that I added as a test to `spec.json` next

PR #6, opened 2015-10-14
Make Device constructor more generic 

ddemidov:
The previous version of `Device.__init__` took three positional arguments:
class name (`'tacho-motor'`), device name pattern (`'motor*'` or just `'*'`),
and port name (`'outA'`).

The new version takes two positional arguments (class name and device
name) and an arbitrary set of keyword arguments. The keyword arguments
are used for matching against device attributes. A match is considered a
match when the provided value is a substring of attribute value. If a
list is provided as keyword argument, then a match against any single
entry of the list is enough.

Compare the previous call:
```py
d = Device('tacho-motor', 'outA', 'motor*')
```
with the new calls
```py
# Get a motor on output port A:
d = Device('tacho-motor', 'motor*', port_name='outA')

# Get large motor on output port A:
m = Device('tacho-motor', port_name='outA', driver_name='lego-ev3-l-motor')

# Get an ultrasonic sensor (don't care if EV3 one or NXT one):
m = Device('lego-sensor', driver_name=['lego-ev3-us', 'lego-nxt-us'])
```
This would allow for easy implementation of such classes as `LargeMotor`
or `MediumMotor`.  This also makes Device constructor interface compatible
with boostc version of python bindings.

### DIFF
--- a/ev3dev/__init__.py
+++ b/ev3dev/__init__.py
@@ -1,1 +1,2 @@
+__all__ = ['ev3dev']
 from ev3dev import *
--- a/ev3dev/ev3dev.py
+++ b/ev3dev/ev3dev.py
@@ -39,29 +39,48 @@ class Device(object):
 
     DEVICE_ROOT_PATH = '/sys/class'
 
-    def __init__(self, class_name, port='auto', name='*' ):
+    def __init__(self, class_name, name='*', **kwargs ):
         """Spin through the Linux sysfs class for the device type and find
-           the first unconnected device"""
+        a device that matches the provided name and attributes (if any).
 
-        self._classpath = os.path.abspath( Device.DEVICE_ROOT_PATH + '/' + class_name )
+        Parameters:
+            class_name: class name of the device, a subdirectory of /sys/class.
+                For example, 'tacho-motor'.
+            name: pattern that device name should match.
+                For example, 'sensor*' or 'motor*'. Default value: '*'.
+            keyword arguments: used for matching the corresponding device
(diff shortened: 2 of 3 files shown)


rhempel:
Can the default device name, eg `motor*` be generated from the `systemDeviceNameConvention` property in the spec? When I created the binding I was using the generic device name, the `port_name` parameter was needed for things like the LEDs that don't have arbitrary device names from the spec. How do we plan to instantiate LEDs (for example) using these methods?
Ah, I see it's the optional second parameter - OK, I'll merge this as well - and I like what you've done with the kwargs matching. All specified kwargs have to match, and if a specific arg is a list, then any match in the list counts just for that keyword. Very elegant!

ddemidov:
> Can the default device name, eg motor* be generated from the systemDeviceNameConvention property in the spec?

I thought about that. The propety has format `motor{0}`. By coincidence, we directly use it in python as `'motor{0}'.format('*')`. But I am not sure we can rely on this.

> Ah, I see it's the optional second parameter

Its not optional second parameter, its part of kwargs. So yes, we can still specify `port_name`. Moreover, now its also possible to implement sensor classes correctly, which require to pass `driver_name` as well. `i2c_sensor` even needs an `address`.

rhempel:
I was talking about the LED devices, for example, that do not have a `port_name` - they would use the second parameter as the filename, in the same way that `motor*` as the second parameter means "pick the first motor device you find".
I really do like the improvement in the call for device instantiation!

PR #7, opened 2015-10-15
Expose public Device interface

ddemidov:
The first commit removes the `__exit__` method from `Device` which has no use without the corresponding `__enter__` method anyway (see https://www.python.org/dev/peps/pep-0343/). If this was intended to be a destructor, it should be called `__del__`, but that is not needed either, since default destructor should delete the file cache and close the file handles automatically.

The second commit exposes methods of the `Device` class as public, and makes the names compatible with boostc version of python bindings. If you don't like the new method names, we can discuss it, but I think it should be possible for the user to use raw `Device` class (hence, the methods should definitely be public).

The commit also gets rid of leading double underscores in private member names. See https://www.python.org/dev/peps/pep-0008/#method-names-and-instance-variables.

### DIFF
--- a/ev3dev/ev3dev.py
+++ b/ev3dev/ev3dev.py
@@ -70,11 +70,11 @@ class Device(object):
                 self._path = os.path.abspath( classpath + '/' + file )
 
                 # See if requested attributes match:
-                if all([self.__matches(k, kwargs[k]) for k in kwargs]):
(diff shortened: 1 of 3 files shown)


ddemidov:
I also found a bug in my initial implementation of `Device._matches()`. ef8c019 fixes that.

rhempel:
Thanks!

PR #8, opened 2015-10-16
Use standard decorators to create properties

ddemidov:
This may be subjective, but I think the changes make the generated getters/setters easier to read, and the classes are less polluted with private methods/fields.

### DIFF
--- a/ev3dev/ev3dev.py
+++ b/ev3dev/ev3dev.py
@@ -173,20 +173,21 @@ class Motor(Device):
 #~autogen python_generic-get-set classes.motor>currentClass
 
 
-    def _set_command(self, value):
-        self.set_attr_string( 'command', value )
-
-    _doc_command = """
+    @property
+    def command(self):
+        """
         Sends a command to the motor controller. See `commands` for a list of
         possible values.
         """
+        raise Exception( "command is a write-only property!" )
 
-    command = property( None, _set_command, None, _doc_command )
+    @command.setter
+    def command(self, value):
+        self.set_attr_string( 'command', value )
 
-    def _get_commands(self):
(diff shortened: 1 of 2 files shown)


rhempel:
Thanks! I'll move this into the main `ev3dev-lang` repo tonight. Are you happy enough with what we (mostly you lately) have accomplished here? Are we OK to kill the `pure-python` directory, make this the official `python` directory, and then merge with the `master` branch in `ev3dev-lang`?

ddemidov:
We are still missing things like `MediumMotor`, `LargeMotor`. Also sensors are not initialized properly (any `Sensor`-based class will happily connect to the first available sensor). I'll make a PR with those in a couple of days.

ddemidov:
And there is no LCD class. I remember there was a comment by @dlech somewhere about how to implement LCD (or rather framebuffer) IO properly in python.

rhempel:
Sensors - I'll look forward to your ideas there. It's Friday here so I have some free time on the weekend. And if we're thinking about the LCD framebuffer, let's also keep in mind that anything we want to standardize as a cross-language API, should be specifiable in `spec.json`
I don't know if you have worked with Lua before, but they tend to follow a mechanisms, not policy rule. That means give users primitives and let the libraries be built on those primitives. We should be able to get by with some very basic framebuffer API funcitons and then build a separate library on top of that - and if I know you, you've already got the Python image library in mind for this :-)

ddemidov:
See #9 re sensors.

Re framebuffer: I think the only thing we can put into spec is the ability to read/write the framebuffer. And whatever image manipulation library would be used in each of the bindings would dictate the interface.

PR #9, opened 2015-10-16
Add LargeMotor and MediumMotor implementation

ddemidov:
This implements LargeMotor and MediumMotor classes. They simply initialize motor with the added `driver_name` constraint. I am planning to implement classes that are based on `Sensor` in the same way later on.

### DIFF
--- a/ev3dev/ev3dev.py
+++ b/ev3dev/ev3dev.py
@@ -1381,6 +1381,24 @@ class LightSensor(Sensor):
       }
 
 #~autogen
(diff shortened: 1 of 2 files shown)


ddemidov:
I've added commits that autogenerate Motor- and Sensor-derived classes. These rely on the latest changes to the specification (I've pushed those both to develop and pure-python branches of the main repo).

rhempel:
Thanks again!

Issue #10, opened 2015-10-16
Module organization


ddemidov:
One thing I don't like is how the module is organized. When I write
```py
from ev3dev import *
```
I expect to be able to write `d = Device()`. But now, since `ev3dev` module (the top folder) has the `ev3dev` submodule (the `ev3dev.py` file) I have to do `d = ev3dev.Device()`. Normally if I wanted that, I would just `import ev3dev` instead.

So I think we should keep just the `ev3dev.py` at the top level and get rid of the `ev3dev` folder with `__init__.py`.

rhempel:
I'm not so sure about that - we run into namespace pollution risks - besides, once the `ev3dev.Device` or `ev3dev.Motor` is instantiated, the prefix is no longer an issue

ddemidov:
When I care about namespace pollution, I do
```py
import ev3dev
m = ev3dev.Motor()
```
And when I just want convenience, I do
```py
from ev3dev import *
m = Motor()
```
I think this is standard practice for python.

rhempel:
Ah, now I understand - so just pull `ev3dev.py` and `__init__.py` up from the ev3dev folder. Does this affect things if we submit to any other repositories ? Sorry for my ignorance on this topic

ddemidov:
I am not sure what you mean by other repositories. Something like PyPI to make the bindings installable through `easy_install` or `pip`? I think that would be fine.

rhempel:
Yes, that kind of repo - when we install from there, what does the directory structure look like?

ddemidov:
When we install it, it resides somewhere in dist-packages folder (as single `ev3dev.py` file). I've added a `setup.py` in #11, you can try to install the package system-wide with `sudo python setup.py install`.

antonvh:
This document explains how the module should be structured (it was ok, now it isn't anymore). http://peterdowns.com/posts/first-time-with-pypi.html

If you want to use it in another project without updating pypi, you can use subtree . Then test, change, and cherry pick back into master.

`git filter-branch --prune-empty --subdirectory-filter ev3dev libonly`

antonvh:
The new structure also breaks autogen for me. - I pulled the latest ev3dev-lang repo in the pure-python branch.

ddemidov:
autogen should be fixed with https://github.com/ev3dev/ev3dev-lang/commit/07c14ef53bc2e6195df3f06db87f84594fccf9a8.

[...]

PR #11, opened 2015-10-17
Moving ev3dev.py out of ev3dev folder

ddemidov:
This makes the module flat-organized (see #10).
This also adds a `setup.py` file so that the module may be installed with easy_install or uploaded to pypi.

The package may be installed system-wide with
```
sudo python setup.py install
```
The name `python-ev3dev` is currently occupied on PyPI by boostc version of the python bindings, but I can delete or rename the package when we decide we are ready to completely replace it.


ddemidov:
@rhempel, if you decide to merge it, the `autogen/autogen-list.json` in the main repo would need to be updated as well.

rhempel:
Can we have multiple authors in the author line?
I have fixed the `autogen-list.json` in my branch - it takes advantage of the ability to locate the templates with the specific language binding.

rhempel:
Are we about ready to push this into the official `ev3dev-lang` branch?

ddemidov:
> Can we have multiple authors in the author line?

I think so. This is just a text field.

> Are we about ready to push this into the official ev3dev-lang branch?

This is not yet a direct replacement to boostc version, but that is still available in my repository. I think we can do this and let people start complaining.

rhempel:
I prefer to think of it as "helpful suggestions" :-) OK, I'll get around to merging with the main branch of `ev3dev-lang` later this morning.
Thanks for all of your help on this!

PR #14, opened 2015-10-18
Property values

antonvh:
Added useful property values in the autogen. What is the use of the _propval dictionaries? How are they used in python?

This pull request is weird, I only want to merge 1 commit.


rhempel:
Anton - can you please sync up with the current branch and then resubmit a pull request - we may have got our lines crossed on this one.

antonvh:
I will push it directly when I'm added as author. OK?

Still: what is the use of the _propval dictionary?

ddemidov:
> Still: what is the use of the _propval dictionary?

It does not exist anymore :). See #12

antonvh:
Lol. I had done exacty the same, except that I used uppercase for the constant names, I thought it was more PEP. My pull request crossed your commit.

ddemidov:
You are probably right about uppercase. Personally, I think uppercase names for these things would be ugly (`m.polarity = m.POLARITY_INVERTED`, ugh), but since the primary purpose for these is IDE autocompletion, uppercase could work better.

PR #16, opened 2015-10-19
Generate led color functions for the EV3 platform

ddemidov:
Use the updated specification to generate functions managing EV3 leds.


rhempel:
Thanks!

PR #17, opened 2015-10-20
Added button class

antonvh:
Please only merge commit where I add the button class

I don't understand why my liquid files are out of sync with yours. At some point git is beyond me. Wouldn't you add me as author?

### DIFF
--- a/ev3dev.py
+++ b/ev3dev.py
@@ -1889,3 +1889,69 @@ class LegoPort(Device):
 
 #~autogen
 
+#------------------------------------------------------------------------------
+# Read Ev3 button states
+# Credits for this go to @topikachu and his python-ev3 library
+
+class attach_ev3_keys(object):
+	"""
+	This is a decorator class, used to add properties to the Button class.
+	"""
+    def __init__(self, **kwargs):
+        self.kwargs = kwargs
+
+    def __call__(self, cls):
+        key_const = {}
+        for key_name, key_code in self.kwargs.items():
+            def attach_key(key_name, key_code):
+                def fget(self):
+                    buf = self.polling()
+                    return self.test_bit(key_code, buf)
+                return property(fget)
+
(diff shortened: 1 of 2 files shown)


ddemidov:
You can sync your repo with @rhempel's by doing this:
```
git checkout -b local # keep local changes in a local branch
git checkout master
git fetch origin
git reset --hard origin/master # this will hard reset your master branch to origin (i assume origin points to @rhempel's repo)
git cherry-pick <commit-from-local-branch> # pick commits from local branch you want to push to main repo
```

antonvh:
Thanks. Git is back in sync now. This is a nice and clean pull request now.

ddemidov:
> Wouldn't you add me as author?

I think when @rhempel said he would add us as authors, he meant [here](https://github.com/rhempel/ev3dev-lang-python/blob/c07900d37e542443d03d2b2d8a7bd603c15423b4/setup.py#L7). Anyway, doing pull requests has an advantage of mandatory code review, so I would stick with those even if we are given push permissions.

antonvh:
Sure. Let's be careful to agree on who does what so pull requests don't collide. 

The advantage of push rights is seeing what the others are doing faster, and the ability to rebase and keep the tree clean.

ddemidov:
Also, you can press `watch` button at the top of the repo page (and recieve email notifications when somebody pushes something or comments on an issue).

rhempel:
I will get to this over the weekend, I think - just installed pycharm and I now understand better what we want to do with hinting and so on.

rhempel:
I have fixed up the templates to remove whitespace, and cleaned up the driver name list template. For buttons, we need to do something similar to the LEDs  - which is to create a generic class for Button, and platform specific instances for each button.
Also, @topikachu's button class implementation does not cache the file descriptor for the `ioctl` calls - and I'd like the generic driver to be able to handle that.
I'm trying to get this fixed up this morning, but life might get in the way :-)

PR #21, opened 2015-10-27
Rename a couple of methods in Button

ddemidov:
The first commit moves some code from template into `ev3dev.py`. The code does not depend on `spec.json`, so I think there is no need for it to reside in the template.

The other two commits do some renames:

*  `check_buttons`  -> `buttons_pressed`
* `these` -> `check_buttons`

also, `which` property is removed, since its just an alias for `buttons_pressed`. I had to look at the implementation of `which` and `these` to understand what was the purpose of the methods, which is a bad sign.


rhempel:
I'll need to get better at API names - years of using Forth have made me a bit silly....thks for the pull request, will merge this evening

rhempel:
Thanks! Is there some benefit to making the name/value for each button into a tuple instead of a dict?

ddemidov:
You mean for the `_buttons` dict? We only use it for iteration, so there indeed could be some minor speedup. I am not sure how important that would be wrt to file io though.

rhempel:
I'm not as concerned with speed for a few items - if the code is clear enough the way it's written, then we can leave it

PR #22, opened 2015-10-27
Implement FbMem and LCD classes

ddemidov:
`FbMem` provides generic framebuffer info and also maps framebuffer memory. The implementation is adopted from https://github.com/LinkCareServices/cairotft/blob/master/cairotft/linuxfb.py
(also see @dlech's [comment](https://github.com/ev3dev/ev3dev/issues/404#issuecomment-144135010)).

`LCD` is a convenience class that wraps `FbMem` and provides drawing functions from the Python Imaging Library (PIL).

### DIFF
--- a/ev3dev.py
+++ b/ev3dev.py
@@ -2141,7 +2141,7 @@ class FbMem(object):
             offset=0
         )
 
-class LCD(FbMem):
+class Screen(FbMem):
     """
     A convenience wrapper for the FbMem class.
     Provides drawing functions from the python imaging library (PIL).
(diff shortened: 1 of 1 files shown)


ev3dev.py
```diff
@@ -1987,3 +1991,214 @@ def status(self):
...
+        """Map the framebuffer memory."""
+        return mmap.mmap(
+            fbfid,
+            fix_info.smem_len,
+            mmap.MAP_SHARED,
+            mmap.PROT_READ | mmap.PROT_WRITE,
+            offset=0
+        )
+
+class LCD(FbMem):
```
dlech: `Screen` or `Display` would be a more generic name.

ddemidov:
I've made some changes after @dlech's comments. @dlech, could you please try if the code now works on all of your devices? `/dev/fb0` is still hard-coded, but exporting `FRAMEBUFFER` should be enough.

dlech:
Yes, I will give it a try this evening.

ddemidov:
I used [this demo](https://github.com/ddemidov/ev3dev-lang-python/blob/master/demo/lcd.py) to test the code on my own ev3 (`LCD` classname has to be replaced with `Screen`).

rhempel:
Thanks!

ev3dev.py
```diff
@@ -27,12 +27,16 @@
...
+import os
 import os.path
 import fnmatch
 import numbers
 import platform
 import fcntl
 import array
+import mmap
+import ctypes
+from PIL import Image, ImageDraw
```
rhempel: This raises an error on my system - I don't have PIL by default - do we need to add this to the base image? Or does the latest image already have PIL?

dlech:
When python-ev3dev gets packaged up in a Debian package, these should be automatically installed as dependencies. But yes, it is not included in the current image release.

Issue #25, opened 2015-10-28
Color framebuffer modes


dlech:
Well, it looks like this is going to be a bit more complicated than I though. PIL doesn't support arbitrary bits-per-pixel. The valid modes are listed [here](http://effbot.org/imagingbook/concepts.htm). So, when getting the mode, it needs to be something like `self.var_info.bits_per_pixel == 1 and "1" or "RGB"` [here](https://github.com/rhempel/ev3dev-lang-python/blob/master/ev3dev.py#L2223).

However, the color devices we are using are 16-bit RGB565. PIL doesn't have a mode that corresponds to this. `"RGB"` is 24-bit. So, `self._img.tobytes` is only going to work for 1bpp. There are a number of solutions for the 16bpp though. [Stack Overflow has some suggestions](http://stackoverflow.com/questions/1119754/using-python-to-convert-color-formats) and there is also the [code used by the official PiStorms software](https://github.com/adafruit/Adafruit_Python_ILI9341/blob/54d52bde26cec29bb497b1b8a5a904c93119efe9/Adafruit_ILI9341/ILI9341.py#L107). Any of those should do.

So, something like:

    if self.var_info.bits_per_pixel == 1:
        self.mmap[:] = self._img.tobytes("raw", "1;IR")
    elif self.var_info.bits_per_pixel == 16:
        self.mmap[:] = self._img_to_rgb565_bytes()
    else:
        raise Exception("Not supported")

...for [here](https://github.com/rhempel/ev3dev-lang-python/blob/master/ev3dev.py#L2270).

PR #26, opened 2015-10-28
Get working on 16bpp RGB565 framebuffer device.

dlech:


dlech:
See #25

ev3dev.py
```diff
@@ -35,6 +35,7 @@
 import fcntl
 import array
 import mmap
+import struct
 import ctypes
 from PIL import Image, ImageDraw
 from struct import unpack
```
dlech: I see now that I should have added pack here.

ev3dev.py
```diff
@@ -2262,11 +2263,24 @@ def clear(self):
...
 
+    def _color565(self, r, g, b):
+	"""Convert red, green, blue components to a 16-bit 565 RGB value. Components
+	should be values 0 to 255.
+	"""
+	return (((r & 0xF8) << 8) | ((g & 0xFC) << 3) | (b >> 3))
+
+    def _img_to_rgb565_bytes(self):
+        pixels = [self._color565(r,g,b) for (r,g,b) in self._img.getdata()]
+        return struct.pack('H' * len(pixels), *pixels)
```
dlech: Don't know if this is the most efficient way to do this, but it works.

rhempel: Make it work, then make it fast!

ddemidov: The most efficient way would probably be using numpy, but I am not sure we want that as dependency. May be array would also work, I can look at it later.

ddemidov: Ok, it seems python array is just an efficient way to *store* data arrays; it does not provide any array manipulation functions. However, I measured performance of @dlech's list-based solution and PiStorms numpy-based one. numpy is about 64 times faster (on my desktop computer): http://nbviewer.ipython.org/gist/ddemidov/84053a30e443645fcbeb

rhempel:
Thanks!

PR #28, opened 2015-10-28
Implement RemoteControl class

ddemidov:
The implementation mimics `Button` class interface. It also lets the user define button event handlers:

```py
rc = RemoteControl()

# Check button state:
if rc.red_up: do_something()

# Event loop
def make_handler(button):
  def handler(state):
    print(button, state)
  return handler

def on_change(buttons):
  for b,s in buttons:
    print(b,s)

rc.on_red_up = make_handler('red_up')
rc.on_change = on_change

while True:
  rc.process()
```

Two points:

1. The event handling functionality could be easily added to the `Button` class as well.
2. The implementations of `Button` and `RemoteControl` are so similar I am thinking of introducing `ButtonBase` class.

### DIFF
--- a/ev3dev.py
+++ b/ev3dev.py
@@ -2255,10 +2255,10 @@ class Screen(FbMem):
         self._draw.rectangle(((0,0), self.shape), fill="white")
 
     def _color565(self, r, g, b):
-	"""Convert red, green, blue components to a 16-bit 565 RGB value. Components
-	should be values 0 to 255.
-	"""
-	return (((r & 0xF8) << 8) | ((g & 0xFC) << 3) | (b >> 3))
+        """Convert red, green, blue components to a 16-bit 565 RGB value. Components
+        should be values 0 to 255.
+        """
+        return (((r & 0xF8) << 8) | ((g & 0xFC) << 3) | (b >> 3))
 
     def _img_to_rgb565_bytes(self):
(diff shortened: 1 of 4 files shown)


ddemidov:
Also, we should probably move both the button list and the `_BUTTON_VALUES` code book into `spec.json`.

ddemidov:
Rebased on the latest master.

rhempel:
+1 for implementing a `ButtonBase` class - both the on-device and remote buttons will benefit. There are merge conflicts that prevent me from merging this automatically - I probably did something out of order so this will need to wait until this evening.

rhempel:
While you're at is - what about a generic cache mechanism (we can add aging or least used policies later)?

ddemidov:
I've just rebased from your latest master, so merge conflicts should be gone now. But wait with the merge anyway, I'll try to implement `ButtonBase`.

rhempel:
+1 for moving the RC sensor button specifics to `spec.json` - focus on getting the implementation for EV3 hardware specified and stable, then we'll see how the platform specific aspects of the problem unfold.

ddemidov:
Is remote controller ev3-specific? We only interface with it through InfraredSensor, which (I think) may be attached to any platform either supported now or planned to be supported.

ddemidov:
I think this is ready to merge (after rhempel/ev3dev-lang#1).

rhempel:
Thanks!

PR #30, opened 2015-10-29
Stop polluting global namespace

ddemidov:
As suggested by @ensonic in https://github.com/ensonic/ev3dev-lang-python/commit/c868289a19e85e969ed100073dcf33fe368767aa#commitcomment-14053913


rhempel:
Thanks @ddemidov and @ensonic!

PR #31, opened 2015-10-29
Implement generic File cache class.

ddemidov:
This is branched off of #28, so that should be merged first (when #28 is merged this PR should consist of just 00ed242, e4fd573, and 54a8c32).


rhempel:
Thanks again to @ddemidov for really polishing up the edges and adding great new features!

PR #32, opened 2015-10-30
Get rid of size parameter in FileCache.read

ddemidov:
See discussion in 91bfa71551d8a9f6649705fda68de904254ef745.

### DIFF
--- a/ev3dev.py
+++ b/ev3dev.py
@@ -75,20 +75,17 @@ class FileCache(object):
 
         return f
 
-    def read(self, path, size=-1):
+    def read(self, path):
         f = self.file_handle(path, 'r')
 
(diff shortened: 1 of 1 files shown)


ddemidov:
I've split the commit into smaller parts. You may want to pick the first three even if you decide to omit c9b6d35 for now.

rhempel:
Have you tested the code on the target yet :-)

ddemidov:
I've tested in on fake target. Why, do you get any failures?

ddemidov:
Ok, I've just tested it on EV3 and both attribute IO and keyboard work.

rhempel:
Thanks!

PR #33, opened 2015-10-30
Unit testing

ddemidov:
See ev3dev/ev3dev-lang#112.

This adds two fake nodes for MediumMotor and InfraredSensor and adds some tests for the classes.

@rhempel, in order to enable Travis-CI integration for the repo, you need to go to https://travis-ci.org/, login with your github account, go to account settings and flip the switch on the `rhempel/ev3dev-lang-python` repo there. This should preferably be done before merging this pull request.


ddemidov:
A better way would be to create fake sys/class in the ev3dev-lang repo, but I could not come up with a way to do that and have ability to just run `./tests/api_tests.py` both locally and on Travis-CI.

ddemidov:
One possible way to do this would be to move the fake /sys/class folder to a separate repo and add it as a submodule here.

rhempel:
Thanks!

Issue #15, opened 2015-10-19
Things to do to catch up with boostc


ddemidov:
- [x] Implement [`device_index`](https://github.com/ev3dev/ev3dev-lang/blob/develop/wrapper-specification.md#special-properties) property for Device class.
- [x] Implement platform-specific `Led` instances and color functions.
- [x] Implement `measured_amps` and `measured_volts` [properties](https://github.com/ev3dev/ev3dev-lang/blob/develop/wrapper-specification.md#special-properties-4) for `PowerSupply` class (the names in the generated doc seem to be messed up).
- [x] Implement `Button` class for ev3 buttons.
- [x] Implement `Sound` class.
- [x] Implement `LCD` class.
- [x] Implement `RemoteControl` class.
- [x] Implement `bin_data_format` and `bin_data` for `Sensor`.
- [x] ~~Implement [`steering`](https://github.com/ddemidov/ev3dev-lang-python/blob/3e1a70a686771b4bd4a92220c91075f6b163e223/ev3dev/__init__.py#L85-L88) function.~~

antonvh:
What would you like me to do? Where would you like help? Apart from the steering function I'd like to make a `MotorPair` class, so you can `m_pair.run_forever(direction=-50, speed_sp=75)`

ddemidov:
All of these are pretty straightforward to implement except for `Button` and `LCD` class. Both of these need low-level access to input/output devices, and both of them probably belong to EV3 platform (are there any buttons on RPI except for full-sized USB keyboard? Is there an LCD screen on RPI? Does it work through framebuffer?).

I was going to try and finish `Led` class later this week. Feel free to choose any other task. By the way, since you are running RPI, can you please show the contents of `/sys/class/leds` folder?

antonvh:
Will do. 

I'm still unclear on how exactly you want to divide the work.

rhempel:
I think @ddemidov is going to work on the LED class implementation we talked about. @antonvh, could you send him the contents of your rPi `/sys/class/led` directory please?
@antonvh - would you like to try to fill out the `Button` or `Sound` classes?
Coming from the `Lua` world, I am in favour of "mechisms, not policy" - what that means for us is implementing a minimal feature set for each class that allows users to build what they need on top of it.
[...]

PR #34, opened 2015-10-30
Implement Device.device_index

ddemidov:
See #15


rhempel:
Thanks!

ddemidov:
I've rebased this on master and added a test for device_index (and also for repeated attribute reads, while I was there).

PR #36, opened 2015-10-31
Implement Sound class

ddemidov:
See #15


rhempel:
Thanks - this is helpful - we may want to consider adding parameters to the `speak` function to control volume, speed, inter-word space, and possibly language?

I really appreciate the work you are doing on this binding !

ddemidov:
The espeak parameters I just dragged from C++ implementation. We can indeed provide second optional parameter for espeak arguments.

PR #41, opened 2015-10-31
Generate reference documentation with Sphinx

ddemidov:
This adds sphinx configuration, so that refence documentation could be generated automatically. I've registered my own fork on the [readthedocs](https://readthedocs.org/) site, so that you have an idea of how the generated doc would look like:

http://ev3dev-lang-python-1.readthedocs.org/en/latest/

Once this is merged, I'll register the main repo there. It may be setup to rebuild the documentation automatically after each commit pushed to github.


rhempel:
Would be nice if `commands` was a bullet list or at least one per line

rhempel:
Thanks!

PR #52, opened 2015-11-02
Clean up ev3dev/version.py leftovers

ddemidov:


dlech:
ev3dev/version.py can also be deleted from .gitignore

PR #56, opened 2015-11-03
Replace pil with Pillow in python package requirements

ddemidov:
Debian package `python-pil` does install Pillow, the alive fork of dead
PIL project. When, however, `pil` is specified as package dependency in
`setup.py`, and `python-ev3dev` is installed with either
`easy_install python-ev3dev` or `python setup.py install`,
an attempt is made to install the actual PIL from PyPI. This fails on my
host system, and I think is not desired behavior anyway.

Also see ev3dev/ev3dev-lang#113

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -12,6 +12,6 @@ setup(
     url='https://github.com/rhempel/ev3dev-lang-python',
     include_package_data=True,
     py_modules=['ev3dev'],
-    install_requires=['pil']
+    install_requires=['Pillow']
     )
 


ddemidov:
I hope this does not break debian packaging.

ddemidov:
By the way, should not `python-pil` be mentioned [here](https://github.com/rhempel/ev3dev-lang-python/blob/6132b16a1944eb062048842b513a314244c189d1/debian/control#L12)?

rhempel:
I think `${python:Depends}` gets filled in automagically by the Debian package builder

ddemidov:
Does it use `setup.py` for that? We may be in trouble then. @dlech, is it possible to override it?

rhempel:
Argh - on Debian `python-pil`  is Pillow - not sure of the impact on packaging - @dlech is probably asleep.

rhempel:
Well, let's try it and see what happens...

dlech:
I don't think debian packaging cares what is in setup.py. It probably just looks at the import statements in the .py files that actually get installed. Anyway, just tried a build and it works, so no worries.

Any yes, `${python:Depends}` magically turns into `python-pil`.

PR #58, opened 2015-11-03
Use default readthedocs theme

ddemidov:
Its better integrated with the site, its mobile friendly, etc.


ddemidov:
There is a slight issue with badge size in mobile view, but their own page has same problem: http://read-the-docs.readthedocs.org/en/latest/badges.html. I tried hardcoding image height  in pixels, it did not help.

ddemidov:
https://github.com/snide/sphinx_rtd_theme/issues/144

PR #61, opened 2015-11-04
Added BrickPi Powersupply

antonvh:
Is this helpful?

### DIFF
--- a/ev3dev/brickpi.py
+++ b/ev3dev/brickpi.py
@@ -27,6 +27,7 @@ An assortment of classes modeling specific features of the BrickPi.
 """
 
 from .core import *
+import smbus
 
 
 OUTPUT_A = 'ttyAMA0:outA'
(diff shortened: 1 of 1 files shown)


dlech:
It would be better to write a kernel driver so that this can be used by any programming language, not just python. See ev3dev/ev3dev#413.

This code is very helpful as a reference to do that.

dlech:
I have added a kernel driver partly based on this. I did the math and got 0.017725 which is close to your empirical results. Enabling the driver causes my Pi to crash since there is nothing there (actually more of an infinite loop than a crash) so you will have to test it out after the next kernel release.

rhempel:
I think we're going to skip this PR based on having a kernel driver to support it, yes?

@dlech - we get attributes for voltage and current with your driver?

dlech:
See <https://github.com/ev3dev/ev3dev/issues/412#issuecomment-154242776>. PiStorms is voltage only.

dlech:
And yes, we have an (semi-untested) kernel driver.

antonvh:
Will do! Thanks for adding this!

rhempel:
I will close this PR - when you have code that works with the new kernel driver, open a new PR please.

Issue #67, opened 2015-11-05
file handle cache breaks sensors


ensonic:
try this in python (with sensor 0 being the touch sensor):
```
f = open("/sys/class/lego-sensor/sensor0/value0", "r")
print f.readline()
0

f.seek(0)
print f.readline()
0
```

ensonic:
Found a fix. PR is comming in a sec.

dlech:
There is no guarantee that the value returned by the sysfs attribute includes a newline, it is just convention. I think instead of flushing, it might be more efficient to open with no buffering.

    f = open("/sys/class/lego-sensor/sensor0/value0", "r", 0)

These files can't be mmapped (except the `bin_data` attribute).

ddemidov:
I also wonder if replacing `readline` with `read` would also solve this. `readline`-based implementation was only introduced recently.

dlech:
I know from experience that you need to turn off buffering or read() does not work correctly.

ddemidov:
Good to know. I'll do a PR after some testing then.

ensonic:
The flush() is better than nothing, but I can try differently. @ddemidov no need for another PR I can change this one.

ddemidov:
@ensonic, ok, thanks!

ensonic:
PR updated using with with buffering disabled and using read()

rhempel:
Agreed, I merged the PR, if this fixes your issue, please close.

PR #68, opened 2015-11-05
Also flush on read() to not return cached data.

ensonic:
Fixes issue #67

### DIFF
--- a/ev3dev/core.py
+++ b/ev3dev/core.py
@@ -58,11 +58,11 @@ class FileCache(object):
         """Manages the file handle cache and opening the files in the correct mode"""
 
         if path not in self._cache:
-            f = open(path, mode)
+            f = open(path, mode, 0)
             self._cache[path] = f
         elif reopen:
             self._cache[path].close()
-            f = open(path, mode)
+            f = open(path, mode, 0)
             self._cache[path] = f
         else:
             f = self._cache[path]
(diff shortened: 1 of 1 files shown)


ddemidov:
Did you also try to open files without buffering and get read of f.flush(), as @dlech [suggested](https://github.com/rhempel/ev3dev-lang-python/issues/67#issuecomment-154111940)?

ddemidov:
Sorry, you did exactly that.

rhempel:
Thanks!

ddemidov:
This breakes python3 compatibility. The following error is raised:
```
  File "./tests/../ev3dev/core.py", line 61, in file_handle
    f = open(path, mode, 0)
ValueError: can't have unbuffered text I/O
```
We probably don't care about python3 compatibility, but its worth mentioning.

rhempel:
Can we fix this by opening in `b`inary mode? I know attributes are text only, but it never really hurts to open in binary mode.

ddemidov:
I tried just replacing moded with binary versions, and it did not work. The error is:
```
  File "./tests/api_tests.py", line 18, in test_device
    d = ev3dev.Device('tacho-motor', 'motor*', driver_name='lego-ev3-m-motor')
  File "./tests/../ev3dev/core.py", line 136, in __init__
    if all([self._matches(k, kwargs[k]) for k in kwargs]):
  File "./tests/../ev3dev/core.py", line 136, in <listcomp>
    if all([self._matches(k, kwargs[k]) for k in kwargs]):
  File "./tests/../ev3dev/core.py", line 159, in _matches
    return value.find(pattern) >= 0
TypeError: a bytes-like object is required, not 'str'
```
I guess we can fix this by explicitly converting strings to bytes and the other way around where appropriate, but I am not sure is this is the best way to go.

ensonic:
Yep, I think we need to do the conversion explicitly. I'll give it a try and send a new PR.

Issue #70, opened 2015-11-05
automatic platform detection should not be done in ev3dev package __init__


EricPobot:
The automatic detection based on the current architecture makes impossible to have the appropriate imports done when working on the dev system and targeting something else than an EV3. When developing an app for a BrickPi for instance, IDE code hints, completion and checking mechanisms will not work since the ev3 import will always be done.

The platform dependant import should be left to the responsibility of the robot code developer, who is supposed to know which platform he targets. Instead of using  `import ev3dev` in the application code modules, he will use something like `form ev3dev import ev3`. 

To have this mechanism work properly platform specific modules must include `from .core import *`, so that the generic definitions will be made available automatically to them. BTW doing this makes possible for platform specific modules to override generic definitions by re-defining them if needed.

dlech:
Also see ev3dev/ev3dev#412

ddemidov:
I agree the way its implemented now is not right. For some reason I assumed that `import * from ev3dev.brickpi` would not load `ev3dev.__init__`. I just checked and `__init__` does get loaded. I still would like for a way to autoload default platform. May be could move the code from `ev3dev.__init__` to `ev3dev.auto`.

I also like what @dlech is saying in ev3dev/ev3dev#412. Basically, we don't need a lot of platform-specific code, we just look at platform properties and expose what's available.

EricPobot:
Beware that specificities go beyond properties only. 

For instance, the current implementation of the LEDs for the EV3 is not satisfying, because they are dual colour ones (I'm modifying the ev3.py module to illustrate this, and the point will be more explicit then). Platform specificities will include behaviours, and thus logic. I'm not sure that a declarative mechanism for properties will be enough.

dwalton76:
@ddemidov should we just close this one?  It has been 2 years, if we haven't fixed it by now...

WasabiFan:
8c654d630727f1843ea11ee9671f28f605cb04be ?

ddemidov:
Yeah, it looks like it was resolved in #72.

PR #72, opened 2015-11-06
Move platform-guessing from __init__.py to auto.py

ddemidov:
See #70


rhempel:
Thanks

ensonic:
Instead of 'import ev3dev' one now needs to use 'from ev3dev import ev3 as ev3dev' unless one wants to change the code.

Issue #66, opened 2015-11-05
Sensor.commands throws IOError with EPERM instead of EOPNOTSUPP as said in the docs.


ensonic:
```
s = ev3dev.TouchSensor(ev3dev.INPUT_1)
s.commands
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "build/bdist.linux-armv5tejl/egg/ev3dev/core.py", line 1183, in commands
  File "build/bdist.linux-armv5tejl/egg/ev3dev/core.py", line 187, in get_attr_set
  File "build/bdist.linux-armv5tejl/egg/ev3dev/core.py", line 184, in get_attr_line
  File "build/bdist.linux-armv5tejl/egg/ev3dev/core.py", line 165, in _get_attribute
  File "build/bdist.linux-armv5tejl/egg/ev3dev/core.py", line 79, in read
  File "build/bdist.linux-armv5tejl/egg/ev3dev/core.py", line 65, in file_handle
IOError: [Errno 13] Permission denied: '/sys/class/lego-sensor/sensor1/commands'
```

While the docs
http://python-ev3dev.readthedocs.org/en/latest/#ev3dev.Sensor.commands
say 'EOPNOTSUPP, this seems to not hold for python.
```
cat /sys/class/lego-sensor/sensor1/commands
cat: /sys/class/lego-sensor/sensor1/commands: Operation not supported
```

ddemidov:
The problem is [this](https://github.com/rhempel/ev3dev-lang-python/blob/f9a9d2ee4d5b3107b2bf9ec4135c18072d530b3b/ev3dev/core.py#L60-L68) code in `FileCache.file_handle()` method:
```py
        if path not in self._cache:
            f = open(path, mode)
            self._cache[path] = f
        elif reopen:
            self._cache[path].close()
            f = open(path, mode)
            self._cache[path] = f
        else:
            f = self._cache[path]
```

When `commands` attribute of `TouchSensor` is opened for reading, an exception is thrown: `IOError: [Errno 95] Operation not supported`. The current code assumes an IOException was thrown because file needs to be reopened in read-write mode, and tries to do that. Then `IOError: [Errno 13] Permission denied` is thrown and reaches user code. I guess we only need to catch `not writable` exception, but I am not yet sure how to do this in python.

dlech:
[...]

PR #73, opened 2015-11-06
Use os.access() to open attribute files in correct mode

ddemidov:
Attribute files are opened in 'r', 'a', or 'a+' mode based on result of `os.access()` function call. 'a' and 'a+' are preferred to 'w' and 'w+' to not overwrite the files until actually asked by user.

This solves #66 for me in the sense that TouchSensor.commands now raises `IOError: [Errno 95] Operation not supported`. The rest of the bindings work as expected.

### DIFF
--- a/ev3dev/core.py
+++ b/ev3dev/core.py
@@ -54,14 +54,23 @@ class FileCache(object):
         for f in self._cache.values():
             f.close()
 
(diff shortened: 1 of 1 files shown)


rhempel:
Thks

PR #75, opened 2015-11-07
Fix doc generation

ddemidov:
Broken by #72


ddemidov:
This also wraps autosummary table contents (borrowed the solution from snide/sphinx_rtd_theme#117) and removes horizontal scrollbar from the table. It seems though that autosummary tables are currently broken on readthedocs anyway (See rtfd/readthedocs.org#1290). Should be fixed when they update sphinx (and works locally).

Issue #37, opened 2015-10-31
Transition from boostc to pure-python


ddemidov:
I think we are now feature complete w.r.t. boostc version (pinging @ensonic). We can push this to PyPI and advertise on [ev3dev.org](http://www.ev3dev.org/docs/libraries/).

Now, `python-ev3dev` package name on PyPI is taken by boostc version. The latest version published there was [0.2.2](https://pypi.python.org/pypi/python-ev3dev). I suggest we tag the pure-python bindings with something like 0.3 and just keep pushing to the same package on PyPI. This way users may still download older (boosts-based) versions. And I am going to put deprecation warning with a link to this repo in boostc version README.

ddemidov:
We also may want to wait for @EricPobot's changes before version tagging. I've seen some nice work done already in his fork.

ddemidov:
Also see #15

rhempel:
@ddemidov, I will follow your lead on this - the Python packaging is something you are much more familiar with.
I think I can start making this into a Debian package.

ddemidov:
@ensonic did some work on debian package for boosc version [here](https://github.com/ddemidov/ev3dev-lang-python/tree/master/debian). Can we use it?

rhempel:
Yes, but now we need to add `python-pil`, `beep`, `aplay`, and `espeak` to the dependencies, fortunately we're not making changes to upstream sources so there's no patches necessary.

dlech:
debhelper should take care of all of the python dependencies automatically. That is what `${python:Depends}` is for in the `debian/control` file.

ddemidov:
The 0.4.0 version of the bindings is uploaded to pypi: https://pypi.python.org/pypi?:action=display&name=python-ev3dev&version=0.4.0.

So the library may now be installed with `easy_install -U python-ev3dev`. We can put this into README, or we can wait until debian package is ready and use `apt-get install python-ev3dev`.

ericpascual:
@ddemidov I'm far from home currently, and thus has no access to my machine. However as you could notice, I've pushed the latest version of my fork. Since I'm not sure to be able to isssue a pull request very soon, feel free to pick what is ok for you already.

One of the main points of my version (demoed in the small `motors.py` demo) is that the user of the lib does not make an `import ev3dev`, but an `import <my_platform>`, which in turn includes an `from ev3dev import *`, thus exposing transparently the base layer in addition to the extensions or overrides specific to that platform.

[...]

PR #76, opened 2015-11-07
Switch to group-based interface in Leds

ddemidov:
This is based on discussion at https://github.com/rhempel/ev3dev-lang-python/issues/37#issuecomment-153442761.

This introduces concepts of led groups (see also ev3dev/ev3dev#412) and colors into `Leds` class. The led groups may be operated as in:
```py
# Set group color:
Leds.set_color(Leds.LEFT, Leds.AMBER)
Leds.set_color(Leds.RIGHT, (1.0, 0.75))

# Batch set arbitrary led properties in group:
Leds.set(Leds.LEFT, trigger='timer', brightness_pct=0.5)
```

I hope this is close to what @EricPobot had in mind.

This is accompanied by rhempel/ev3dev-lang@28361dad5e4f33749ee76634eeedfc1b972b6507

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -30,6 +30,7 @@ Module interface
     InfraredSensor
     RemoteControl
     Led
+    Leds
     PowerSupply
     Button
     Sound
@@ -107,6 +108,9 @@ Other
 ^^^^^
 
 .. autoclass:: Led
+    :members:
+
+.. autoclass:: Leds
     :members:
 
 .. autoclass:: PowerSupply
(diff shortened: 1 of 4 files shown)


ddemidov:
Also, I hope @ensonic is ok with this change.

ddemidov:
rhempel/ev3dev-lang#5, if/when merged into ev3dev/ev3dev-lang, will break js and C++ bindings, so pinging @WasabiFan as well.

rhempel:
Welcome to the world of hair splitting when it comes to bicoloured LED control. I can give one good reason for retaining individual control, and that is to multiplex more visual information on one LED - ie set one background color for system state, but still allow flashing for disk access. Maybe that's in the realm of overthinking it..

ddemidov:
`Leds` class exposes all the leds present in the system individually as well as in groups. I think that should be enough for any use case.

Issue #47, opened 2015-11-01
Demo scripts


ddemidov:
I think this repo needs some examples. We can adapt some from boostc, or create something new, but it would help new users of the bindings.

E.g . see http://ev3dev.org/irclog/#id-29877

EricPobot:
There is a basic one for motors in my fork. Feel free to pick it and adapt as you like.

ddemidov:
@laurensvalk kindly agreed to provide simple robot instructions from his book for use with ev3dev examples: https://github.com/ev3dev/ev3dev/issues/329#issuecomment-149884227. We could use the robot as a base for our code examples.

rhempel:
Please review a43897fe0d380a6449b4dcd1c691be551056b3dd to see if this fixes the issue. I have added links to "Demo Robot" and "Demo Code" in the docs - this will be part of ReadTheDocs as well after the next release.

The resulting GitHub README.rst page [is here](https://github.com/rhempel/ev3dev-lang-python/blob/update-readme-content/README.rst)

bittner:
Please note the [proposed directory structure](https://github.com/rhempel/ev3dev-lang-python/commit/a43897fe0d380a6449b4dcd1c691be551056b3dd#commitcomment-15235513) for demos! Thank you.

PR #80, opened 2015-11-07
Adding auto-drive demo

ddemidov:
Ref #47

### DIFF
--- a/demo/README.rst
+++ b/demo/README.rst
@@ -1,0 +1,3 @@
+The examples in this folder, unless stated otherwise, are based on Explor3r
+robot by Laurens Valk. The assembling instructions for the robot may be found
+here: http://robotsquare.com/2015/10/06/explor3r-building-instructions.
--- a/demo/auto-drive.py
+++ b/demo/auto-drive.py
@@ -1,0 +1,140 @@
+#!/usr/bin/python
+
+# -----------------------------------------------------------------------------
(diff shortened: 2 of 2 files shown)


ddemidov:
The demo depends on #76.

rhempel:
Thanks for adding this, what I'll do is post about it to the ev3dev.org site - there's not much going on there right now.

ddemidov:
There is a slight issue I had with the new bindings while implementing this. The boostc version stopped the motors in [their destructors](https://github.com/ddemidov/ev3dev-lang-python/blob/master/ev3dev/__init__.py#L123-L140), and pure python version does not. When something goes wrong in the middle of the script, motors keep rotating, and there is no easy way to stop the motors. May be we should reintroduce the boostc behavior?

ensonic:
+1 I'll do this on the code generated by open-roberta too.

rhempel:
Good idea - destructors should reset the peripheral - this may apply for sensors as well...

ddemidov:
How do you reset a sensor?

ensonic:
Maybe @rhempel meant to restore e.g. the mode.

ddemidov:
But then we either need a concept of default mode for each sensor type, or to remember the state the sensor had at construction time.

rhempel:
I just mean to consider resetting the sensor. For example the NXT light sensor allowed to use the LED as a multi colored light. The destructor should turn the light off. I see the motor destructor PR is done - leave the sensors for later - it won't break anything :-)

PR #85, opened 2015-11-08
Use os.stat instead of os.access

### LINKED ISSUE #84: leds api defunct (for root)
Previously I could do:
```
ev3dev.led.orange_on()
```
later this got changed to
```
ev3dev.Led.orange_on()
```
and now it is supposedly
```
ev3dev.Leds.orange_on()
```
but it does not exist. There seems to be no wait to set both leds in one call anymore? Instead I seem to have to do
```
ev3dev.Leds.set_color(ev3dev.Leds.LEFT, ev3dev.Leds.ORANGE)
```
but that is broken:
```
ev3dev.Leds.set_color(ev3dev.Leds.LEFT, ev3dev.Leds.ORANGE)
  File "ev3dev/ev3.py", line 76, in set_color
    l.brightness_pct = v * pct
  File "ev3dev/core.py", line 1688, in brightness_pct
    self.brightness = value * self.max_brightness
  File "ev3dev/core.py", line 1606, in max_brightness
    return self.get_attr_int('max_brightness')
  File "ev3dev/core.py", line 169, in get_attr_int
    return int(self._get_attribute(attribute))
  File "ev3dev/core.py", line 162, in _get_attribute
    return self._attribute_cache.read(abspath(self._path + '/' + attribute))
  File "ev3dev/core.py", line 82, in read
    f = self.file_handle(path)
  File "ev3dev/core.py", line 74, in file_handle
    f = open(path, mode, 0)
IOError: [Errno 13] Permission denied: '/sys/class/leds/ev3-left0:red:ev3dev/max_brightness'
```

ddemidov:
This gives true file permissions irregardless of the current user.

Tested both for normal user and root.

Fixes #84

### DIFF
--- a/ev3dev/core.py
+++ b/ev3dev/core.py
@@ -36,6 +36,7 @@ import array
 import mmap
 import ctypes
 import re
+import stat
 from os.path import abspath
 from PIL import Image, ImageDraw
 from struct import pack, unpack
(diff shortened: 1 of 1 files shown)


rhempel:
Nice - that looks like it will do the job - using group attributes is a good idea

PR #88, opened 2015-11-10
Rename EV3 leds after the latest kernel update

ddemidov:
See http://www.ev3dev.org/news/2015/11/09/Kernel-Release-Cycle-8.

Depends on rhempel/ev3dev-lang#6

### DIFF
--- a/ev3dev/ev3.py
+++ b/ev3dev/ev3.py
@@ -47,10 +47,10 @@ class Leds(object):
 
 # ~autogen led-colors platforms.ev3.led>currentClass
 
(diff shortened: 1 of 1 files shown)


ddemidov:
Tested with the new kernel.

ddemidov:
We can in principle switch to automatic discovery of led groups at this point (see ev3dev/ev3dev#412). We would only need to put user-friendly group names into spec (`ev3:red` -> `RED`, `pistorms:BA` -> `A`, etc). It may probably be a good idea to postpone this until after 0.5.0 is released.

PR #89, opened 2015-11-11
update debian/changelog for package release

dlech:
I've gone ahead and packaged this and pushed it to ev3dev.org.

You can now install with `sudo apt-get install python-ev3dev` (after `sudo apt-get update` of course).

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,5 +1,5 @@
-python-ev3dev (0.4.2) stable; urgency=low
+python-ev3dev (0.5.0) stable; urgency=low
 
-  * source package automatically created by stdeb 0.8.2
+  * Initial Release.
 
- -- Ralph Hempel <rhempel@hempeldesigngroup.com>  Sun, 01 Nov 2015 16:51:45 -0500
+ -- David Lechner <david@lechnology.com>  Tue, 10 Nov 2015 21:30:53 -0600


dlech:
For git-buildpackage, this commit should be tagged with `debian/0.5.0`, but I don't know if that will mess up the other hooks you have on this git repository.

ddemidov:
We can do two things:

1. If git-buildpackage supports unannotated tags, we can create `debian/0.5.0` as unannotated tag. Travis-CI will trigger upload to PyPI anyway, but our setup script will ignore the tag, and will upload it as version `0.5.0.post1` (because its one commit since 0.5.0).
2. We can update our [git_version.py](https://github.com/rhempel/ev3dev-lang-python/blob/master/git_version.py) to parse tags like `debian/0.5.0`. In this case we will have to only make tags like this, because I think PyPI won't allow to reupload same version.

dlech:
In the future, if @rhempel makes sure that the debian changelog is up to date before he tags, this won't be an issue. and yes, I imagine that the the git-buildpackage tags are not annotated. And really it is up to @rhempel if he wants to use git-buildpackage. I find it useful to know which commit a debian package was built from if it was not the regular release tag.

dlech:
git-buildpackage can also be configured to use any patter you like for the tag.

rhempel:
I will gather the steps I need to do for a full release cycle - right now I just push an annotated (x.y.z) style tag in my reference repo, and that triggers a Travis suite, ReadTheDocs rebuild, and a build and push to PyPi.
Then I update the `ev3dev\ev3dev-lang` reference repo - and that should make a new pacage build possible - so the trick is to remember to get the Debian packaging files updated. Probably adding a ReleaseChecklist file would help :-)

rhempel:
So do I need to move the 0.5.0 tag to point to this (and maybe the next) commit?

dlech:
> So do I need to move the 0.5.0 tag to point to this (and maybe the next) commit?

No. The package on PyPi was created from your tag, so don't move it. The debian tag is used by git-buildpackage. If you don't plan on using that, then you don't need to add a tag here. We can just let this one instance slide.

rhempel:
OK, so if I remember to update the Debian changelog, I just create the annotated `x.y.z` tag and then everything "Just Works"?

dlech:
Yes. And you should build a debian source package from that tag as well. `debuild -S`. I'm going to work on setting up ftp accounts so that you and @ensonic can upload source packages to get published on ev3dev.org.

PR #90, opened 2015-11-11
Implement device enumeration functions

ddemidov:
This adds ability to enumerate devices available on the system. This is a multi-step PR, with steps being:

* Rename `name` argument of Device constructor to `name_pattern`. Some devices in `sys/class` have `name` attribute (e.g. `/sys/class/input*/name`). Having a positional argument `name` would
prohibit filtering such devices by `name` attribute in kwargs (because positional argument `name` would clash with keyword argument `name`).
* Implement `list_device_names` generator function. The function takes same arguments as `Device` constructor and lists all device names that match the provided arguments. Example:
```py
>>> [n for n in list_device_names('/sys/class/tacho-motor', 'motor*')]
['motor0', 'motor1']
```
* Implement `list_devices` generator function. The function takes same argument as `Device` class and enumerates all devices present in the system that match the provided arguments. In order to implement the function, optional argument `name_exact` was added to `Device` constructor. When `name_exact=True`, the constructor assumes that `name_pattern` argument contains exact device name and constructs the `Device` directly, without any checks. Example:
```py
[m.get_attr_string('port_name') for m in list_devices('tacho-motor', 'motor*')]
['outB', 'outC']
```
* Implement list_motors generator function. Enumerates all tacho motors present in the system. Example:
```py
[m.port_name for m in list_motors()]
['outB', 'outC']
```

More generator functions may be easily added later for different class types.

### DIFF
--- a/ev3dev/brickpi.py
+++ b/ev3dev/brickpi.py
@@ -47,8 +47,8 @@ class Leds(object):
 
 # ~autogen led-colors platforms.brickpi.led>currentClass
 
-    blue_one = Led(name='brickpi1:blue:ev3dev')
-    blue_two = Led(name='brickpi2:blue:ev3dev')
+    blue_one = Led(name_pattern='brickpi1:blue:ev3dev')
+    blue_two = Led(name_pattern='brickpi2:blue:ev3dev')
 
     BLUE_ONE = ( blue_one, )
     BLUE_TWO = ( blue_two, )
--- a/ev3dev/core.py
+++ b/ev3dev/core.py
@@ -104,14 +104,14 @@ class Device(object):
 
     _DEVICE_INDEX = re.compile(r'^.*(?P<idx>\d+)$')
 
-    def __init__(self, class_name, name='*', **kwargs):
+    def __init__(self, class_name, name_pattern='*', **kwargs):
         """Spin through the Linux sysfs class for the device type and find
-        a device that matches the provided name and attributes (if any).
+        a device that matches the provided name pattern and attributes (if any).
 
         Parameters:
             class_name: class name of the device, a subdirectory of /sys/class.
                 For example, 'tacho-motor'.
-            name: pattern that device name should match.
+            name_pattern: pattern that device name should match.
                 For example, 'sensor*' or 'motor*'. Default value: '*'.
             keyword arguments: used for matching the corresponding device
                 attributes. For example, port_name='outA', or
@@ -131,7 +131,7 @@ class Device(object):
         self._attribute_cache = FileCache()
 
         for file in os.listdir(classpath):
-            if fnmatch.fnmatch(file, name):
+            if fnmatch.fnmatch(file, name_pattern):
                 self._path = abspath(classpath + '/' + file)
 
                 # See if requested attributes match:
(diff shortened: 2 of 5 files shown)


ddemidov:
The existing code should not be affected by this.

dlech:
Why not use <https://pyudev.readthedocs.org/en/latest/>?

ddemidov:
I was not aware of that (or, rather, I forgot about that). This PR just moves the enumeration code out of [`Device` constructor](https://github.com/rhempel/ev3dev-lang-python/blob/59e27bbf40d820795bd4d60847f52e2f333fbd07/ev3dev/core.py#L133-L147) into [`list_device_names`](https://github.com/ddemidov/ev3dev-lang-python-1/blob/e113e466b9394a6b860023bd1026bbf082408fc4/ev3dev/core.py#L126-L130) function. So we could use pyudev even before this PR. I wonder which solution is more lightweight though.

Also, is it possible to filter devices in pyudev based on sysfs attributes? Something like
```py
[d for d in list_devices('tacho-motor', 'motor*', driver_name='lego-ev3-l-motor')]
```

dlech:
> Also, is it possible to filter devices in pyudev based on sysfs attributes?

Yes, we also have properties for this as well, so it would look something like this...

    [d for d in context.list_devices(subsystem='tacho-motor', LEGO_DRIVER_NAME='lego-ev3-m-motor')]

The properties are listed by the `uevent` attribute of each sysfs node.

There is also `match_attribute()` for matching attributes on the enumerator returned by `list_devices()` if there is not property.

ddemidov:
> Yes, we also have properties for this as well, so it would look something like this...

Right now the keyword arguments (attribute matches) may come from the user, so we either need to ask of him/her to know how attributes translate to properties, or add another mapping into `spec.json`. Unless there is an easy name-translation rule.

rhempel:
So go ahead with this PR - we're not going to use `pyudev` ?

ddemidov:
I'd like to do some tests on pyudev first. If it works good, may be we want to switch to it for all interaction with sysfs?

rhempel:
There is no rush to change anything now that 0.5.0 is released :-)

ddemidov:
Here is a test of relative performance of attribute access with pyudev and python-ev3dev: http://nbviewer.ipython.org/gist/ddemidov/ca99b74d054cb56cf037. It shows pyudev is about twice as fast, so we may want to switch to using it and get rid of some complexity in our code. @rhempel, what do you think?

rhempel:
I have a few questions:

- How much complexity is it actually getting rid of?
- Does it affect the `ev3dev` package load time?
- The "it may be caching results" statement is bothering me 

I think we always want to return the actual result of a property query - caching can result in strange behaviour of robots :-)

ddemidov:
> How much complexity is it actually getting rid of?

We will replace device enumeration code (currently in Device constructor), we will remove FileCache class, we will replace attribute access methods in Device with pyudev ones. Also, we will restore python3 support, which is [currently broken](https://github.com/rhempel/ev3dev-lang-python/pull/68#issuecomment-154380837).

> Does it affect the ev3dev package load time?

I am away from ev3 now, so I can not answer this right away. Will come back in several hours.

> The "it may be caching results" statement is bothering me

If I remove `d.get_attr_string('name')` from [1], it reports 26 times difference between best and worst times for python-ev3dev (and it guesses that caching is going on there as well).

ddemidov:
> Does it affect the ev3dev package load time?

Here are some results on ev3:
```
$ time python -c 'pass'

real	0m0.843s
user	0m0.600s
sys	0m0.160s

$ time python -c 'import ev3dev.ev3'

real	0m1.806s
user	0m1.390s
sys	0m0.250s

$ time python -c 'import ev3dev.ev3; import pyudev'

real	0m2.712s
user	0m1.890s
sys	0m0.610s

```

dlech:
Yes, it does cache results, so you should not use `d.get_attr_*` for attributes that change value.

ddemidov:
So I guess we stick to what we have now.

rhempel:
Implementing the device enumeration feature is still a good idea and does not break anything, right?
I'll merge into `develop` - including the LED fix, knowing that it will change when the kernel changes.

rhempel:
Thanks!

PR #108, opened 2016-01-03
Update tests to use the new functionality in fake-sys

ddemidov:
See https://github.com/ddemidov/ev3dev-lang-fake-sys/issues/1, https://github.com/ddemidov/ev3dev-lang-fake-sys/issues/2.


rhempel:
Thanks!

PR #110, opened 2016-01-05
Make use of special sensor helper functions defined in spec

ddemidov:
This uses [specification](https://github.com/ev3dev/ev3dev-lang/blob/8f4b4e61da33048304d5cd1c26a085114e6f8160/autogen/spec.json#L920) of special sensor classes added by @WasabiFan in ev3dev/ev3dev-lang#127. In effect, this gets rid of manual enumeration of special sensor classes in `ev3dev/core.py` in favor of liquid loop, and also adds helper functions (for example, `TouchSensor.is_touched()`).

### DIFF
--- a/ev3dev/core.py
+++ b/ev3dev/core.py
@@ -1443,7 +1443,36 @@ class I2cSensor(Sensor):
 
 
 # ~autogen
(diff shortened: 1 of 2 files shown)


ddemidov:
Wrong branch

PR #107, opened 2016-01-02
Initial pass at expanding the README for this project

rhempel:

PR #112, opened 2016-01-05
Get PR #107 that was inadvertently applied against master

rhempel:


rhempel:
Whoops - wrong direction

PR #118, opened 2016-01-13
Add demo/lego_official_projects with support for TRACK3R

dwalton76:


ev3dev/core.py
```diff
@@ -458,7 +459,7 @@ def position_sp(self, value):
     @property
     def speed(self):
         """
-        Returns the current motor speed in tacho counts per second. Not, this is
+        Returns the current motor speed in tacho counts per second. Note, this is
```
ddemidov: This is autogenerated from [this line](https://github.com/ev3dev/ev3dev-lang/blob/946c4622aa97fb4d7d2efd73f41d404db58dc7e5/autogen/spec.json#L131) in spec.json. The change should go there.

demo/lego_official_projects/common.py
```diff
@@ -0,0 +1,86 @@
...
+            if motor not in OUTPUTS:
+                log.error("%s in an invalid output, choices are %s" % (motor, ''.join(OUTPUTS)))
+                sys.exit(1)
+
+        self.left_motor = LargeMotor(left_motor)
+        self.right_motor = LargeMotor(right_motor)
+
+        for x in (self.left_motor, self.right_motor):
+            if not x.connected:
+                log.error("%s is not connected" % x)
```
ddemidov: This will print something like
```
<ev3dev.core.LargeMotor object at 0x7f10ec129e10> is not connected
```
Not very informative. May be its better ti use the port name somehow?

demo/lego_official_projects/common.py
```diff
@@ -0,0 +1,86 @@
...
+                sys.exit(1)
+
+        self.left_motor.reset()
+        self.right_motor.reset()
+        self.speed = 90
```
ddemidov: The name `speed` is confusing, because (1) it is used to control `duty_cycle_sp`, and (2) there is actually `Motor.speed` attribute which is read-only and returns current speed. I suggest renaming this to `duty_cycle_sp`.

demo/lego_official_projects/common.py
dwalton76: How about we add something like this to the Motor class, then we get the nicer format for all debugs, etc that are trying to print an instance of a Motor

def __str__(self):
   return "MOTOR_%s" % self.address

ddemidov: That is probably a nice idea, but won't help in this case, since `self.address` will raise an exception if motor is not connected.

demo/lego_official_projects/common.py
```diff
@@ -0,0 +1,86 @@
...
+
+    def move_right_track_forward(self, state):
+        if state:
+            self.right_motor.run_forever(duty_cycle_sp=self.speed)
+        else:
+            self.right_motor.stop()
+
+    def move_left_track_backward(self, state):
+        if state:
+            self.left_motor.run_forever(duty_cycle_sp=self.speed * -1)
```
ddemidov: I think there were a couple of models where the you had to reverse motor speeds to go forward. May be the tank class should provide a way to do the inversion.

demo/lego_official_projects/common.py
```diff
@@ -0,0 +1,86 @@
...
+        else:
+            self.left_motor.stop()
+
+    def move_right_track_backward(self, state):
```
ddemidov: Do you think parameterizing the functions would be a good idea? Something like [this](https://github.com/rhempel/ev3dev-lang-python/blob/82db0ce711b7e845123be2a72516712811317a50/demo/remote-control.py#L52-L79)?

demo/lego_official_projects/common.py
dwalton76: What variable does the motor class use to store the port that it is supposed to be connected to?

demo/lego_official_projects/common.py
dwalton76: ack

demo/lego_official_projects/common.py
dwalton76: Agreed, will add a method that will flip the polarity for both motors.

demo/lego_official_projects/common.py
dwalton76: I did think about whether or not a tank would always want to use run_forever() vs. one of the other run methods.  I think the majority of them will use run_forever() but certainly not all of them.  Once I build one that needs a different run call, passing a parameter to control which is called would do the trick.

ev3dev/core.py
dwalton76: Ack

demo/lego_official_projects/common.py
ddemidov: There is no such thing. The `address` is no different from any other attribute when device is connected. One can ask for a motor on port 'outA', or for a motor that has zero speed at the moment. The [`Device` constructor](https://github.com/rhempel/ev3dev-lang-python/blob/master/ev3dev/core.py#L139) just takes a bunch of keyword arguments, and tries to match each of those to a sysfs attribute.

What we could do, is to save the actual `kwargs` used for device initialization, and then use them inside `__str__` or `__repr__`. So that
```py
"%s" % LargeMotor(address='outA') == "LargeMotor(address=outA)"
```
and
```py
"%s" % LargeMotor(speed=0) == "LargeMotor(speed=0)"
```
@rhempel, what do you think?

demo/lego_official_projects/common.py
ddemidov: That is not exactly what I had in mind. The functions look very similar and only differ by the motor they are applied to and the direction. So all four of those could be replaced by
```py
def make_move(motor, dc_sp):
    def move(state):
        if state:
            motor.run_forever(duty_cycle_sp=dc_sp)
        else:
            motor.stop()
    return move
```

Then,
```py
    self.remote.on_red_up = self.make_move(self.left_motor, self.speed)
    ...
```

dwalton76: That works for me

demo/lego_official_projects/common.py
dwalton76: I have this now so that the Device class remembers the kwargs as self.kwargs along with:

    def __str__(self):
        return "%s-%s" % (self.__class__.__name__, self.kwargs.get('address'))

So the output for the code above is now "LargeMotor-outB is not connected"

rhempel: @dwalton - this looks interesting.

ddemidov: What if address was not specified during construction?

rhempel:
@dwalton - I really appreciate your work to get this library put together, and the suggested improvements to the Python API. I am working on the tacho-motor driver right now so I'll defer to @ddemidov to review this PR.

For future reference (and I'll update the README or add a CONTRIBUTING page) PRs should go against the `develop` branch, not `master`

demo/lego_official_projects/common.py
dwalton76: If address was not specified get() will return None. Normally with __str__ I try to keep the output fairly brief since you are typically using it to bring the object in debugs, etc.  We could add a method that returns pformat(self.kwargs) in case the user needs to see the full dictionary of what was passed.

dwalton76:
@rhempel ah I had not noticed the dev branch but will work off of that in the future.

demo/lego_official_projects/common.py
ddemidov: In that case this particular implementation of `__str__` should go into `Motor` and `Sensor` classes. There are devices (`Led`, for example), that do not have an associated address.

dwalton76: Hmm what if we do this instead, that way we can have a pretty generic `__str__` that lives in Device?
```
def __str__(self):
    if 'address' in self.kwargs:
      return "%s-%s" % (self.__class__.__name__, self.kwargs.get('address'))
  else:
    return self.__class__.__name__
```

ddemidov: I guess that could work. And we always can override this in derived classes if we need to.

dwalton76: Sample output with latest commit:

```
robot@ev3dev[projects]# ./TRACK3RWithClaw 
2016-01-15 01:50:21,810  INFO: Starting TRACK3RWithClaw
2016-01-15 01:50:22,071 ERROR: MediumMotor-outA is not connected
robot@ev3dev[projects]# ./TRACK3RWithClaw 
2016-01-15 01:50:34,971  INFO: Starting TRACK3RWithClaw
2016-01-15 01:50:35,120 ERROR: RemoteControl is not connected
robot@ev3dev[projects]# 
```

ev3dev/core.py
```diff
@@ -605,9 +616,20 @@ def time_sp(self):
     @time_sp.setter
     def time_sp(self, value):
         self.set_attr_int('time_sp', value)
+# ~autogen
 
+    def wait_for_running(self):
+        while True:
+            if 'running' in self.state:
+                return
+            sleep(0.01)
```
ddemidov: @dlech, is it possible for this function to be useful?

1. If the motor is running, and we call the function, it will return immediately.
2. If the motor is not running, is there any possibility for the function to actually return? (except for multithreaded operations).

I only see it used [here](https://github.com/rhempel/ev3dev-lang-python/pull/118/files#diff-99966894a0f2bcb1a5bc3f35219da892R39), right after the call to `run-to-rel-position`. Is it possible for the state to not contain `'running'` after the run command was sent to the motor?

ev3dev/core.py
```diff
@@ -182,8 +184,16 @@ def get_index(file):
                 self._device_index = get_index(name)
                 self.connected = True
             except StopIteration:
+                self._path = None
+                self._device_index = None
                 self.connected = False
 
+    def __str__(self):
+        if 'address' in self.kwargs:
+            return "%s-%s" % (self.__class__.__name__, self.kwargs.get('address'))
```
ddemidov: I think
```py
return "%s('%s')" % (self.__class__.__name__, self.kwargs.get('address'))
```
would look better (e.g. `MediumMotor('outA')`). It looks closer to what may actually be written in code, and hence more familiar.

ev3dev/core.py
rhempel: I think you mean - is there a bug waiting to happen?

If the programmer has not turned the motor on yet, then this function will wait quite some time!

ddemidov: That was what I implied, yes :)

ev3dev/core.py
dwalton76: Sounds good

ev3dev/core.py
dwalton76: I'll throw a counter in so that it raises an Exception if the motor hasn't started after 1s, same for wait_for_stop but maybe with a longer timeout there.

I needed this when I was working on MINDCUB3R (will submit that in the next few weeks, I need to rebuild it). None of the run_XYZ() calls block so if I say run_to_rel_pos(180) and need to take some action once it reaches 180 I need to know that A the motor started moving and B it stopped.  I don't remember the exact scenario but there were times when I got buggy behavior prior to implementing wait_for_running().

dlech: > Is it possible for the state to not contain 'running' after the run command was sent to the motor?

Yes, with polled motor controllers (I2C, UART, etc.) it could take until the next poll for the running flag to be set.

ddemidov: > I'll throw a counter in so that it raises an Exception if the motor hasn't started after 1s

May be not a counter, but a timeout? And the function could be more generic, something like

```py
def wait_for(condition, timeout=1e5, interval=0.01):
  tic = time.time() + timeout
  done = condition()
  while time.time() < tic and not done:
    sleep(interval)
    done = condition()
  return done
```

Then it could be used as
```py
started = wait_for(lambda: 'running' in motor.state, 1)
stopped = wait_for(lambda: 'running' not in motor.state)
```

Although I am not sure the function belongs to the core ev3dev package. @rhempel, what do you think?

rhempel: You beat me to that comment @ddemidov - it's a helper function, you want the timeout with default, not a fixed counter. And specifying the sleep interval is a nice touch.

For consistency with other langauges that may provide their own helpers, please specify the `timeout` and `interval` in milliseconds and have the function translate to the language specific units.

I don't think there's any benefit to higher than msec resolution.

You might get some pushback on the use of `lambda` to create the anonymous methods `started` and `stopped` but I think it's elegant.

ddemidov: > You might get some pushback on the use of lambda to create the anonymous methods started and stopped but I think it's elegant.

`started` and `stopped` in my example are not anonymous methods, they are just `bool`s. If `wait_for()` exits because the condition was met, it returns `True`. When the timeout has been hit, it returns `False`.

dwalton76: Regarding where to put these wait_for_XYZ() methods, agreed they are not core functionality. I'll create a child Motor class in my common.py and move these there.

As for common.py, any chance I can talk you into letting me move this into the ev3dev directory?  That way someone could easily "import ev3dev.common".  Maybe helper.py would be a more appropriate name.

I think we'll end up with a lot of good stuff in common/helper that will make life much easier for anyone programming EV3 in python.  The easier it is to find/use the better.

ev3dev/helper.py
```diff
@@ -0,0 +1,139 @@
...
+# Motor classes
+# =============
+class MotorMixin(object):
+
+    def wait_for_running(self):
+        return wait_for(lambda: 'running' in self.state, 1)
+
+    def wait_for_stop(self):
+        return wait_for(lambda: 'running' not in self.state, 1)
+
```
ddemidov: Does this eliminate the need in [wait functions](https://github.com/rhempel/ev3dev-lang-python/pull/118/files#diff-52ffd7ac54a3727e62910423b8b5a485R621) in the core module? Should those be removed from there?

dwalton76: oops, yeah those were supposed to be removed from core.py.  Done now.

ev3dev/core.py
```diff
@@ -41,6 +41,7 @@
 from PIL import Image, ImageDraw
 from struct import pack, unpack
 from subprocess import Popen
+from time import sleep
```
ddemidov: This import is no longer needed here.

ddemidov:
@rhempel, this looks good to me. One last thing to do would be to squash all the commits together to keep the history clean.

dwalton76:
All squashed now :)

rhempel:
I'm going to defer to @ddemidov and @dwalton76 to have tested and approved this PR, merging now.

rhempel:
Actually, I'm going to ask that you resubmit against the `develop` branch

dwalton76:
Closing this pull request, will re-submit against develop branch

PR #119, opened 2016-01-17
Switch to bootstrap theme for documentation

ddemidov:
This is closer in design to both http://www.ev3dev.org and http://ev3dev-lang.readthedocs.org/. We can also potentially add documentation for python as a subproject to http://ev3dev-lang.readthedocs.org/ (see https://github.com/ev3dev/ev3dev-lang/issues/146#issuecomment-172049467).

The result may be previewed here: http://ev3dev-lang-python-1.readthedocs.org/en/docs-theme/


rhempel:
Thanks!

PR #120, opened 2016-01-17
Restructure documentation

ddemidov:
Split index.rst into several subdocuments:

  - motors
  - sensors
  - other classes

Use the autogen system to generate list of the special sensors.

The preview is here: http://ev3dev-lang-python-1.readthedocs.org/en/restruct-docs/


rhempel:
Thanks!

PR #125, opened 2016-01-21
Add a led constant for black.

ensonic:
This way we can avoid doing ev3.Leds.set_color(ev3dev.Leds.RIGHT, (0,0))
which is hard to discover.


ddemidov:
Nice idea! Should the same change be applied to the brickpi platform [here](https://github.com/rhempel/ev3dev-lang-python/blob/master/ev3dev/brickpi.py#L56)?

ensonic:
@ddemidov indeed. updated the PR.

PR #127, opened 2016-01-21
Add a led constant for black.

ensonic:
This way we can avoid doing ev3.Leds.set_color(ev3.Leds.RIGHT, (0,0))
which is hard to discover.

### DIFF
--- a/ev3dev/brickpi.py
+++ b/ev3dev/brickpi.py
@@ -53,6 +53,7 @@ class Leds(object):
     LED1 = ( blue_led1, )
     LED2 = ( blue_led2, )
 
+    BLACK = ( 0, )
     BLUE = ( 1, )
 
     @staticmethod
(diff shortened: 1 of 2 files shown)


ev3dev/brickpi.py
```diff
@@ -53,6 +53,7 @@ class Leds(object):
     LED1 = ( blue_led1, )
     LED2 = ( blue_led2, )
 
+    BLACK = ( 0, )
```
dlech: Wouldn't it make more sense to call this `OFF`?

ensonic: I though about this, but then we need to call BLUE -> ON instead. I don't mind adding ON/OFF aliases, since this way we have a 'portable' way to turn an led on/off regardless of the type (that is ON is going to be as ON as possible :))

ensonic:
ping?

ev3dev/brickpi.py
ensonic: So how should be go forward? 
1. Keep this as it is and merge?
1. Change BLACK to OFF and add an ON constant that is whatever is full brightness.

rhempel:
Thks!

PR #132, opened 2016-02-28
Make YELLOW yellower

alesolano:
Reduced red brightness from 50% to 10% in YELLOW tuple to actually get yellow light.

(Second try :dancer:)


rhempel:
Thanks!

PR #134, opened 2016-02-29
Document Sensor.value function

### LINKED ISSUE #133: sensor docs miss the value() function
http://python-ev3dev.readthedocs.org/en/latest/sensors.html#sensor
shouldn't there be the value() method too?

ddemidov:
Fixes #133


ensonic:
Perfect and thanks!

rhempel:
Thanks!

PR #135, opened 2016-03-01
core: improve motor docs

ensonic:
Add a paragraph to explain when attributes are handled.

See ev3dev/ev3dev#558


ddemidov:
The docstring is generated from spec.json file in ev3dev/ev3dev-lang repo. The change should go there instead: https://github.com/ev3dev/ev3dev-lang/blob/develop/spec.json#L7-L14

ensonic:
Thanks, sent new PR: https://github.com/ev3dev/ev3dev-lang/pull/151

PR #140, opened 2016-03-10
25% read/write speed increase

antonvh:
Removing the abspath call increases r/w speed by 25% on my ev3


rhempel:
Please resubmit PR against develop branch :-)

rhempel:
And thanks for catching this! I guess abspath() has significant overhead - there's no real need to call it as long as the root string `self._path` is an absolute path :-)

PR #149, opened 2016-03-17
Expose Screen._img as Screen.image.

### LINKED ISSUE #142: Screen().img got renamed to Screen()._img and is now also missing from the docs.
In open-roberta I was using this code to show image in the screen:
``` python
image = Image.frombytes('1', (178, 128), IMAGES[picture], 'raw', '1;IR', 0, 1)
lcd = ev3dev.Screen()
lcd.img.paste(image, (0, 0))
lcd.update()
```
I could fix this by now doing
```patch
- lcd.img.paste(image, (0, 0))
+ lcd._img.paste(image, (0, 0))
```
I can't find when that changed. Anyway this is not about blaming :) Can we expose this in a nicer way properly?

ensonic:
This is useful for blitting images to the screen.

Fixes #142

### DIFF
--- a/ev3dev/core.py
+++ b/ev3dev/core.py
@@ -2537,6 +2537,18 @@ class Screen(FbMem):
         """
         return self._draw
 
+    @property
+    def image(self):
+        """
+        Returns a handle to PIL.Image class that is backing the screen. This can
+        be accessed for blitting images to the screen.
+
+        Example::
+
+            screen.image.paste(picture, (0, 0))
+        """
+        return self._img
+
     def clear(self):
         """
         Clears the screen


ddemidov:
Just out of interest: does it accept images in any format, or do they have to be in b/w already?

ensonic:
The images need to be converted. This is what I do:
``` python
picture = Image.frombytes('1', (178, 128), IMAGES[0], 'raw', '1;IR', 0, 1)
````
and the static images are defined in:
https://github.com/OpenRoberta/robertalab-ev3dev/blob/develop/roberta/StaticData.py

rhempel:
Thanks!

PR #151, opened 2016-03-29
Update motor tests for BrickPi

dlech:
I've updated the motor test to work with BrickPi. I have it passing all of `motor_unittest.py` and half of the relevant tests from `motor_motion_unittest.py`. I didn't realize this until too late, but it looks like `motor_param_unittest.py` and `motor_unittest.py` are almost exactly the same. I imagine these two files need to be merged into a single file. Also, we will need some kind of command line parameter or platform check so that you can select which parameters to use for testing.


dlech:
You really should make `develop` the default branch.

PR #152, opened 2016-03-29
Update motor tests to work with BrickPi

dlech:
I've updated the motor test to work with BrickPi. I have it passing all of motor_unittest.py and half of the relevant tests from motor_motion_unittest.py. I didn't realize this until too late, but it looks like motor_param_unittest.py and motor_unittest.py are almost exactly the same. I imagine these two files need to be merged into a single file. Also, we will need some kind of command line parameter or platform check so that you can select which parameters to use for testing.


rhempel:
That's me not finding time to clean this stuff up before pushing - I want to get out of the habit of making everything gold-plated before release, but a bit of buffing would not hurt. `motor_unittest.py` was the original file and is now broken into two parts, I'll delete the original one.

dwalton76:
@dlech github says there are some conflicts to resolve with this one

WasabiFan:
I think the conclusion here was that these changes had already been made but there were two copies of the file, so the changes weren't necessary. Not sure if that's right though.

dlech:
This really should be cleaned up and merged, but it is not a priority for me right now.

dwalton76:
Closing this once since it was re-worked via #213

PR #170, opened 2016-05-29
PiStorms support

dwalton76:


ev3dev/auto.py
```diff
@@ -8,11 +8,17 @@ def current_platform():
         return 'ev3'
     elif machine == 'armv6l':
         return 'brickpi'
+    elif machine == 'armv7l':
```
dlech: So you only attach BrickPi to RPi 1 and PiStorms to RPi 2? What about RPi 3 (`armv8...`)? Also, Beaglebone is `armv7l`, so it will be incorrectly detected as `pistorms`.

See ev3dev/ev3dev#412

dwalton76: True :(  Looks like I need to wait for https://github.com/ev3dev/ev3dev/issues/523 and then revisit auto.py

dwalton76:
I'll re-open this post #523

Issue #166, opened 2016-05-21
MINDCUB3R


dwalton76:
I did a bunch of work earlier this year on the MINDCUB3R implemenation at https://github.com/cavenel/ev3dev_examples

I am working on porting that to here.

dwalton76:
@rhempel is tradition to leave the bugs open until the dev branch is merged into master or is it ok to go ahead and close them if they are resolved in develop?

PR #173, opened 2016-06-10
MINDCUB3R

dwalton76:
Right now the arm that flips the cube is a little aggresive and sometimes screws up. This is because ramp_down_sp is broken (see https://github.com/ev3dev/ev3dev/issues/661).  Once ramp_down_sp is fixed I'll submit an update and that should help make the flipping of the cube much more reliable.

rubiks.py and rubiks_rgb_solver.py were ported from https://github.com/cavenel/ev3dev_examples but I wrote all of rubiks_rgb_solver.py there so I am very familiar with it.  rubiks_rgb_solver.py takes about 7s to run, this is the pause after we have finished scanning the cube but haven't started solving it.  This originally took ~90s so I spent a lot of time getting it down to something reasonable....7s seems ok for now.  On a raspberry pi 2 it takes about 50ms!

rubiks.py I've done a lot of work on at this point but a huge amount of credit for it goes to cavenel.

The kociemba code that solves the cube is from the following. In the C directory I only changed the Makefile and the in python directory I made a few minor changes.  Originally I wanted the entire MINDCUB3R solution to be python but running the python version of kociemba takes about 90 seconds on the EV3 vs about 1s for the C version.
https://github.com/muodov/kociemba


dwalton76:
This is for issue #166

rhempel:
@dwalton76 - thks for this! I am currently moving the Python binding away from Python2 and making it Python3 only - mainly in anticipation of less support for Python2 in most distributions in the future.

Unfortunately, I like to do things one at a time, but I may have to fix the ramping issue before I get full Python3 compatibility.

dwalton76:
@rhempel is there a "kernel dev how-to" for ev3dev so I could take a stab at the ramp issue? I have family here that want to see the MINDCUBER work :)

dlech:
[This](https://github.com/ev3dev/ev3dev-buildscripts/) will get you building the kernel. Most of the relevant motor code is [here](https://github.com/ev3dev/lego-linux-drivers/blob/master/ev3/legoev3_motor.c) and [here](https://github.com/ev3dev/lego-linux-drivers/blob/master/motors/tacho_motor_class.c). 

Tip: the official LEGO firmware does not do any kind of ramping, so you should be able to make something work without the ev3dev ramping feature.

dwalton76:
@rhempel this is working now (it requires your ramp_down_sp fix)...ok for me to merge it?

rhempel:
@dwalton - you mean it requires the kernel that has the fix ? I will merge this evening.

dwalton76:
@rhempel yes

PR #185, opened 2016-08-01
Fix names of outputs for BrickPi.

fryfrog:


fryfrog:
Also, I have no idea how to test this besides the bit of fiddling in IRC that @dlech did with me.

dlech:
I think it's pretty safe to say that this will work. You should also fix the input ports while you are at it.

ddemidov:
Would it be a good idea to move this info into the [platforms section](https://github.com/ev3dev/ev3dev-lang/blob/develop/spec.json#L1460) of the ev3dev/ev3dev-lang specification? Something like 
```json
platforms = {
    "brickpi" : {
        "ports" : {
            "input" : {...},
            "output" : {...}
        }
    }
}
```

fryfrog:
It'll be this evening before I can fire up that Pi3 w/ BrickPi to get the input port names, I can create a second pull request for that.

rhempel:
Merging this PR, thanks! Looking forward to the input port fix.

PR #192, opened 2016-09-04
Ignore failure to import fcntl

### LINKED ISSUE #190: Remove dependency on fcntl or don't import it in a test environment
The library is almost entirely portable to the point that the tests can even run on Windows. The only issue is that `core` imports `fcntl`, which is only available on Linux systems. The functions from that module are only used for buttons and the display (only two calls), which aren't needed anyway for testing. So, it would be good to either remove the dependency on `fcntl` or not import that module when it isn't needed.

ddemidov:
This wraps `import fcntl` into a `try` - `except` block. If the module is not present, an attempt to use the `Button` class will fail anyway, but users will be able to run tests on windows.

Fixes #190

### DIFF
--- a/ev3dev/core.py
+++ b/ev3dev/core.py
@@ -41,7 +41,6 @@ import os
 import io
 import fnmatch
 import numbers
-import fcntl
 import array
 import mmap
 import ctypes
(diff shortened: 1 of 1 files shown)


WasabiFan:
Nice!

dwalton76:
@ddemidov can you change the "except:" to "except Exception"....the blanket "except:" will catch things that you do not want to catch such as KeyboardInterrupt

ddemidov:
Good catch, done

dwalton76:
Looking at this some more....how is this not barfing later on (line 2170, 2603, etc) where fcntl.ioctl() is called?

ddemidov:
It does. But the point of this is to make possible to run unit tests on windows (the tests do not touch the buttons).

dwalton76:
Do the tests use the framebuffer?  That class uses fcntl too.

WasabiFan:
If the tests used anything that required fnctl, wouldn't they throw an error anyway because they're accessing ev3dev-specific devices? If the offboard tests require a display to be connected to the PC, something else is probably wrong...

dwalton76:
True true

WasabiFan:
Cool! With this merged, the tests should be completely portable.

PR #201, opened 2016-09-07
Overhaul README with relevant links and step-by-step intro

WasabiFan:
This is currently unfinished, but for the most part is what I expect to have in the end.

What I've done here is attempt to restructure the README so that it tells you what you need to know as a beginner who hasn't used this library (or even ev3dev necessarily) before. I am waiting on some content on the main site before I finish this up, as I intend to put links to that content in place of the current TODOs.

Overall, how is it? Is anything missing or unclear?

### DIFF
--- a/ISSUE_TEMPLATE.md
+++ b/ISSUE_TEMPLATE.md
@@ -1,0 +1,10 @@
+<!--
+    Before we can help you, we need some information about what you are trying to do and what software you are using.
+    If you are reporting a bug or asking a usage question, please fill out the information below; otherwise you can delete this template.
+-->
+
+- **ev3dev version:** PASTE THE OUTPUT OF `uname -r` HERE
+- **ev3dev-lang-python version:**
+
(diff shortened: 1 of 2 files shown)


WasabiFan:
Umm... "this is unfinished" -> there are TODOs still in the file 😕

dwalton76:
I saw those, I figured you would be cleaning them up soon....else why submit a pull request?

WasabiFan:
I submitted the PR to get feedback, so that once those links light up on the site this doesn't then have to wait for review; otherwise it just spends time waiting when I could be making changes based on feedback. I do that fairly regularly (to get feedback before I invest effort in things that might be changed) but it isn't common practice so I don't expect that to be clear. Regardless, the rest of the content was ready as fas as I was concerned, so it doesn't make a difference. I'll be finishing those ones within the next few days as ev3dev.org changes get merged in.

Issue #202, opened 2016-09-08
web interface


dwalton76:
About two years ago I did a web interface for an ev3dev bot
http://programmablebrick.blogspot.com/2014/11/seanbot-ev3-robot-controlled-via-web.html

I kind of hacked it together, I did a REST API in python to control the motors but then I also had an apache web server running to serve the files.  It was a bit of a mess but I've had multiple people take that code and run with it in their own projects.

This ticket is to take that code and implement in helper.py but do it the right way to make it much easier to use...the goals are:
- do not use apache, use one python web server that provides the REST API and serves the files
- desktop interface similar to what seanbot had (click on the arrows to drive)
- mobile touchscreen interface.  This will have a "joystick" area where there is a large circle with a smaller circle in the middle.  The user will drag the small circle around within the larger circle to drive the robot.
- buttons to engage the medium motor.  

You'll be able to use this to drive many of the tank EV3 robots like TRACK3R, etc but it will put enough groundwork in place to be able to write different web interfaces for other types of robots.

dwalton76:
index.html for selecting desktop or mobile controls
![ev3-web-main](https://cloud.githubusercontent.com/assets/468309/18335844/81e56a76-7550-11e6-8da9-0c06f73106c0.png)

dwalton76:
Desktop UI
![ev3-web-desktop](https://cloud.githubusercontent.com/assets/468309/18335862/946ca448-7550-11e6-9d63-61b3f41f6258.png)

dwalton76:
Mobile UI
![ev3-web-mobile](https://cloud.githubusercontent.com/assets/468309/18335866/9ccebaae-7550-11e6-9643-1e3ad22495bd.png)

PR #203, opened 2016-09-08
Web interface for controlling Tank class robots

dwalton76:
For issue #202 

All of the icons came from http://www.freeiconspng.com/


WasabiFan:
This is very cool; if I have time, I think I'm going to try this out for myself so I can see how it is. A web interface like this is something I've wanted to build for a long time.

That being said, IMHO I think it might make more sense to give this its own repository and then link to it from the README and/or a document in the demo directory. My logic is this:

- It feels odd to have JS and even JQuery in a Python library's repo.
- This is something that is more than just a simple demo, but a project that I could see people opening PRs against directly.

P.S. It would be really interesting to see what would happen if you made it possible to take input from a game controller through the HTML5 gamepad APIs!

dwalton76:
I went back and forth about whether or not to include the jquery libraries or just link to google and download it from there. Including it allows you to only need wifi but not internet access in order to make this all work.  I'm not stuck on doing it this way if everyone would feel better about not including the jquery libraries in the repo here

As for whether this should go in its own repo my vote for now would be no just because the community here (ev3dev-lang-python) is pretty small  and if it goes into its own repo then you have to make people aware that it is there and build a community of people to work on it.  Maybe keeping it here brings in some folks interested in web development for ev3 and we get them to contribute.  That said if we started seeing lots of forks and pull requests specific to the web interface I think then it would be time to move it to its own repo.

HTML5 gamepad APIs...I've never touched them but from reading this:
http://gamedevelopment.tutsplus.com/tutorials/using-the-html5-gamepad-api-to-add-controller-support-to-browser-games--cms-21345

I don't think it would be a huge stretch to make this work here.  I see jquery there so it is just a matter of firing off the current ajax calls in response to the gamepad event instead of a touch/mouse event.

WasabiFan:
Ok, that makes sense. Let's just make sure that the demo directory isn't ever packaged in releases 😆

dwalton76:
I dropped the jquery and jquery ui files and am downloading them via google instead.  That dropped 390k characters of javascript from the pull request :)  @WasabiFan would love it if you get a chance to try this out and see if you hit any snags. I plan on leaving the pull request open for a week or so in case anyone wants to chime in.

WasabiFan:
Sorry that I haven't gotten to test this yet; I started setting it up but then got sidetracked by other work. I still plan to do it at some point, but evidently that may not be for another few days.

PR #206, opened 2016-09-20
develop needs four commits from master

dwalton76:


WasabiFan:
Honestly, I think it would be cleaner to cherry-pick the commit(s) you need, rebase them onto develop, and then reset master based on develop. That would require that no releases have been tagged to include those commits, though.

dwalton76:
@WasabiFan or @rhempel do either of you know which of these is correct for brickpi?  There were a few commit in master that had not made it in to develop
```
<<<<<<< HEAD
INPUT_1 = 'in1'
INPUT_2 = 'in2'
INPUT_3 = 'in3'
INPUT_4 = 'in4'
=======
INPUT_1 = 'ttyAMA0:S1'
INPUT_2 = 'ttyAMA0:S2'
INPUT_3 = 'ttyAMA0:S3'
INPUT_4 = 'ttyAMA0:S4'
>>>>>>> develop
```

dwalton76:
@WasabiFan I should be able to 'git pull origin master' from develop and pick these up that way....the github UI here is a little funky I assumed the pull request in this direction was going to do something like that.  I'll cancel the pull request once we figure out what the INPUT_ values should be.

WasabiFan:
They should be the latter, IIRC.

>  I should be able to 'git pull origin master' from develop and pick these up that way....

The part I was concerned about is that you will be creating a merge commit to apply these commits onto develop, and then when we do a release you will be bringing that merge into the master branch. It won't cause any problems, but it _feels_ dirty. For now, go with the simplest option as you described, and then in the future we should just make sure to not commit directly to master.

dlech:
BrickPi ports were recently fixed in the develop branch.

6084ed9fd2a5c653b89f4d9f06d3a76abdcc7e4e
0f145c4e6d03fcc6cb707d83c0fd626542908dbf

dlech:
Also, I think merging/pulling master into develop is a really bad idea. The branches have diverged too much and it will result in a spectacular mess and break many things in develop that have been fixed.

WasabiFan:
@dlech How so? AFAIK, it's just those four commits that are missing, so I wouldn't expect it to be too troublesome. Still, it would definitely be ideal to nuke the master branch and start fresh... but doing that requires working around tagged releases, which is difficult.

dlech:
I think in the past, it was a bit of a mess. It looks like it has been cleaned up now. I thought there were more accidental commits to master, but you are right, it should be fine.

I don't understand what you mean about working around tagged releases though.

dlech:
> which of these is correct for brickpi?

```
INPUT_1 = 'ttyAMA0:S1'
INPUT_2 = 'ttyAMA0:S2'
INPUT_3 = 'ttyAMA0:S3'
INPUT_4 = 'ttyAMA0:S4'
```

WasabiFan:
> I don't understand what you mean about working around tagged releases though.

If we were to force-push to master to clean up stray commits, we would need to make sure that we don't change any commits that have a tag pointing at them (or any descendants). If we did, those releases would be broken.

dwalton76:
I don't have any plans on doing a force push :)

I went into develop and did "git pull origin master" and picked up the commits from master that way so will close this pull request.

Issue #200, opened 2016-09-07
Getting version number of installed library


WasabiFan:
What is the most consistent way to get the version number of the installed version(s) of this library? Considering that it could theoretically installed via either `apt-get` or `pip`/`easy_install` _and_ there are Python 2 and Python 3 versions floating around, this could get a bit interesting. It would be fine to skip Python 2 and just worry about v3.

dwalton76:
Not sure about pip off the top of my head but for apt-get 'dpkg --list' will show you the version 
```
robot@ev3dev[~]# dpkg --list python-ev3dev
Desired=Unknown/Install/Remove/Purge/Hold
| Status=Not/Inst/Conf-files/Unpacked/halF-conf/Half-inst/trig-aWait/Trig-pend
|/ Err?=(none)/Reinst-required (Status,Err: uppercase=bad)
||/ Name                                      Version                   Architecture              Description
+++-=========================================-=========================-=========================-========================================================================================
ii  python-ev3dev                             0.6.0                     all                       Python language bindings for ev3dev
robot@ev3dev[~]# 
```

WasabiFan:
Hmmmm... That seems to be a good start; I think I'll do it like that for now. Do you know if there's an accessible `__version__` property exposed by this library? If not, it might make sense to add one (although I have no idea how to do that) to make the version easily accessible in the future.

dwalton76:
yep it is defined in debian/changelog
https://github.com/rhempel/ev3dev-lang-python/blob/develop/debian/changelog

dlech:
http://stackoverflow.com/questions/10214827/find-which-version-of-package-is-installed-with-pip

dwalton76:
@WasabiFan ok to close this one?  There really isn't anything to change to the python library for this.

WasabiFan:
I don't believe this is resolved. In my mind, asking someone to look up the version through apt-get is too complicated (regardless of the fact that it's so easy for us to do); the fewer moving parts, the better. Could we define a `__version__` property? That would solve the problem, I think.

dwalton76:
I guess it couldn't hurt anything to stick a ```__version___``` in core.py...we just have to make sure we keep it and the version in the debian changelog in sync.

WasabiFan:
Fixed by #208

ddemidov:
[...]

PR #208, opened 2016-09-22
Getting version number of installed library

dwalton76:
For issue #200

### DIFF
--- a/ev3dev/core.py
+++ b/ev3dev/core.py
@@ -62,7 +62,7 @@ INPUT_AUTO = ''
 OUTPUT_AUTO = ''
 
 # Keep the __version__ in sync with the package version in debian/changelog
-__version___ = '0.7.0'
+__version__ = '0.7.0'
 
 # -----------------------------------------------------------------------------
 def list_device_names(class_path, name_pattern, **kwargs):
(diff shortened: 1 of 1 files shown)


ev3dev/core.py
```diff
@@ -61,6 +61,9 @@
 INPUT_AUTO = ''
 OUTPUT_AUTO = ''
 
+# Keep the __version__ in sync with the package version in debian/changelog
+__version___ = '0.7.0'
```
dlech: Is the triple underscore intentional?

Surely there is a way to do this automatically, e.g. with setuptools?

dwalton76: triple underscore...good eye that is a typo, will fix.

There is probably a way to do this automatically but we aren't doing releases at a frantic pace so for now I think this is fine.  We can automate it down the road.

WasabiFan:
Nice! I should add a note about this in the README.

P.S. I'd recommend using the "squash and merge" option when merging PRs to squash all the commits into a single one. It's nice in cases like this where there's an extra commit just to fix a typo.

dwalton76:
Ah did not notice that github had a "squash and merge" option....slick.

PR #209, opened 2016-09-22
Handle README TODOs

WasabiFan:
I didn't actually add a link to one of them because I am not prepared to choose appropriate link targets, but I did add a link to new ev3dev.org content. Also fixed some bad formatting.


dwalton76:
Looks good to me.  Will wait a bit before I merge in case David wants to comment.

PR #216, opened 2016-09-24
Specify driver name in device constructors

### LINKED ISSUE #215: driver_name is missing from device constructors
4a873844cdeff38ebc5d870f8c8bdfedefa2e8e3 removed filtering by `driver_name` attribute from device constructors. Example:

https://github.com/rhempel/ev3dev-lang-python/commit/4a873844cdeff38ebc5d870f8c8bdfedefa2e8e3?diff=split#diff-52ffd7ac54a3727e62910423b8b5a485L800

As a result, when there is single large motor and single medium motor connected to the system, the following script, without explicitly specifying output ports, is not guaranteed to pick correct devices:
```py
lm = LargeMotor()
mm = MediumMotor()
```

ddemidov:
Fixes #215

### DIFF
--- a/ev3dev/core.py
+++ b/ev3dev/core.py
@@ -842,7 +842,7 @@ class LargeMotor(Motor):
 
     def __init__(self, address=None, name_pattern=SYSTEM_DEVICE_NAME_CONVENTION, name_exact=False, **kwargs):
 
-        super(LargeMotor, self).__init__(address, name_pattern, name_exact, **kwargs)
+        super(LargeMotor, self).__init__(address, name_pattern, name_exact, driver_name=['lego-ev3-l-motor'], **kwargs)
 
 
 # ~autogen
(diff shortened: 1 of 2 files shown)


ddemidov:
Just checked, this resolves #215 for me. I guess @rhempel omitted driver names by mistake? Will wait a couple of days before merging in case anybody objects.

ev3dev/core.py
```diff
@@ -893,7 +893,7 @@ class FirgelliL1250Motor(Motor):
 
     def __init__(self, address=None, name_pattern=SYSTEM_DEVICE_NAME_CONVENTION, name_exact=False, **kwargs):
 
-        super(FirgelliL1250Motor, self).__init__(address, name_pattern, name_exact, **kwargs)
+        super(FirgelliL1250Motor, self).__init__(address, name_pattern, name_exact, driver_name=['fi-l12-ev3-50'], **kwargs)
```
WasabiFan: This should be the new driver name (the company is now called Actuonix, or something like that); we might need to update the spec for that. The class name needs to be renamed as well.

ddemidov: ev3dev/ev3dev-lang#168

ddemidov: And c5faf77

WasabiFan:
When you merge, please choose the "squash" option and add a note to the message about renaming Firgelli.

PR #211, opened 2016-09-22
Update changelog for 0.7.0

dwalton76:

PR #217, opened 2016-09-24
Update changelog for 0.7.0 release

WasabiFan:
Updated version of #211 with some added items and its own entry. I decided to put my name on the release, based on what I've read about Debian packaging procedures; if that's not right, I can change the name.

Note that these changes are a delta from the RC release; is that the correct thing to do? Or should it be relative to the last stable release (0.6.0)?

The list was based off of this diff: https://github.com/rhempel/ev3dev-lang-python/compare/fb3cdbc...develop

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,14 +1,28 @@
 python-ev3dev (0.7.0) stable; urgency=medium
 
-  * Support "-13-ev3dev" and newer kernels. 
+  [Denis Demidov]
+  * Support "-13-ev3dev" and newer kernels.
+  * Rename FirgelliL1250Motor and FirgelliL12100Motor to ActuonixL1250Motor
+    and ActuonixL12100Motor.
+  * Allow passing espeak options to Sound.speak
+  * Make sure that device classes connect to devices of the right
+    type (driver name)
+
+  [Daniel Walton]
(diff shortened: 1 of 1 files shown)


WasabiFan:
@dlech can you review?

dlech:
It is traditional to give credit by starting a list item with `[<name>]` if it is not the person signing the changelog.

Also make sure to wrap to 80 lines. I didn't count, but one line looked kind of long, especially if you add the author.

WasabiFan:
OK, I'll add that.

WasabiFan:
:new:

WasabiFan:
Can I get a quick review from @dlech on this to confirm that it's properly formatted? If it is, we should be ready for release.

dlech:
LGTM. You can use `lintian` to tell for sure that it is properly formatted.

WasabiFan:
I wasn't able to persuade lintian to do what I wanted with my low level of patience, so I'm going to assume it's fine :laughing: It might be good for whoever actually does the release to run it on the `.deb` that they produce just in case, but even that's probably overkill; even if the changelog isn't perfect, it will still do its job.

I'm going to merge this and draft a GitHub release (but not actually release it). Is there anything else that people want to do before the library is ready?

Issue #172, opened 2016-06-05
Python3 Compatibility vs abandoning Python2 Compatibility


rhempel:
I am not a Python expert, so I'm going to solicit other opinions on this topic.

In the course of fixing up some other defects in the motor drivers and the Python bindings, and in adding unit tests for other EV3 sub-systems, I have had to use Python3 features and libraries in some scripts.

Besides the slower startup time, there are other minor, yet possibly incompatible differences from Python2 that may make force us to to go with one language version over the other. Strings are unicode in Python3 vs raw bytes in Python2, for example.

Can anyone thing of good reasons to stick with both Python2 and Python3 compatibility for the bindings, or are we better off moving to Python3 and abandoning Python2?

dlech:
I'm no python expert either, but I know that Debian has a plan to phase out python2 in the next 5 years or so. It will still be in stretch (currently unstable) but probably won't be in buster (stretch + 1). I'm all for python3 only.

rhempel:
I was leaning that way but wanted to collect some opinions :-) I'll wait a bit longer while I fix up the testing framework and fix a few binding bugs for more comments.

ensonic:
+1 one for python3, but I'd make this an explicit goal and not a side effect of some commit. We definitely want to make sure that all the new example actually work with python3.

rhempel:
You were the main opinion I was looking for :-) Stay tuned

WasabiFan:
This change has been made, correct? If so, can this be closed?

dwalton76:
yep python3 is now enforced
```
robot@ev3dev[lego_official_projects]# ./EV3D4WebControl 
Traceback (most recent call last):
  File "./EV3D4WebControl", line 3, in <module>
    from EV3D4 import EV3D4WebControlled
  File "/home/robot/ev3dev-lang-python/demo/lego_official_projects/EV3D4.py", line 5, in <module>
    from ev3dev.auto import OUTPUT_A, OUTPUT_B, OUTPUT_C, OUTPUT_D
  File "build/bdist.linux-armv5tejl/egg/ev3dev/auto.py", line 21, in <module>
  File "build/bdist.linux-armv5tejl/egg/ev3dev/ev3.py", line 29, in <module>
  File "build/bdist.linux-armv5tejl/egg/ev3dev/core.py", line 36, in <module>
SystemError: Must be using Python 3.4 or higher
robot@ev3dev[lego_official_projects]# 
```

closing this one

[...]

PR #219, opened 2016-09-25
Re-enabling python2 support

ddemidov:
Also enable python 2.7 in Travis build matrix

python-ev3dev is still installed in the latest nightly builds, because its a dependency of openroberta. But the version installed (0.6.0) is incompatible with the kernel. It turns out nothing in the library code prevents us from supporting python2, so this removes import time check for python version.

All tests are able to pass in python2 without any other changes. Also tested with [ev3rstorm](https://gist.github.com/ddemidov/fc6987229e8f846dbf157f6f063aac26) demo example; it works in both python versions.

openroberta will still need to be updated to the latest API though (pinging @ensonic).

Also there need to be some changes to debian packaging (pinging @dlech).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,6 @@
 language: python
 python:
+- 2.7
 - 3.4
 sudo: false
 install:
--- a/ev3dev/core.py
+++ b/ev3dev/core.py
@@ -30,13 +30,6 @@
 
 # -----------------------------------------------------------------------------
 
-import sys
-
-if sys.version_info < (3,4):
-    raise SystemError('Must be using Python 3.4 or higher')
-
-# -----------------------------------------------------------------------------
-
 import os
 import io
 import fnmatch


dlech:
See #172. What is the reason for re-enabling python 2?

dlech:
For packaging, [this commit](https://github.com/rhempel/ev3dev-lang-python/pull/180/commits/1fc22b23cf763f221550fc2b52eba279582cf60e) would need to be partially reverted - as in keep the python3 stuff and add the python2 stuff back.

WasabiFan:
> What is the reason for re-enabling python 2?

Primarily because we noticed that OpenRoberta depends on Python 2. Simplistically, it would be easier for users such as that to be able to upgrade at their own pace rather than be forced to do it. More practically, we don't want to have the `python-ev3dev` package installed in future image releases if it's out of date, and right now because we preinstall the OpenRoberta connecter we are also installing that old library. I would rather handle supporting a current version of the Python 2-based library than supporting people complaining of getting errors when they accidentally run their code with `python` instead of `python3`. My position is this: either remove everything depending on `python-ev3dev` from the default image or publish a version of that package (or an equivalent) that we can support.

We've demonstrated that the code is fully compatible, so now it's just a matter of policy: whether we want to continue providing Python 2 support or not. I'm not particularly set on a specific option, however I want to make sure that whatever we do doesn't result in people getting confused by that old code.

ddemidov:
I see that @ensonic [okayed](https://github.com/rhempel/ev3dev-lang-python/issues/172#issuecomment-223828431) the switch to python3. So I guess we need to open an issue with openroberta to follow suit and do the switch as well? Or may be they just want to stick with the 2015-12-30 image for now.

rhempel:
I removed Python 2 support because I believed that I was using features in Python 3 to speed up attribute access and the photo-booth was Python 3 only. If Python 2 still works then I'm ok with reenabling it as long as speed and API is ok.
Thanks for picking up this task !

WasabiFan:
For simplicity, I think we should put this off until we get a Python 3 release out the door. I would prefer it if we got it sorted out beforehand so that we didn't have that problematic package installed by default, but I think it's more important to get our main package published so that we aren't blocking an ev3dev release.

It would be great if @ensonic could comment on this.

dlech:
You might ping @ensonic on IRC in case he is not getting GitHub notifications.

WasabiFan:
Yeah, I know he's been hanging around in there over the past few days, but I only caught some of his messages. I might head over there later today and see if I can get his attention.

ensonic:
I just ported the operobertalab package to python3 today. Only a few niggles left, but works mostly.

WasabiFan:
Glad to hear that! I think we will probably want to hold off on an ev3dev image release until you get a new version of the OpenRoberta lab connector package published which doesn't depend on the old library, so let us know when that happens. We expect to release a new version of the Python lib (including many bug fixes) soon as well.

ddemidov:
Looks like we don't need python2 after all.

ensonic:
@WasabiFan I'll do some more testing next week.

Issue #207, opened 2016-09-20
release 0.7.0


dwalton76:
@WasabiFan I think you said there is one more tweak to make to the README but it is waiting on a PR for the docs to go through correct?  After that I think we can do a release.

WasabiFan:
Ok, sounds good. I am waiting on a review/merge of https://github.com/ev3dev/ev3dev.github.io/pull/237 so that the README isn't confusing to new users. After that, I think this'll be ready.

dwalton76:
These are the items that are currently in the changelog for 0.7.0
- Drop python 2.x support.
- Performance improvements for reading/writing sysfs attributes.
- Updates for breaking ev3dev kernel changes.

Should we mention that many demos have been added? They are not installed as part of the package but in the changelog we could point people to https://github.com/rhempel/ev3dev-lang-python/tree/develop/demo I think we should add this to the changelog but I am biased as I did a lot of work on the demo programs.  Proposed text:
- Added many demo programs including BALANC3R, MINDCUB3R, EXPLOR3R, TRACK3R and EV3D4. A touchscreen web interface is available for tank robots, see EV3D4WebControl. The demo programs are available at https://github.com/rhempel/ev3dev-lang-python/tree/develop/demo

Other things to add:
- support for ev3dev kernel  4.4.15-13-ev3dev-ev3
- Added ev3dev/helper.py which includes classes for Tanks, Motors, Web UI controls, etc
- Added ev3dev/GyroBalancer.py for robots that use a gyroscope to balance on two wheels

dlech:
> Should we mention that many demos have been added?

Not in the debian changelog. If there is another changelog, then sure, mention the demos if you like.

WasabiFan:
> Should we mention that many demos have been added?

We should definitely advertise them, whether it is in the changelog or somewhere else. If it's in something version-specific, we should make sure to link to the version of the tree at the release tag instead of `master` to make sure it works in the future.

----

Also note that the important kernel version number is the `13`; the rest of it is specific to the platform. So, we should say "`-13-ev3dev` kernels".

WasabiFan:
The dependency that we had on ev3dev.org content has fallen into place, so I have opened #209 to update the README accordingly. Once that is merged we should be ready to go.

dwalton76:
[...]

PR #224, opened 2016-09-30
Preparing debian release for 0.7.0

ddemidov:
Summary of changes:

1. Update copyright information in debian/copyright (see https://github.com/rhempel/ev3dev-lang-python/issues/207#issuecomment-250687801)
2. Fix lintian warnings (a line too long, and wrong weekday)
3. Get rid of hard-coded `__version__` in `ev3dev/core.py`, and replace it with auto-generated `ev3dev/version.py`.

### DIFF
--- a/ev3dev/core.py
+++ b/ev3dev/core.py
@@ -61,15 +61,12 @@ except ImportError:
 INPUT_AUTO = ''
 OUTPUT_AUTO = ''
 
-# Keep the __version__ in sync with the package version in debian/changelog
-__version__ = '0.7.0'
-
 # -----------------------------------------------------------------------------
 def list_device_names(class_path, name_pattern, **kwargs):
 
     if not os.path.isdir(class_path):
         return
-    
+
     """
     This is a generator function that lists names of all devices matching the
     provided parameters.
@@ -174,7 +171,7 @@ class Device(object):
         mode = stat.S_IMODE(os.stat(path)[stat.ST_MODE])
         r_ok = mode & stat.S_IRGRP
         w_ok = mode & stat.S_IWGRP
-            
+
         if r_ok and w_ok:
             mode = 'r+'
         elif w_ok:
@@ -1420,7 +1417,7 @@ class Sensor(Device):
 # ~autogen
 
         self._value = [None,None,None,None,None,None,None,None]
-        
+
         self._bin_data_format = None
         self._bin_data = None
 
--- a/.gitignore
+++ b/.gitignore
@@ -5,4 +5,5 @@ __pycache__
 dist
 *.egg-info
 RELEASE-VERSION
+ev3dev/version.py
 build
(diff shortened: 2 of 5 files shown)


ddemidov:
@dlech, I followed http://www.ev3dev.org/docs/devtools/packaging-for-ev3dev/ up to the point I have `python3-ev3dev_0.7.0_all.deb` in `pbuilder-ev3dev/debian/jessie-armel/`. `lintian` seems to like the package, and I am able to install it with `dpkg -i python3-ev3dev_0.7.0_all.deb`

```py
>>> from ev3dev.version import __version__
>>> print(__version__)
```
gives me `0.7.0`. Assuming nobody objects against the changes proposed here, what are my next steps?

Do I need to generate the package for other values of `DIST` and `ARCH`? Or this does not matter since the deb name suggests it is suitable for all arches?

How do I release the package? Just send the deb to you?

WasabiFan:
Thanks for doing this!

ddemidov:
Can I get an approve from @dlech?

dlech:
> armel

Since this package does not contain any binary components, you can actually replace `armel` with `amd64` and the build will run much faster.

You will also need to build a second time with `OS=raspbian ARCH=armhf DIST=jessie` to produce a package for raspberry pi.

> what are my next steps?

If you will be the uploader, you need to run `dch -r` in the source code directory to update the changelog with your name/email. You will also need to send me an SSH public key that will be used for uploading.

> How do I release the package? Just send the deb to you?

I will update http://www.ev3dev.org/docs/devtools/packaging-for-ev3dev/ shortly with more detailed steps on uploading.

dlech:
Since this git repo has special handling of tags, you may not want to push the debian package tag. For the releases I did, I just pushed the tag to my personal fork.

ddemidov:
Does `gbp` create annotated tags? If not, it should not affect anything. Also, the debian packaging script does not actually use tag info: it just dumps the version found in changelog into `RELEASE-VERSION`: https://github.com/rhempel/ev3dev-lang-python/blob/develop/debian/rules

dlech:
> Does gbp create annotated tags?

no

> If not, it should not affect anything

I thought that it will update the readthedocs site.

> Also, the debian packaging script does not actually use tag info: it just dumps the version found in changelog into RELEASE-VERSION:

Yes, this is intentional.

ev3dev/core.py
```diff
@@ -61,15 +61,12 @@
 INPUT_AUTO = ''
 OUTPUT_AUTO = ''
 
-# Keep the __version__ in sync with the package version in debian/changelog
```
WasabiFan: Why was this removed?

ddemidov: Because a4a296a creates ev3dev/version.py automatically. Less chances to forget to update this.

WasabiFan: Ah, I see. Missed that.

PR #225, opened 2016-09-30
Distribute fonts from xfonts-75dpi in PIL format

ddemidov:
This adds fonts converted from `xfonts-75dpi` package into PIL format. The list of all fonts provided may be obtained with
```py
import ev3dev.fonts
print(ev3dev.fonts.list())
```
Individual font may be loaded with `ev3dev.fonts.load('<fontname>')`. Here is an overview of all fonts (the grid corresponds to EV3 screen size):

download

This was created with http://nbviewer.jupyter.org/gist/anonymous/f25d8007c095d9a45cb169b6476d7370


ensonic:
Great! Does this need some extra change for the api to show up in the api docs?

ddemidov:
> Does this need some extra change for the api to show up in the api docs?

Good point. I'll add the changes to the documentation.

WasabiFan:
By the way, this looks good to me (although I haven't done a technical test). Feel free to merge if you are happy with it.

WasabiFan:
I think you should include that image in the docs as well.

ddemidov:
I've updated the pull request with the documentation changes. Preview of the documentation is available here: http://ev3dev-lang.readthedocs.io/projects/python-ev3dev/en/dist-fonts/other.html#screen

Issue #228, opened 2016-10-14
"Special" sensors use methods instead of properties


WasabiFan:
After some conversation with @ndward about errors he was seeing, I have realized that the "special" sensors are implemented with normal methods for each property instead of "real" Python properties. Why is this? If it's possible, I think it would probably be best to use the standard templates for the special sensors as well so any changes we make to the other classes also affect these ones.

I expect that the main reason for this is that the special sensors take a parameter to determine if they set the mode. I think it would be better to create some separate way of doing that so that we don't sacrifice the property-style access of these values.

ddemidov:
One reason I can give, is that special methods now take optional parameter ([`do_set_mode=True`](https://github.com/rhempel/ev3dev-lang-python/blob/develop/ev3dev/core.py#L1639), result of discussion in ddemidov/ev3dev-lang-cpp#10). It would be impossible to reimplement this with a property.

ddemidov:
Oops, did not see your edit. Well, we could create a set of two properties for each method, or maybe create a class-wide flag (`s.set_mode_on_each_access=False; print(s.is_pressed)`)

WasabiFan:
> we could create a set of two properties for each method

I could get on board with that. We'd need to make sure that it doesn't get confusing when looking at the docs though.

> or maybe create a class-wide flag (s.set_mode_on_each_access=False; print(s.is_pressed))

This requires some more thought on our part about the use cases here, but I think this option would be preferable over the first.

Overall, do you think that this change would be worth the benefit of using properties? I think it's moderately more intuitive, but at the same time these are all get-only, so presumably the idea of a method that returns a property value wouldn't be foreign to users anyway.

ddemidov:
Resolved by #229

PR #229, opened 2016-10-14
[breaking] Special properties

ddemidov:
This converts special sensor methods to properties for the sake of uniformity throughout the library. See discussion in #228.

**This will break existing code!**

Instead of
```py
print( cs.ambient() )   # function call syntax
cs.mode = 'COL-AMBIENT'
while True:
    log(cs.ambient(do_set_mode=False)) # do not set mode when polling the sensor value
```
one has to
```py
print( cs.ambient )   # property syntax
cs.mode = 'COL-AMBIENT'
cs.auto_mode = False  # do not set mode when polling the sensor value
while True:
    log(cs.ambient)
```

I've updated my part of the demo scripts. @dwalton76, if we decide to merge this, could you take care of yours?

### DIFF
--- a/demo/EV3RSTORM/ev3rstorm.py
+++ b/demo/EV3RSTORM/ev3rstorm.py
@@ -146,7 +146,7 @@ class ev3rstorm:
 
         # Now that the event handlers are assigned,
         # lets enter the processing loop:
-        while not self.ts.is_pressed():
+        while not self.ts.is_pressed:
             rc1.process()
             rc2.process()
             time.sleep(0.1)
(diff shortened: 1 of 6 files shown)


ensonic:
I never use do_set_mode in openroberta so fine for me.

ddemidov:
@ensonic, I was more concerned about function-call vs get-property syntax (`cs.ambient()` vs `cs.ambient`). I think openroberta was implemented before we introduced special sensor properties though, so I guess you use `cs.value()` instead? That should still work.

WasabiFan:
This looks good to me by the way, although I'd like to hear back from @ensonic on whether this will be an issue to get some user feedback.

ensonic:
@ddemidov , yes I use value(), see
https://github.com/OpenRoberta/robertalab-ev3dev/blob/develop/roberta/ev3.py#L452
FYI, all the code that interfaces with hardware is in there.

PR #233, opened 2016-10-20
Update changelog for v0.8.0

WasabiFan:
Now that we've added the fonts module, I think we should publish a release to get it out there.

I didn't add the release attribution yet; whenever we actually do this release we can add the attribution.

Also, at some point I think we should align with semver and release a v1.0. Honestly, this seems like a reasonable time to do it. Currently this PR just increments the minor version though.

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,5 +1,5 @@
 python-ev3dev (0.8.0) stable; urgency=medium
-  
+
   [Denis Demidov]
   * Add option to not set mode before reading sensor value
   * Return a tuple for multiple values from a sensor property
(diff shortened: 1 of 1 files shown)


dwalton76:
0.8.0 or 0.7.1? Not sure if we have any defined criteria for when we do a major version bump

WasabiFan:
Yeah, that's why I'd like to align with semver at some point... for now, my thinking was essentially that the breaking change on the `NXTMotor` class warranted something more than a patch increment. But it's arbitrary either way.

WasabiFan:
Haha sorry that I hadn't made these changes yet; looks like you did it for me. What name do we want at the bottom? Or do we wait until someone actually builds the release?

ddemidov:
I can try to build the package this weekend (while I still remember how to do this from 0.7.0). I'll put my name then with the rest of the changes required for packaging. This may be merged now.

PR #242, opened 2016-10-29
Use constants instead of repeating the strings.

### LINKED ISSUE #65: shoudn't the bindings use their own commands
In the code I see a lot of cases like:

```
COMMAND_RUN_FOREVER = 'run-forever'
...
self.command = 'run-forever'
```

Shouldn't the impl use the constants? If okay, I'll send a PR to fix this,

ensonic:
Fixes #65.

### DIFF
--- a/ev3dev/core.py
+++ b/ev3dev/core.py
@@ -750,7 +750,7 @@ class Motor(Device):
         """
         for key in kwargs:
             setattr(self, key, kwargs[key])
-        self.command = 'run-forever'
+        self.command = self.COMMAND_RUN_FOREVER
 
     def run_to_abs_pos(self, **kwargs):
         """Run to an absolute position specified by `position_sp` and then
@@ -758,7 +758,7 @@ class Motor(Device):
         """
         for key in kwargs:
             setattr(self, key, kwargs[key])
-        self.command = 'run-to-abs-pos'
+        self.command = self.COMMAND_RUN_TO_ABS_POS
 
     def run_to_rel_pos(self, **kwargs):
         """Run to a position relative to the current `position` value.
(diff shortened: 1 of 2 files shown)


ensonic:
Please let me know how to run the tests. I'll test this 'manually' on the ev3, but if you let me know, I can send another PR to add this to the README (a HACKING section).

ddemidov:
The tests should be run automatically by Travis-CI on each pull request (this PR just passed the tests).

But the changes you made should go into autogen templates, so that they won't be overwritten on next regeneration. Namely, [here](https://github.com/rhempel/ev3dev-lang-python/blob/ea66472d8ba7544fc54515790089e98bd019d8b8/templates/motor_commands.liquid#L12). Also, since both the constants and the commands are generated automatically from the same script, I am not sure that this is so big a problem.

ddemidov:
Well, the constants are generated [here](https://github.com/rhempel/ev3dev-lang-python/blob/ea66472d8ba7544fc54515790089e98bd019d8b8/templates/generic-property-value.liquid#L8), so not technically by  the same script. But still, both use same input data from `spec.json`, so I'd say there is little room for error.

ensonic:
The snesors use the constant. I'll change it. What is the command to regenerate?

ddemidov:
You need to update the `ev3dev-lang` submodule, and run `./ev3dev-lang/exec-autogen.js`. But first you need to setup the autogen (inside `ev3dev-lang`) as described here: https://github.com/ev3dev/ev3dev-lang/tree/develop/autogen.

ensonic:
@ddemidov, thats for the pointers. Here is a variant that also updates the template. The changes to core.py are made by `./ev3dev-lang/exec-autogen.js`.
I agree that it does not matter much, but now the motor code is in sync to the sensor code.

ddemidov:
Thanks!

PR #243, opened 2016-10-31
Replace autogenerated properties for trigger, delay_on/off with custom ones

### LINKED ISSUE #234: LED "delay_on"/"delay_off" broken?
- **ev3dev version:** 4.4.19-15-ev3dev-ev3
- **ev3dev-lang-python version:** 0.7.0

I'm getting odd errors while attempting to set the `delay_on`/`delay_off` attributes for use with the `timer` trigger. For example:

```
>>> Leds.set(Leds.LEFT, brightness_pct=1, trigger="timer", delay_on=2000)
  File "/usr/lib/python3/dist-packages/ev3dev/ev3.py", line 90, in set
    setattr(led, k, kwargs[k])
  File "/usr/lib/python3/dist-packages/ev3dev/core.py", line 2051, in delay_on
    self._delay_on = self.set_attr_int(self._delay_on, 'delay_on', value)
  File "/usr/lib/python3/dist-packages/ev3dev/core.py", line 213, in set_attr_int
    return self._set_attribute(attribute, name, str(int(value)))
  File "/usr/lib/python3/dist-packages/ev3dev/core.py", line 202, in _set_attribute
    attribute.write(value.encode())
io.UnsupportedOperation: File not open for writing
```

I have also seen a different error, which I can no longer reproduce:

```
>>> Leds.set(Leds.LEFT, brightness_pct=1, trigger="timer", delay_off=100)
  File "/usr/lib/python3/dist-packages/ev3dev/ev3.py", line 90, in set
    setattr(led, k, kwargs[k])
  File "/usr/lib/python3/dist-packages/ev3dev/core.py", line 2065, in delay_off
    self._delay_off = self.set_attr_int(self._delay_off, 'delay_off', value)
  File "/usr/lib/python3/dist-packages/ev3dev/core.py", line 213, in set_attr_int
    return self._set_attribute(attribute, name, str(int(value)))
  File "/usr/lib/python3/dist-packages/ev3dev/core.py", line 202, in _set_attribute
    attribute.write(value.encode())
OSError: [Errno 19] No such device
```

The first one (and likely the second) seem to be related to the file handle cache, but I can't quite tell. @ddemidov IIRC you have had a fair amount of experience with the caching system; thoughts?

I can write to these attributes without issue from Bash.

ddemidov:
This introduces a workaround for ev3dev/ev3dev#225 and fixes #234.

* First, when `trigger` is set to `timer`, the setter blocks until `delay_on` and `delay_off` files have been created and got the correct attributes. The maximum wait time is 1 second.
* Second, since `delay_on` and `delay_off` attributes are created and destroyed depending on value of `trigger`, their cache entries may become outdated. This commit makes an attempt to reopen the files in this case.

One thing I am not fond of in this PR is that code for `delay_on`/`delay_off` getters and setters is mostly duplicated. We could move the functionality (try to reopen file once on OSError) into `Device._get_attribute` and `Device._set_attribute` to make it universal, but that would introduce slight inefficiency for all attribute getters and setters, when this is only required for leds. On the other hand, this would make our cache a bit more robust. @WasabiFan, @dwalton76, what do you think?

### DIFF
--- a/ev3dev/core.py
+++ b/ev3dev/core.py
@@ -2211,12 +2211,12 @@ class Led(Device):
         # to 'timer', and destroyed when it is set to anything else.
         # This means the file cache may become outdated, and we may have to
         # reopen the file.
-        for i in range(2):
+        for retry in (True, False):
             try:
                 self._delay_on, value = self.get_attr_int(self._delay_on, 'delay_on')
                 return value
             except OSError:
-                if i == 0:
+                if retry:
                     self._delay_on = None
                 else:
                     raise
@@ -2228,12 +2228,12 @@ class Led(Device):
         # to 'timer', and destroyed when it is set to anything else.
         # This means the file cache may become outdated, and we may have to
         # reopen the file.
-        for i in range(2):
+        for retry in (True, False):
             try:
                 self._delay_on = self.set_attr_int(self._delay_on, 'delay_on', value)
                 return
             except OSError:
-                if i == 0:
+                if retry:
                     self._delay_on = None
                 else:
                     raise
@@ -2251,12 +2251,12 @@ class Led(Device):
         # to 'timer', and destroyed when it is set to anything else.
         # This means the file cache may become outdated, and we may have to
         # reopen the file.
-        for i in range(2):
+        for retry in (True, False):
             try:
                 self._delay_off, value = self.get_attr_int(self._delay_off, 'delay_off')
                 return value
             except OSError:
-                if i == 0:
+                if retry:
                     self._delay_off = None
                 else:
                     raise
@@ -2268,12 +2268,12 @@ class Led(Device):
         # to 'timer', and destroyed when it is set to anything else.
         # This means the file cache may become outdated, and we may have to
         # reopen the file.
-        for i in range(2):
+        for retry in (True, False):
             try:
                 self._delay_off = self.set_attr_int(self._delay_off, 'delay_off', value)
                 return
             except OSError:
-                if i == 0:
+                if retry:
                     self._delay_off = None
                 else:
                     raise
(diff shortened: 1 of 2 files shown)


ev3dev/core.py
```diff
@@ -2171,19 +2173,70 @@ def trigger(self):
...
         """
-        self._delay_on, value = self.get_attr_int(self._delay_on, 'delay_on')
-        return value
+
+        # Workaround for ev3dev/ev3dev#225.
+        # 'delay_on' and 'delay_off' attributes are created when trigger is set
+        # to 'timer', and destroyed when it is set to anything else.
+        # This means the file cache may become outdated, and we may have to
+        # reopen the file.
+        for i in range(2):
```
WasabiFan: This construction with the loop and the `if` inside it seems odd. Is there a more concise way of doing this?

ddemidov: I guess this could be turned into a `whileTrue`, but then I would need a boolean variable to distinguish the first run from the second. And `i` does that for me already.

WasabiFan: What about recursion? The issue I have here is that it's a loop that only runs twice, which feels wrong. If it isn't easy to improve, don't waste time on it; this looks good otherwise.

WasabiFan:
I think it makes sense to keep this localized to the LEDs. It would be nice to make it generic to reduce duplication, but I'm fine with this workaround being only here.

ev3dev/core.py
ddemidov: The loop with a try block inside seems to be a pattern in python: http://stackoverflow.com/questions/15232465/how-to-repeat-try-except-block, http://stackoverflow.com/questions/2083987/how-to-retry-after-exception-in-python. And here I just need to retry once, hence the switch from `while True` to `for i in range(2)`. What I could do is switch `for i in range(2)` to `for retry in (True, False)`.

PR #246, opened 2016-11-05
Add FAQ page with info on line endings

WasabiFan:
I _really_ hate reStructured Text. 😑

### DIFF
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -1,8 +1,7 @@
-Frequently-Asked Questions
+requently-Asked Questions
 ==========================
 
-When I try to run my Python file from the shell or Brickman,
-it throws errors or exits immediately
+When I try to run my Python file from the shell or Brickman, it throws errors or exits immediately
 -----
 
 This may occur if your file includes Windows-style line endings, which are often
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -1,0 +1,21 @@
+Frequently-Asked Questions
+==========================
+
+When I try to run my Python file from the shell or Brickman,
+it throws errors or exits immediately
+-----
+
+This may occur if your file includes Windows-style line endings, which are often
+inserted by editors on Windows. To resolve this issue, open an SSH session and
+run the following command, replacing ``<file>`` with the name of the Python file
+you're using:
+
+.. code::
+
+    sed -i 's/\r//g' <file>
+
+This will fix it for the copy of the file on the brick, but if you plan to edit
+it again from Windows you should configure your editor to use Unix-style endings.
+For PyCharm, you can find a guide on doing this `here <https://www.jetbrains.com/help/pycharm/2016.2/configuring-line-separators.html>`_.
+Most other editors have similar options; there may be an option for it in the
+status bar at the bottom of the window or in the menu bar at the top.
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -1,7 +1,7 @@
 requently-Asked Questions
 ==========================
 
-When I try to run my Python file from the shell or Brickman, it throws errors or exits immediately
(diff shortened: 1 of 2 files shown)


docs/faq.rst
```diff
@@ -0,0 +1,21 @@
...
+-----
+
+This may occur if your file includes Windows-style line endings, which are often
+inserted by editors on Windows. To resolve this issue, open an SSH session and
+run the following command, replacing ``<file>`` with the name of the Python file
+you're using:
+
+.. code::
+
+    sed -i 's/\r//g' <file>
```
ddemidov: There used to be a dos2unix utility, not sure if it's available by default on the ev3 though.

WasabiFan: It still exists but is not installed by default, so I decided to go with the built-in option.

ddemidov: :+1:

docs/faq.rst
```diff
@@ -0,0 +1,21 @@
+Frequently-Asked Questions
+==========================
+
+When I try to run my Python file from the shell or Brickman,
+it throws errors or exits immediately
+-----
```
WasabiFan: You can tell I was distracted while writing this :laughing: Three times I wrapped this line, and three times I realized it was broken and fixed it.

WasabiFan:
I couldn't think of a better title for the entry, but I'm not particularly happy with this one. If you have a better idea I'd happily change it.

ddemidov:
> My script works when launched as `python3 script.py` but exits immediately when launched from brickman or as `./script.py`?

Seems as long as the original though.

WasabiFan:
I like that; it explains the issue a bit more explicitly. Look good with that swap?

ddemidov:
http://ev3dev-lang.readthedocs.io/projects/python-ev3dev/en/latest/faq.html

The page title misses an 'F' for me

ddemidov:
Not sure how did I miss it during review :open_mouth:

WasabiFan:
https://github.com/rhempel/ev3dev-lang-python/commit/251b3a71444517d4af3cc22d5834fe7bb7206fec#commitcomment-19706795 Working on a fix

WasabiFan:
It might take me a few hours though 😆

ddemidov:
Take your time, we want it perfect this time

WasabiFan:
I'm trying to link to the FAQ page from the README. How should I do that? Link to the (currently nonexistent) page on `stable`, or link to `latest`? Or maybe to the RST file on GitHub?

ddemidov:
I am starting to think we should point the README docs badge to the latest branch, or maybe provide two badges.

The README ends up as a frontpage for readthedocs, so I don't think we should link to the `latest` docs from the `stable` README. On the other hand, the README that is shown on github is the `latest` one (`develop`). Linking to the RST file github does not seem right as well. May be link to the latest version for now, and switch it to the stable when we are ready to tag 0.8.1? We could add an issue for this and link it to 0.8.1 milestone so that we don't forget.

WasabiFan:
> I am starting to think we should point the README docs badge to the latest branch, or maybe provide two badges.

Honestly, the idea set-up in my eyes would be to have GitHub show the `master` branch by default and everything would be fine, in the sense that having links to `stable` from the development branch is much less of an issue than the inverse because it doesn't affect the average user. However, the argument I've heard against this (which I believe is a good one) is that we want people who contribute code to do it for the `develop` branch. It's a difficult thing to manage over all.

> May be link to the latest version for now, and switch it to the stable when we are ready to tag 0.8.1? We could add an issue for this and link it to 0.8.1 milestone so that we don't forget.

I'm thinking that that is the best option for now. I'll do it.

ddemidov:
> However, the argument I've heard against this (which I believe is a good one) is that we want people who contribute code to do it for the develop branch. 

Now that github allows to switch target branches without reopening the PR, this should be less of an issue.

PR #249, opened 2016-11-06
Suggest using "--only-upgrade" flag

WasabiFan:


dlech:
why?

WasabiFan:
So that if you accidentally type "python-ev3dev" (without the 3) or something similar it will complain at you, assuming we release an image that doesn't have the obsolete package at some point soon.

PR #250, opened 2016-11-07
Update auto-drive.py

DrClick:
fixed bugs in the code. is_pressed is a function, as well as proximity. Also when run_timed, you must set the speed not duty cycle. Thanks again for your work! You had as up and running and a great first day with ev3dev!

### DIFF
--- a/demo/EXPLOR3R/auto-drive.py
+++ b/demo/EXPLOR3R/auto-drive.py
@@ -48,6 +48,8 @@ assert all([m.connected for m in motors]), \
 ir = InfraredSensor(); assert ir.connected
 ts = TouchSensor();    assert ts.connected
 
+print('Robot Starting')
+
 # We will need to check EV3 buttons state.
 btn = Button()
 
@@ -64,9 +66,9 @@ def backup():
     Back away from an obstacle.
     """
 
-    # Sound backup alarm.
+    # Sound backup alarm. 
     Sound.tone([(1000, 500, 500)] * 3)
-
+    
     # Turn backup lights on:
     for light in (Leds.LEFT, Leds.RIGHT):
         Leds.set_color(light, Leds.RED)
@@ -76,7 +78,7 @@ def backup():
     # until both motors are stopped before continuing.
     for m in motors:
         m.stop(stop_action='brake')
-        m.run_timed(duty_cycle_sp=-50, time_sp=1500)
+        m.run_timed(speed_sp=-500, time_sp=1500)
 
     # When motor is stopped, its `state` attribute returns empty list.
     # Wait until both motors are stopped:
@@ -99,7 +101,7 @@ def turn():
     t = randint(250, 750)
 
     for m, p in zip(motors, power):
-        m.run_timed(duty_cycle_sp=p*75, time_sp=t)
+        m.run_timed(speed_sp=p*750, time_sp=t)
 
     # Wait until both motors are stopped:
     while any(m.state for m in motors):
@@ -109,7 +111,7 @@ def turn():
 start()
 while not btn.any():
 
-    if ts.is_pressed:
+    if ts.is_pressed():
         # We bumped an obstacle.
         # Back away, turn and go in other direction.
         backup()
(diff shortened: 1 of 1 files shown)


demo/EXPLOR3R/auto-drive.py
```diff
@@ -109,7 +111,7 @@ def turn():
 start()
 while not btn.any():
 
-    if ts.is_pressed:
+    if ts.is_pressed():
```
WasabiFan: This is actually something that changed in our most recent release, which is not included in the most recent ev3dev image by default. See https://github.com/rhempel/ev3dev-lang-python/releases/tag/0.8.0 for details on that release. If you'd like to upgrade, you can also find instructions on doing that in the README.

demo/EXPLOR3R/auto-drive.py
```diff
@@ -99,7 +101,7 @@ def turn():
     t = randint(250, 750)
 
     for m, p in zip(motors, power):
-        m.run_timed(duty_cycle_sp=p*75, time_sp=t)
+        m.run_timed(speed_sp=p*750, time_sp=t)
```
WasabiFan: Good catch; I guess we didn't fix all cases of `duty_cycle` in the demos after we made that change in the library.

demo/EXPLOR3R/auto-drive.py
```diff
@@ -64,9 +66,9 @@ def backup():
     Back away from an obstacle.
     """
 
-    # Sound backup alarm.
+    # Sound backup alarm. 
     Sound.tone([(1000, 500, 500)] * 3)
-
+
```
WasabiFan: Is this whitespace consistent with what we have in other places? While we're here, we should normalize this.

ddemidov: There should be no trailing spaces in the code.

WasabiFan: @DrClick I think this looks good aside from the whitespace. If you can remove that (both here and two lines above it) I'd say it's ready!

WasabiFan: You're right, it has absolutely no effect on the code's functionality. I ask for those changes mainly because they make the diff easier to read and keep the file uniform. For example, if I find a new bug next week and want to use git's "blame" feature to figure out when this section of code was last modified, I would find that _you_ were the last one to change the code, even though all you did was add some whitespace. Keeping those extraneous changes out of our commit history makes things like that significantly easier.

WasabiFan:
Thanks!

demo/EXPLOR3R/auto-drive.py
```diff
@@ -48,6 +48,8 @@
 ir = InfraredSensor(); assert ir.connected
 ts = TouchSensor();    assert ts.connected
 
+print('Robot Starting')
```
ddemidov: Since this is a demo, we could use this as an opportunity to demonstrate espeak functionality, and change the line to
```py
Sound.speak('Robot starting').wait()
```
It will also be easier to notice when launched from brickman interface.

DrClick: Is there a demo somewhere on how to vim a .sh file and chmod + x it so it can be launched? Or is there some other way to make these scripts executable on the brickman?

DrClick: It not, it would be a valuable edition to the readme.

ddemidov: There could be something about it here:
https://github.com/rhempel/ev3dev-lang-python#user-resources.

WasabiFan: > Since this is a demo, we could use this as an opportunity to demonstrate espeak functionality, and change the line to

I can't say I'm particularly excited to talk at people every time they start his robot, but I suppose it could be a good way to help people discover this functionality.

> Is there a demo somewhere on how to vim a .sh file and chmod + x it so it can be launched? Or is there some other way to make these scripts executable on the brickman?

I believe the README has a section on preparing Python files for execution.

Issue #258, opened 2016-11-18
sound volume not doing anything?


ensonic:
- **ev3dev version:** 4.4.19-15-ev3dev-ev3
- **ev3dev-lang-python version:** 0.8.0

``` python
from ev3dev import ev3
for v in range(0,100,10):
  ev3.Sound.volume = v
  ev3.Sound.tone(400, 50)
```
https://github.com/ev3dev/ev3dev/wiki/Using-Sound#sysfs-attributes-and-module-parameters

This probably came from the transition to use the beep command, which does not have a volume parameter.

Should we shell out the same way and call, e.g.
```
amixer -q set Playback 50%
```

ddemidov:
Currently `Sound` is just a convenience class that wraps `espeak`, `aplay` and `beep` utilities. It does not inherit from `Device`, so `Sound.volume = v` does nothing except for creating a class level field named `volume`.

Did you try writing to `/sys/devices/platform/snd-legoev3/volume` directly? Does that work? In case it does, we could easily introduce the volume attribute to the `Sound` class.

dlech:
FYI, I don't intend to keep `/sys/devices/platform/snd-legoev3/volume` around forever, so I would not suggest becoming reliant on it. Better to use the ALSA mixer instead.

ensonic:
yep, this works too:

     echo "10" | sudo tee /sys/devices/platform/snd-legoev3/volume

I thought using amixer would be more portable to e.g. brickpi. But on the other hand writing to the sysfs is probably faster than spawning a binary.
I am bringing this up, because this worked in the past before we switched to the beep command.

dlech:
There is also [python-pyalsa](https://packages.debian.org/jessie/python-pyalsa) and [python-alsaaudio](https://packages.debian.org/jessie/python-alsaaudio) but these don't have python3 packages.

ddemidov:
See #259

PR #259, opened 2016-11-18
Provide Sound.set_volume(pct)

ddemidov:
See #258 

This uses a call to `amixer`, as opposed to using a sysfs attribute, since @dlech [hinted](https://github.com/rhempel/ev3dev-lang-python/issues/258#issuecomment-261574133) that could disappear in the future. Besides, every other method in the `Sound` class is a call to an executable, so why break the tradition?

### DIFF
--- a/ev3dev/core.py
+++ b/ev3dev/core.py
@@ -3054,7 +3054,7 @@ class Sound:
         """
         Sets the sound volume to the given percentage [0-100] by calling
         ``amixer -q set <channel> <pct>%``.
-        If channel is not specified, it tries to determine the default channel
+        If the channel is not specified, it tries to determine the default one
         by running ``amixer scontrols``. If that fails as well, it uses the
         ``Playback`` channel, as that is the only channel on the EV3.
         """
--- a/ev3dev/core.py
+++ b/ev3dev/core.py
@@ -50,7 +50,7 @@ import select
 import time
 from os.path import abspath
 from struct import pack, unpack
-from subprocess import Popen
+from subprocess import Popen, check_output
 
 try:
     # This is a linux-specific module.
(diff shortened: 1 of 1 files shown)


ddemidov:
@ensonic, would this work for you?

WasabiFan:
Actually, I suspect that the main device may be called something other than "Playback" on other ev3dev platforms. I know it was on my desktop PC last time I messed around with ALSA. @ensonic can you confirm?

ddemidov:
I checked, and this works on the EV3. The Playback device is indeed missing on my PC though. @dlech, do you know if this would be portable across all ev3dev-supported platforms?

dlech:
No, it won't be portable. You could have a parameter with a default value of "Playback" that could be changed for other platforms.

WasabiFan:
I think we should avoid exposing any details (e.g. the element name) to the user. If we can't do that with this method, I think we should either make it EV3-only or use a Python library and choose a primary device with the lower-level interface.

ev3dev/core.py
```diff
@@ -3048,3 +3048,9 @@ def speak(text, espeak_opts='-a 200 -s 130'):
         with open(os.devnull, 'w') as n:
             cmd_line = '/usr/bin/espeak --stdout {0} "{1}" | /usr/bin/aplay -q'.format(espeak_opts, text)
             return Popen(cmd_line, stdout=n, shell=True)
+
+    @staticmethod
+    def set_volume(pct):
+        with open(os.devnull, 'w') as n:
+            cmd_line = '/usr/bin/amixer -q set Playback {0:d}%'.format(pct)
```
ensonic: do you need the devnull part? That is what the "-q" is for.

ensonic:
For other platforms it is usually called "Master". I think a default that is overridable would work.

ddemidov:
I've updated the PR. `Sound.set_volume()` now takes optional channel name. When the parameter is omitted, the first channel that pops up in `amixer scontrol` output is used. If that fails as well, default value of `'Playback'` is used.

WasabiFan:
Looks good! Can you add a bit of documentation to this method explaining the behavior?

ddemidov:
Done.

WasabiFan:
Does that comment apply to the method itself such that it would show up in the docs?

ddemidov:
Yes, any string that is defined immediately after method specification is used for auto documentation.

ensonic:
Should provide a similar method to read the volume? This would be calling

    /usr/bin/amixer get "Master"

but we'd need to parse the output.

ddemidov:
@ensonic, see #266

PR #266, opened 2016-11-25
Implement Sound.get_volume()

ddemidov:
See #259

### DIFF
--- a/ev3dev/core.py
+++ b/ev3dev/core.py
@@ -2966,6 +2966,8 @@ class Sound:
 
     """
 
+    channel = None
+
     @staticmethod
     def beep(args=''):
         """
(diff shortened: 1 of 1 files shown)


ddemidov:
There is `alsaaudio` package available for python3 on my Arch linux machine. It allows to do simple things like
```py
import alsaaudio
m = alsaaudio.Mixer()
print(m.getvolume())
```

So the package is compatible with python3, it just is [not available](http://packages.ubuntu.com/search?keywords=alsaaudio&searchon=names) as a python3 package for Ubuntu. @dlech, would it make sense to build the `python3-alsaaudio` package for ev3dev?

ensonic:
Thanks! If we leave it to just set/get the main volume, this is probably okay. I am already a bit concerned with the time the import of the ev3dev package takes.

PR #267, opened 2016-11-27
utils move_motor and stop_motor update to python3

dwalton76:


WasabiFan:
Hmmmmm... I didn't notice this earlier, but it looks like there are stray commits here. What happened with those? Did you base this off of a previous PR branch that was squashed?

dwalton76:
I am scratching my head on why those are there.  This is the same branch where I wrote the EV3D4 README and I did a pull request from this branch that was accepted.  Following that I did a "git pull github.com/rhempel/ev3dev-lang-python develop".

WasabiFan:
Ah, yes; then the problem is that those commits from the previous PR were squashed, meaning that a new commit was created with their combined changes, yet those old commits were included here. I don't think it will cause any issues, but for the future:

- Always start PR branches from the branch that you will be setting as your base (`develop`).
- Merge PRs in "squash" mode unless in extreme circumstances. This keeps the history clean and will avoid issues like this.

dwalton76:
'Merge PRs in "squash" mode unless in extreme circumstances' how do you do that via the github UI?  Maybe a better question is "can we make this the default behavior for this repository?"

WasabiFan:
> how do you do that via the github UI?

There's a drop-down in the merge interface that lets you choose the mode. It remembers that choice for future uses.

> can we make this the default behavior for this repository?

I believe you can do that (or at least disable other options) however we need a repository administrator to modify those settings. I'm not an admin though, so I can't make that change.

dwalton76:
Same here (I am not an admin for the repo). paging @rhempel :)

Issue #269, opened 2016-11-30
Move HTTP code in helper.py to a separate file


dwalton76:
A basic python3 hello_world.py program that imports helper.py takes about 19s to run on my EV3.  Some digging shows that a large chunk of this time is spent in "from http.server import BaseHTTPRequestHandler, HTTPServer".  I moved all of the HTTP code in helper.py to a www.py and my hello_world.py now runs in 9s.

9s still seems like too long but it is a lot better than 19s.  Does www.py sound ok?

WasabiFan:
I'm not sure that "www.py" is accurate as a name; it isn't actually World Wide Web code, it's an HTTP server intended to be used locally. I'd prefer something like "web-server.py". Good changes otherwise.

dwalton76:
Will do webserver.py (dashes in filenames are problematic when importing)

PR #271, opened 2016-12-01
Move HTTP code in helper.py to a separate file 

dwalton76:
Issue #269


demo/EV3D4/README.md
```diff
@@ -1,21 +1,18 @@
 # EV3D4
 EV3D4 is designed to look like R2-D2 from Star Wars. There are two options for
 controlling EV3D4. The first is to use the IR remote to send commands to the IR
-sensor, run EV3D4RemoteControl to use this method. The second means of
-controlling EV3D4 is via a web browser, run EV3D4WebControl to use this method.
+sensor, run EV3D4RemoteControl.py to use this method. The second means of
```
WasabiFan: Can we get a quick explanation of how to "run <file>"? Either instructions to `chmod +x <file>` and run from Brickman or to do it from the command line.

demo/EV3D4/README.md
```diff
@@ -1,21 +1,20 @@
...
+controlling EV3D4 is via a web browser, run EV3D4WebControl.py to use this method.
+You can run both of these from the Brickman interface or if logged in via ssh
```
WasabiFan: Keep in mind that both of these methods need the executable bit to be set on the file. Even if it _is_ set in Git, if they copy the file to their brick it likely will lose the mode. I think you need a `chmod` here.

dwalton76: I guess I am going on the assumption that they git clone this repo....isn't that how everyone is getting these demo programs since we do not install them as part of the package?

dwalton76: What do you say we address how to do this somewhere other than the EV3D4 README?  We do not have 'run' or chmod instructions in the READMEs of the other folders in demo.

WasabiFan: Yeah, I think that sounds like a better plan. Can we add a subsection about this in the main README within the part about creating Python files? You can then just reference that from here.

dwalton76: I will go ahead and merge this one (the moving of the web server code from helper.py to webserver.py seems ok) and do a separate pull request that addresses:
- How do I install the demo programs on my EV3?
- How do I run the demo programs on my EV3?

Issue #270, opened 2016-11-30
Remove kociemba and rubiks_rgb_solver.py from MINDCUB3R demo


dwalton76:
I forked kociemba (sent a pull request to make it python3 compatible) and created a repo for the rubiks cube RGB analyzing code.  We can now remove these from demo/MINDCUB3R and instead point the user to install them via:
```
sudo apt-get install build-essential python3-pip
sudo pip3 install git+https://github.com/dwalton76/kociemba.git
sudo pip3 install git+https://github.com/dwalton76/rubiks-color-resolver.git
```

@WasabiFan @ddemidov sound good?

WasabiFan:
Sounds good to me!

PR #272, opened 2016-12-01
Remove kociemba and rubiks_rgb_solver.py from MINDCUB3R demo

dwalton76:
This is for issue #270


demo/MINDCUB3R/README.md
```diff
@@ -1,26 +1,13 @@
...
-
-* cd ~/ev3dev-lang-python/demo/MINDCUB3R/
-* kociemba DRLUUBFBRBLURRLRUBLRDDFDLFUFUFFDBRDUBRUFLLFDDBFLUBLRBD
-
+This can take several minutes on an EV3
+```
+sudo apt-get install build-essential python3-pip libffi-dev
+sudo pip3 install git+https://github.com/dwalton76/kociemba.git
+sudo pip3 install git+https://github.com/dwalton76/rubiks-color-resolver.git
+kociemba DRLUUBFBRBLURRLRUBLRDDFDLFUFUFFDBRDUBRUFLLFDDBFLUBLRBD
```
WasabiFan: This seems like gibberish :laughing: Is the idea here that you're feeding it a series of motions for testing purposes? If so, can you add a comment (not necessarily a code comment, just a note) with expected output?

WasabiFan:
Is this confirmed at least seeming to work (motors spin) on the brick?

demo/MINDCUB3R/README.md
dwalton76: It does look like greek doesn't it :)  The string is a representation of the colors of each of the 54 squares of a 3x3x3 cube.  So the D at the beginning means that square #1 is the same color as the middle square of the Down side (the bottom), the R means that square #2 is the same color as the middle square of the Right side, etc.  The kociemba program takes that color data and returns a sequence of moves that can be used to solve the cube.  Just to keep things consistent the syntax for the sequence of moves also looks like greek
```
robot@beaglebone[lego-crane-cuber]# kociemba DRLUUBFBRBLURRLRUBLRDDFDLFUFUFFDBRDUBRUFLLFDDBFLUBLRBD
D2 R' D' F2 B D R2 D2 R' F2 D' F2 U' B2 L2 U2 D R2 U
robot@beaglebone[lego-crane-cuber]# 
```

I'll add some of this text to the README

dwalton76: I have that step in the install process because the first time you run it, it takes about 30 seconds because it builds some tables that it caches to the filesystem.  After that first run it is nice and fast.

demo/MINDCUB3R/README.md
```diff
@@ -1,26 +1,34 @@
...
-https://github.com/cavenel/ev3dev_examples
-
-The kociemba code is from
-https://github.com/muodov/kociemba
+### About kociemba
+You may have noticed that the
+`kociemba DRLUUBFBRBLURRLRUBLRDDFDLFUFUFFDBRDUBRUFLLFDDBFLUBLRBD`
+step of the install looks a little odd. The "DRLUU..." string is a
+representation of the colors of each of the 54 squares of a 3x3x3 cube. So
+the D at the beginning means that square #1 is the same color as the middle
```
WasabiFan: This will probably auto-link to the GitHub issue #1; you can put it in code formatting if you want or just leave it as-is.

PR #278, opened 2016-12-21
Error on non-existing arguments/parameters

pepijndevos:
This patch adds `__slots__` to all API classes that use `setattr` magic and `kwargs` for function parameters. Besides reducing memory and speeding up access, the main point of this change is that a typo when calling a function will result in an error, rather than silently setting non-existing properties.

```
In [1]: import ev3dev.ev3 as ev3
In [2]: m = ev3.MediumMotor()
In [3]: m.run_to_abs_pos(position_sp=72*20, speed_sd=500, stop_action='coast')
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-4-08bcba75fb47> in <module>()
----> 1 m.run_to_abs_pos(position_sp=72*20, speed_sd=500, stop_action='coast')

/home/robot/ev3dev-lang-python/ev3dev/core.py in run_to_abs_pos(self, **kwargs)
    799         """
    800         for key in kwargs:
--> 801             setattr(self, key, kwargs[key])
    802         self.command = self.COMMAND_RUN_TO_ABS_POS
    803 

AttributeError: 'MediumMotor' object has no attribute 'speed_sd'
```
Currently the last line would just set the `speed_sd` property on the object, without doing anything, leaving it up to you to figure out why your motor went berserk or did nothing at all.

With the proposed change, you get instant feedback that you're doing something stupid.

### DIFF
--- a/ev3dev/core.py
+++ b/ev3dev/core.py
@@ -960,6 +960,12 @@ class LargeMotor(Motor):
 
 
 # ~autogen
+    __slots__ = [
+# ~autogen generic-class-slots classes.largeMotor>currentClass
+
+
+# ~autogen
+    ]
 # ~autogen generic-class classes.mediumMotor>currentClass
 
 class MediumMotor(Motor):
@@ -977,6 +983,12 @@ class MediumMotor(Motor):
 
 
 # ~autogen
+    __slots__ = [
+# ~autogen generic-class-slots classes.mediumMotor>currentClass
+
+
+# ~autogen
+    ]
 # ~autogen generic-class classes.actuonix50Motor>currentClass
 
 class ActuonixL1250Motor(Motor):
@@ -994,6 +1006,12 @@ class ActuonixL1250Motor(Motor):
 
 
 # ~autogen
+    __slots__ = [
+# ~autogen generic-class-slots classes.actuonix50Motor>currentClass
+
+
+# ~autogen
+    ]
 # ~autogen generic-class classes.actuonix100Motor>currentClass
 
 class ActuonixL12100Motor(Motor):
@@ -1011,6 +1029,12 @@ class ActuonixL12100Motor(Motor):
 
 
 # ~autogen
+    __slots__ = [
+# ~autogen generic-class-slots classes.actuonix100Motor>currentClass
+
+
+# ~autogen
+    ]
 # ~autogen generic-class classes.dcMotor>currentClass
 
 class DcMotor(Device):
--- a/ev3dev/core.py
+++ b/ev3dev/core.py
@@ -1798,6 +1798,8 @@ class TouchSensor(Sensor):
     Touch Sensor
     """
 
+    __slots__ = ['auto_mode']
+
     SYSTEM_CLASS_NAME = Sensor.SYSTEM_CLASS_NAME
     SYSTEM_DEVICE_NAME_CONVENTION = Sensor.SYSTEM_DEVICE_NAME_CONVENTION
 
@@ -1827,6 +1829,8 @@ class ColorSensor(Sensor):
     """
     LEGO EV3 color sensor.
     """
+
+    __slots__ = ['auto_mode']
 
     SYSTEM_CLASS_NAME = Sensor.SYSTEM_CLASS_NAME
     SYSTEM_DEVICE_NAME_CONVENTION = Sensor.SYSTEM_DEVICE_NAME_CONVENTION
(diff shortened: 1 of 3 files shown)


pepijndevos:
What is `_bin_data_size`, what is it used for, and what is that piece of code trying to do? Obviously a class with `__slots__` has no `__dict__`, but it seems adding it to `__slots__` and checking for `None` would have the same effect? The property is not used anywhere else in the project, so maybe it could also just be a local variable?

ddemidov:
> What is _bin_data_size, what is it used for, and what is that piece of code trying to do? 

`_bin_data_size` stores the size of raw binary data to read from the sensor (see `bin_data` [here](http://www.ev3dev.org/docs/drivers/lego-sensor-class/)).

[In the code](https://github.com/rhempel/ev3dev-lang-python/blob/488b89f169c685be26799ae9b73df99f74c6a89f/ev3dev/core.py#L1632-L1641) the field is used as a cache to store the value (and save an access to `bin_data_format` attribute). Initializing it with `None` should work as well.

ensonic:
If this makes property access faster it would be nice to see some timeit before and after. Could be a worthy addition to the release notes, since speed improvements help to cut latency on control loops.

dwalton76:
yep I will give it a try

pepijndevos:
I'm not sure what a good test is to test performance of this change, TBH probably most time is spend talking to the kernel, compared to one dictionary lookup.

I fixed the data size, which was rather straightforward.

The helper classes require a bit more thought. `__slots__` don't mix well with multiple inheritance. The internet consensus seems to be that if you run into this you're using `__slots__` wrong. It is suggested that they are a memory optimisation, not a means to "freeze" classes. The "Pythonic" way that is suggested is to overwrite `__setattr__`, which would make things slower, if anything, though the impact is probably manageable.

As it stands now, using the helper classes would add an internal dict, allowing undefined attributes as before.

dwalton76:
I think this one needs some more testing. I cannot pinpoint what is wrong but when using this patch my robot failed to solve the rubiks cube three times in a row, without this patch it can solve it everytime.  All motors look like they are rotating the correct amount (the cube would get jammed up otherwise) but I suspect one is turning in the wrong direction in some scenario.

pepijndevos:
Weird. No errors? (or things that swallow errors) If the robot is deterministic, I suppose you could use the same scramble and see where it goes wrong.‎ Or just plain old printf debugging.Ik might be able to do some more testing, but simple things seemed to work and without knowing more it's hard to know where to look.Is this robot of yours one I can reproduce with my own ev3 (software and instructions available), or a custom monster with tons of extra parts?

ddemidov:
> Is this robot of yours one I can reproduce with my own ev3 (software and instructions available), or a custom monster with tons of extra parts?

This is not my robot, but I believe its built according to these instructions: http://mindcuber.com/mindcub3r/mindcub3r.html.

And the program is here: https://github.com/rhempel/ev3dev-lang-python/tree/develop/demo/MINDCUB3R

dwalton76:
MINDCUBER is a good way to test, a lot of things have to be perfect for it to solve the cube.  I'm actually using my own cube solver now, if you have a 42009 (the big crane) then you can build it.  I need to put together instructions though.

Let me do this, I can put the cube back in the same scrambled state and record a video of it being solved with and without your patch.  Then I can compare the two videos side by side and see where it is behaving differently. That should help narrow it down a bit.

dwalton76:
Ignore what I said, I suspect I have an error on my end.

ddemidov:
A simple performance test shows there is a bit of improvement, or at least it is definitely not worse:

**Before**
```py
>>> import timeit
>>> timeit.repeat('s.proximity', 'from ev3dev.ev3 import InfraredSensor; s = InfraredSensor()', number=100)
[4.144548124999972, 3.950399667000056, 4.128440291999993]
>>> timeit.repeat('s.proximity', 'from ev3dev.ev3 import InfraredSensor; s = InfraredSensor()', number=100)
[3.9438115419999917, 4.096679416999962, 3.8801820420000013]
>>> 
```

**After**
```py
>>> import timeit
>>> timeit.repeat('s.proximity', 'from ev3dev.ev3 import InfraredSensor; s = InfraredSensor()', number=100)
[3.8830312090000234, 4.003896959000031, 3.78919083400001]
>>> timeit.repeat('s.proximity', 'from ev3dev.ev3 import InfraredSensor; s = InfraredSensor()', number=100)
[3.8567800419999685, 3.932827876000033, 3.8972169590000476]
>>> 
```

But I think the main idea here was to help users by catching semantic errors early.

ddemidov:
Thanks!

pepijndevos:
Awesome. Still curious what happened on @dwalton76 's end though.

dwalton76:
I have a bug where I sometimes assign the wrong color to a cube square based on the RGB values. In a nutshell you have 54 RGB tuples and from that you have to assign each square to be one of the six colors of the Rubik's cube. I had this fairly rock solid but have been working on this code to make it support 4x4x4, 5x5x5, etc cubes and must have introduced a bug because when I took a video of it solving with your patch in place I noticed that it got some of the color assignments wrong.

dwalton76:
In another run I was able to solve a cube with your patch applied so the motor movements seem to be fine.

PR #285, opened 2017-01-26
Update for latest RubiksColorSolver and update README

dwalton76:


dwalton76:
I have someone that confirmed all is working with MINDCUB3R so am going to go ahead and close this since in its current state it is broken.

PR #300, opened 2017-02-12
music module added

EricPobot:
Hello,

This module adds a function which makes easier to play songs by writing them using the conventional musical notation (note and duration). A basic demo illustrating it (and a couple of the other sound related   ev3dev features) is added in the demo tree.


ddemidov:
This looks nice, but does it really need to be in a separate module? Can we add this as a method to the `Sound` class? Moreover, since this is just another way to control the `beep` command, I think the current functionality of [`Sound.tone`](https://github.com/rhempel/ev3dev-lang-python/blob/develop/ev3dev/core.py#L3178) could be merged with the new functionality of `play_song`. Something like `Sound.tone(sequence, notation='beep')` or `Sound.tone(sequence, notation='music')`.

ericpascual:
> does it really need to be in a separate module?

I agree with you David that it should not. This code was initially in my modified version of the Sound class. But since I don't master the template mechanism (I think you remember some discussions around this topic some time ago ;) I couldn't find how to implement properly the extension this way. Hence this separate module version.

If you give me the right way to proceed for integrating the code in the Sound class module, I can move it at the right place.

>  the current functionality of Sound.tone could be merged with the new functionality

I confess not being convinced by such a merge of concerns. I prefer methods having a unique and well defined purpose which can be clearly identified by the name and signature. 

In our case we have on one hand a method for playing a succession of tones (for instance imitating a phone number dialing) and on the other hand a method for playing a song, based on its music score. Merging both in a single entry point which behaviour and purpose change based on the value of a parameter can be a bit confusing for the average user IMHO.

ddemidov:
> But since I don't master the template mechanism (I think you remember some discussions around this topic some time ago ;) 

I do remember that discassion :smiley_cat:.  But it so happens the `Sound` class is implemented without use of templates, so it should be easy to extend it as is.

> I confess not being convinced by such a merge of concerns. I prefer methods having a unique and well defined purpose which can be clearly identified by the name and signature.

Ok, I am not insisting on this.

ensonic:
If we want to play songs, it would be nice to add a paser that translates RTTL (https://en.wikipedia.org/wiki/Ring_Tone_Transfer_Language) to beep commands. There are lots of RTTL files online

ericpascual:
>  But it so happens the Sound class is implemented without use of templates

OK. Since the module it lives in contains markers identifying generated sections, I was thinking that it should not be edited directly. 

Can I modify it directly without any special consideration, apart not touching code inside the marked areas (which are not involved here as you said) ?

_PS: sorry for using my alternate identity for these recent comments. I'm working on professional projects at the same time from this machine, and it's a bit teddious to switch continuously between both ;-)_

ericpascual:
@ensonic 
Nice suggestion. I'll give it a look.

ddemidov:
> Can I modify it directly without any special consideration, apart not touching code inside the marked areas (which are not involved here as you said) ?

Yes, absolutely. As long as the code you are touching is outside of `# ~autogen` marker pairs, it should be fine.

ericpascual:
Fine. I'll move the code in the Sound class then and kill this branch.

ddemidov:
You can force-push new commits to this branch to update the pull request, so there is really no need to kill the branch.

ericpascual:
Hmmm, I'll need to document myself about `force-push`. Never used this before.

ddemidov:
You can basically rewrite the history of your local branch with `git commit --amend` (for the latest commit), or with `git rebase -i origin/develop` (here you can drop commits that you don't need anymore or squash several commits together). Then you can push the updated branch to github with `git push -f`.

ericpascual:
Thanks for the help. I'll try to find my way with these indications ;-)

WasabiFan:
There's no need to rebase or force-push; when a maintainer merges this PR, they can have GitHub combine all the commits into one to clean it up.

ddemidov:
```
- Give me some prepared tablets of acetylsalicylic acid.
- Do you mean aspirin?
- That’s it! I can never remember the name.
```

ericpascual:
Ahem... So which operating mode and command(s) a Git noob like me is supposed to use exactly to avoid creating a mess ? 

It it just : 
- edit the Sound class module, 
- delete the music module I've added, 
- commit and push the result in the same branch ?

ddemidov:
Commit and push until you are satisfied with the results.

ericpascual:
Fine. This should be within my competences then :-) I'll manage it this evening.

I guess I'll have to issue a new PR when done. Right ?

WasabiFan:
Nope, when you push to the existing branch this PR will automatically update. All you need to do is continue making changes until you're happy!

ericpascual:
When will the merger be informed that I've finished with changes and that he can merge the branch into the mainstream ? Let's say that the merge is done some day, what happens if I make more pushes after that point ?

Thanks all for your patience and for making me less stupid (if possible) :-)

WasabiFan:
> When will the merger be informed that I've finished with changes and that he can merge the branch into the mainstream ? 

Anyone watching the repo is informed when you push additional commits. Until you push, they are all on your local machine. So, generally, a maintainer knows to look at it again if you push some commits that appear to attempt to respond to all feedback or you post a comment saying that you have made the changes. Generally a maintainer will submit a review indicating whether they accept the PR or request changes.

> Let's say that the merge is done some day, what happens if I make more pushes after that point ?

Once you have merged a PR, it "detaches" from its branch. That means that future commits to that branch do not affect the PR or other branches. If you wanted to, you could later open a new PR from that same branch to integrate changes since your last. However, that can be problematic because often the maintainer chooses to take your changes and squash them into one commit, meaning that the commits in your branch no longer are the same ones in the master branch, even though they contain the same changes.

ericpascual:
Thanks a lot WasabiFan for your time. Your explanations are crystal clear. 

From the last sentence, I understand that once the PR has been merged, it is safer to close the branch, and start a new one from the version of the moment if some additional work is to be done later. This would avoid the commits discrepency problem you mention.

WasabiFan:
Yep, that's right.

```bash
# if necessary, "git remote add upstream https://github.com/rhempel/ev3dev-lang-python
git checkout master
git pull upstream master
git checkout --branch my-new-feature
```

ericpascual:
Perfect. Thanks again for this "Git for the dummies" improvised course :)

ev3dev/core.py
```diff
@@ -3300,3 +3300,254 @@ def get_volume(channel=None):
...
+            res[n] = freq
+    return res
+
+#: Note frequencies.
+#:
+#: This dictionary gives the rounded frequency of a note specified by its
+#: standard US abbreviation and its octave number (e.g. ``C3``).
+#: Alterations use the ``#`` and ``b`` symbols, respectively for
+#: *sharp* and *flat*, between the note code and the octave number (e.g. ``D#4``, ``Gb5``).
+NOTE_FREQUENCIES = _make_scales((
```
ddemidov: Do this dictionary and the `_make_scales()` function above have use outside of the `Sound` class? I mean, would it make sense to move them into the class scope?

ev3dev/core.py
```diff
@@ -3300,3 +3300,254 @@ def get_volume(channel=None):
...
+#:  e eight note (UK: quaver)
+#:  s sixteenth note (UK: semiquaver)
+#:  = ===============================
+#:
+#:
+#: Triplets can be obtained by dividing the corresponding reference by 3.
+#: For instance, the note value of a eight triplet will be ``NOTE_VALUE['e'] / 3``.
+#: It is simpler however to user the ``3`` modifier of notes, as supported by the
+#: :py:meth:`Sound.play_song` method.
+NOTE_VALUES = {
```
ddemidov: Should this be moved into `Sound` class as well?

ddemidov:
This works:
```py
class C(object):
      @staticmethod
      def foo():
          pass
      STATIC_VAR = C.foo()
```

ddemidov:
Ah, my bad. I was experimenting in an interactive shell, and the `C` I got was a leftover from previous attempt. So, I was wrong, and it does not work.

ddemidov:
So it seems we do have to keep `_make_scales` out of the class scope. I would still prefer to move `NOTE_FREQUENCIES` and `NOTE_VALUES` into the class definition, but I'll leave those to your choice.

ericpascual:
No problem for refactoring the code the way you prefer. I'll take care of this ASAP.

ddemidov:
Thanks!

PR #306, opened 2017-02-17
Document class-scope constants

### LINKED ISSUE #299: Document string constants
It looks like there isn't any documentation on the constants that we define for commands, states, polarities and other string-based properties. What must be changed to make them appear in the rendered docs?

ddemidov:
Closes #299.

The generated documentation: http://ev3dev-lang.readthedocs.io/projects/python-ev3dev/en/issue-229

Motor classes got a lot of constants: http://ev3dev-lang.readthedocs.io/projects/python-ev3dev/en/issue-229/motors.html#tacho-motor

### DIFF
--- a/ev3dev/core.py
+++ b/ev3dev/core.py
@@ -709,77 +709,77 @@ class Motor(Device):
 # ~autogen
 # ~autogen generic-property-value classes.motor>currentClass
 
-    # Run the motor until another command is sent.
+    #: Run the motor until another command is sent.
     COMMAND_RUN_FOREVER = 'run-forever'
 
-    # Run to an absolute position specified by `position_sp` and then
-    # stop using the action specified in `stop_action`.
+    #: Run to an absolute position specified by `position_sp` and then
+    #: stop using the action specified in `stop_action`.
     COMMAND_RUN_TO_ABS_POS = 'run-to-abs-pos'
 
-    # Run to a position relative to the current `position` value.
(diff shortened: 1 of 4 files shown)


ddemidov:
Please don't merge yet, there are some platform-specific constants to do.

WasabiFan:
I may be incompetent when reviewing at 5:00, but at 2:00 I think I have a handle on things :grin:

ddemidov:
I believe I am done. One thing I am not sure about is the documentation for Leds class: http://ev3dev-lang.readthedocs.io/projects/python-ev3dev/en/issue-229/other.html#ev3dev.ev3.Leds. It has manually entered lists for EV3 and for BrickPi platforms, and the documentation generator also picked up constants for the EV3 platform.

I don't think we can get auto-documentation for both platforms, so may be its best to leave the manually entered section as is.

WasabiFan:
> may be its best to leave the manually entered section as is.

I'd agree with that. Honestly, if someone can't figure out what a constant named `GREEN` means in an LED class, they aren't going to get far anyway.

ddemidov:
So, remove the autodoc strings, or keep both?

WasabiFan:
I vote to keep the manual stuff as-is, and put documentation everywhere else (i.e. the common bits),

ddemidov:
Ok, I've stripped led-related commits from the PR.

PR #308, opened 2017-02-18
Sound internal tables made private

EricPobot:
Generated documentation checked for Sound internal tables not been included any more.

Some documentation update included too.

### DIFF
--- a/ev3dev/core.py
+++ b/ev3dev/core.py
@@ -3347,7 +3347,7 @@ class Sound:
 
         It supports symbolic notes (e.g. ``A4``, ``D#3``, ``Gb5``) and durations (e.g. ``q``, ``h``).
 
-        For an exhaustive liste of accepted note symbols and values, have a look at the :py:attr:`_NOTE_FREQUENCIES`
+        For an exhaustive list of accepted note symbols and values, have a look at the :py:attr:`_NOTE_FREQUENCIES`
         and :py:attr:`_NOTE_VALUES` private dictionaries in the source code.
 
         The value can be suffixed by modifiers:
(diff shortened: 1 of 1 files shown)


ev3dev/core.py
```diff
@@ -3339,8 +3347,10 @@ def play_song(cls, song, tempo=120, delay=50):
 
         It supports symbolic notes (e.g. ``A4``, ``D#3``, ``Gb5``) and durations (e.g. ``q``, ``h``).
 
-        Accepted note symbols and values are defined by the :py:attr:`NOTES` and :py:attr:`NOTE_VALUES`
-        dictionaries. The value can be suffixed by modifiers:
+        For an exhaustive liste of accepted note symbols and values, have a look at the :py:attr:`_NOTE_FREQUENCIES`
```
ddemidov: typo: 'liste'

ddemidov:
> Side note: how do you generate the documentation ?

I don't. readthedocs.com picks up new commits and rebuilds the documentation there. The documentation for the latest commit in `develop` branch is here: http://ev3dev-lang.readthedocs.io/projects/python-ev3dev/en/develop/.

But the following should work if you need to build a local copy:
```
cd docs
make html
```
It needs some python packages to be installed, one that I remember of is `sphinx-bootstrap-theme`.

ddemidov:
I may be in slight advantage, since `/usr/bin/python` on Arch linux is `python3`.

ev3dev/core.py
EricPobot: Sorry. Fixed.

ddemidov:
Thanks!

ddemidov:
And readthedocs.com picked this already: http://python-ev3dev.readthedocs.io/en/develop/other.html#sound

ddemidov:
If the workaround works for all cases, then sure.

ddemidov:
readthedocs.org do not use the `Makefile`, the only thing that matters for them is `conf.py`.

It looks like the workaround may work for the case when `python3` is the default one. I can test this if you make the pull request.

Issue #302, opened 2017-02-13
Make the special sensor class for the Infrared Sensor include IR-SEEK


fisherds:
- **ev3dev version:**  4.4.24-16-ev3dev-ev3
- **ev3dev-lang-python version:**0.8.0

If you look at the special sensor classes for the Color Sensor and the Infrared Sensor you'll see the Color Sensor has a very robust special sensor class:
http://python-ev3dev.readthedocs.io/en/latest/sensors.html#special-sensor-classes

However the Infrared Sensor has only the one property: proximity.  There is no support for the IR-SEEK mode within the special sensor class.  You can of course use the sensor values directly if you read the sensor page:
http://www.ev3dev.org/docs/sensors/lego-ev3-infrared-sensor/

Your code looks like this, which isn't terrible.

            ir_sensor = ev3.InfraredSensor()
            ir_sensor.mode = "IR-SEEK"
            current_distance_channel1 = ir_sensor.value(0)
            current_heading_channel1 = ir_sensor.value(1)

Two issues:

1. This approach is less elegant than first class properties / methods on the special sensor class.  It makes IR-SEEK harder for new people to discover and makes the code feel more manual.
Maybe we could add a method to call

            current_distance_channel1 = ir_sensor.beacon_distance(1)
            current_heading_channel1 = ir_sensor.beacon_heading(1)

or a bunch of properties (8 properties):

            current_distance_channel1 = ir_sensor.beacon_distance_ch1
            current_heading_channel1 = ir_sensor.beacon_heading_ch1

2. This other problem I've noticed with the manual approach is that it's a bit buggy.  It has errors sometimes when there is difficulty communicating with the sensor.  If fires errors often enough that I've started wrapping calls in try: expect: blocks just in case (can happen when setting the mode too).  If you run enough tests you'll see an error thrown too.  I'm guessing there is a bit of protection built in with things like .proximity because I never see errors with them.  So I'd like the elegance of adding IR-SEEK and the protection from errors that seems to happen with .proximity. :)

Yet again happy to let others drive, just putting in the feature request.

ddemidov:
[...]

PR #311, opened 2017-02-28
Implement BeaconSeeker class

ddemidov:
See #302.

@fisherds, I went with separate class as opposed to adding the `heading` and `distance` methods to the `InfraredSensor` class. The reason is that the remote controller has four channels. With the separate `BeaconSeeker` class we can set the channel in the class constructor, and implement `heading/distance` as properties:
```py
beacon = BeaconSeeker(channel=1)
print( beacon.heading_and_distance )
```
This is very similar to the way the `RemoteControl` class works.

If we extended the `InfraredSensor` class instead, we would have to specify channel on each call to `heading/distance`:
```py
s = InfraredSensor()
print( s.beacon_heading(channel=1) )
```

What do you think?

### DIFF
--- a/docs/other.rst
+++ b/docs/other.rst
@@ -22,6 +22,13 @@ Remote Control
 
     .. rubric:: Member functions and properties
 
+Beacon Seeker
+-------------
+
+.. autoclass:: BeaconSeeker
+    :members:
+    :inherited-members:
+
 Button
 ------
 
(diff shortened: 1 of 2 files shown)


fisherds:
Yeah setting it once in the constructor is much cleaner.  Totally on board with your approach now.  Once you look at the resulting code I love the way you did it.  Awesome!  Thanks @ddemidov

ddemidov:
Added the new class to the documentation. Preview: http://ev3dev-lang.readthedocs.io/projects/python-ev3dev/en/beacon-seeker/other.html#beacon-seeker

Issue #316, opened 2017-03-20
when shelling out to commands don't use shell=true


ensonic:
PR follows.

Right now we often do:
```
Popen('<cmdline>' % args, stdout=n, shell=True)
```
and return the popen object. This is problematic, since the client expects
```
cmd.terminate()
```
to work, but now this will only kill the shell.

ensonic:
For repro:
``` python
from ev3dev import auto as ev3dev
import time
sound = ev3dev.Sound

# don't do this, this will keep beeping until you play another beep
# x = sound.tone(100,1000); time.sleep(0.1); x.kill()
x = sound.tone(100,1000); time.sleep(0.1); x.terminate()

x = sound.play('/usr/share/sounds/alsa/Front_Center.wav'); time.sleep(0.5); x.terminate()

x = sound.speak('Hello'); time.sleep(0.5); x.terminate()

```

PR #317, opened 2017-03-20
Make sound commands cancelable.

ensonic:
Don't use a subshell to tokenize the commands. Using the subshell breaks
terminate() and kill() - they would only stop the subshell.

See #316 

The downside of this approach is that once we call terminate(), we'll leave defunct subprocs:
```
[beep] <defunct>
```
Those go away, when the parent terminates. Once thing I also noticed that it is not a good idea to kill() a beep, this will beep forever (how?) until you play another short beep.

### DIFF
--- a/ev3dev/core.py
+++ b/ev3dev/core.py
@@ -45,12 +45,13 @@ import array
 import mmap
 import ctypes
 import re
-import stat
 import select
+import shlex
+import stat
 import time
 from os.path import abspath
 from struct import pack, unpack
-from subprocess import Popen, check_output
+from subprocess import Popen, check_output, PIPE
 
 try:
     # This is a linux-specific module.
@@ -1794,13 +1795,13 @@ def list_sensors(name_pattern=Sensor.SYSTEM_DEVICE_NAME_CONVENTION, **kwargs):
 	    For example, 'sensor*'. Default value: '*'.
 	keyword arguments: used for matching the corresponding device
 	    attributes. For example, driver_name='lego-ev3-touch', or
-	    address=['in1', 'in3']. When argument value is a list, 
+	    address=['in1', 'in3']. When argument value is a list,
         then a match against any entry of the list is enough.
     """
     class_path = abspath(Device.DEVICE_ROOT_PATH + '/' + Sensor.SYSTEM_CLASS_NAME)
-    return (Sensor(name_pattern=name, name_exact=True) 
+    return (Sensor(name_pattern=name, name_exact=True)
             for name in list_device_names(class_path, name_pattern, **kwargs))
-        
+
 
 # ~autogen generic-class classes.i2cSensor>currentClass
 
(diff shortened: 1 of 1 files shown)


WasabiFan:
Given that there are many applications where beeping is done constantly, I'd be hesitant to allow processes to hang around. What we _really_ should be doing is beeping through proper APIs instead of hacking a textual command together :wink:

ensonic:
@WasabiFan the processes only hang around if you cancel the beep.

WasabiFan:
> @WasabiFan the processes only hang around if you cancel the beep.

If we're going to enable this functionality, people are going to use it.

---

Could we try the solution from here? http://stackoverflow.com/questions/12354586/python-what-are-the-nearest-linux-and-osx-equivalents-of-winsound-beep

ensonic:
@WasabiFan it is not a problem of the beep command, but how we handle it. I have an idea how to avoid the zombie issue. Let me try ...

WasabiFan:
I'm not specifically focused on this issue of processes left behind; I would much rather replace the hacks with proper calls than put lipstick on a pig.

ensonic:
Beep is the right tool for the job, we also use the same technique for playing wavs and the speech synth. There are no python solution for this available (pyaudio is terrible, trust me). I've fixed the zombie process issue - if you to cmd.terminate(), you need to also call cmd.wait() to get the return value and then the zombie is gone.

WasabiFan:
Ok, that's better; what do you think about returning a wrapper object which simplifies usage? That would make this easier to document in a useful way.

ensonic:
What do you mean by Wrapper object? All those methods return Popen objects. Those have a decent interface for being an async operation.
Also I would rather separate changing the interface and fixing unexpected behavior.

WasabiFan:
> What do you mean by Wrapper object?

Maybe this is just a result of my background in more OOP-focused languages:

```python
class SoundPlayResult:
    def __init__(self, popen_obj):
        self.popen_obj = popen_obj
    
    def cancel():
        self.popen_obj.terminate()
        self.popen_obj.wait()

```

Then we could add additional methods/properties for future additions and whatever other useful attributes the `Popen` objects have. If we later replaced the `Popen` calls with something more direct, we could change the internals of this class and the interface would remain unchanged.

> Also I would rather separate changing the interface and fixing unexpected behavior.

Sure, we should get a review from @ddemidov. I had you as a captive audience so I figured I'd try :wink:

ddemidov:
I see `Popen` with `shell=True` is also used in [`set_volune`](https://github.com/rhempel/ev3dev-lang-python/blob/develop/ev3dev/core.py#L3390). It is different in the sense that it waits for the command completion, but may be we should switch to shlex splitter here as well for consistency?

Otherwise, the changes look good. I am on a fence regarding Popen wrappers suggested by @WasabiFan. Wrapping the returned Popen objects so that `wait()` is called automatically is nice, but Popen objects probably have other functionality that we don't want to loose. Of course, we could solve this by making the wrapper a subclass of `Popen`, but it seems cleaner to just return the result as `Popen` instance.

ensonic:
Also changed `set_volume()`.

PR #319, opened 2017-04-09
Add tuple of all colors to ColorSensor

MaxNoe:
This adds a tuple mapping the color ints to human readable strings.

Could be used like this:

```python
from ev3dev.ev3 import ColorSensor

c = ColorSensor('1')
print(c.COLORS[c.color])
```

### DIFF
--- a/ev3dev/core.py
+++ b/ev3dev/core.py
@@ -1939,6 +1939,17 @@ class ColorSensor(Sensor):
     #: Brown color.
     COLOR_BROWN = 7
 
+    #: Tuple of all colors
+    COLORS = (
+        'nocolor',
+        'black',
+        'blue',
+        'green',
+        'yellow',
+        'red',
+        'white',
+        'brown',
+    )
 
     @property
     def reflected_light_intensity(self):


ddemidov:
This makes sense, but unfortunately the change is in the autogenerated region of the code (its between the [autogen](https://github.com/MaxNoe/ev3dev-lang-python/blob/7f77122312b4330d96c7740393ab152696feac92/ev3dev/core.py#L1853) [tags](https://github.com/MaxNoe/ev3dev-lang-python/blob/7f77122312b4330d96c7740393ab152696feac92/ev3dev/core.py#L2310), also see https://github.com/ev3dev/ev3dev-lang/tree/develop/autogen), so it will be overwritten on next regeneration.

I guess one way to deal with this would be to monkey-patch the `ColorSensor` class outside of any auto-generated region (`ColorSensor.COLORS=['nocolor',...]`), although that may be considered unintuitive by readers of the code.

We could also try to use the information available in the [spec.json](https://github.com/ev3dev/ev3dev-lang/blob/develop/spec.json#L1059-L1069) and generate the color dictionary automatically.

MaxNoe:
Superceded by #320

PR #320, opened 2017-04-14
Autocreate tuples of sensor properties

MaxNoe:
This is the followup for #319 

I adapted the autogen config to automatically create the tuples of the properties.


ddemidov:
Looks great, thank you!

WasabiFan:
Can we add documentation comments so this appears in the online docs?

ddemidov:
Something like
```py
#: List of valid {{ propName | downcase}}s
```
in front of each tuple could probably work.

PR #327, opened 2017-06-23
Added wait_until_not_moving()

dwalton76:
If you call `wait_while('running', timeout=1000)` there are times when the state reads both `running` and `stalled` so you end up waiting for your timeout to expire even though the motor has stalled.  The added function allows you to wait until the motor has stopped, either due to a stall or because it is no longer 'running'.

This helps my rubiks cube robot run a good bit faster as I don't end up waiting on so many timeouts in the wait_while calls.

### DIFF
--- a/ev3dev/core.py
+++ b/ev3dev/core.py
@@ -910,6 +910,19 @@ class Motor(Device):
                 return True
 
     def wait_until_not_moving(self, timeout=None):
+        """
+        Blocks until ``running`` is not in ``self.state`` or ``stalled`` is in
+        ``self.state``.  The condition is checked when there is an I/O event
+        related to the ``state`` attribute.  Exits early when ``timeout``
+        (in milliseconds) is reached.
+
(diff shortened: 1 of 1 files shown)


ev3dev/core.py
```diff
@@ -909,6 +909,8 @@ def wait(self, cond, timeout=None):
             if cond(self.state):
                 return True
 
+    def wait_until_not_moving(self, timeout=None):
```
WasabiFan: You should add a documentation comment here so that it shows up on the docs site.

dwalton76: I used it when solving a 5x5x5 cube which was about 9 minutes of off/on movement of three large motors and a medium motor. If it had exited early it would have caused the cube to jam up but it was smooth sailing so I think it is good to go. 

I'll add a docstring to it in the morning.

ddemidov:
Thanks!

PR #330, opened 2017-07-12
Scale values of type 'float' when returned from special sensor classes

### LINKED ISSUE #329: Ultrasonic sensor doesn't return centimeters for distance_centimeters value in US-DIST-CM mode
- **ev3dev version:** 4.4.68-20-ev3dev-ev3
- **ev3dev-lang-python version:** not sure how to get this info? Came installed with latest jessie ev3dev release

When using the ultrasonic sensor the distance doesn't seem to be in centimeters. The brickman tool on the ev3 does show the correct value when I read the sensor there. In python code I have to divide ultrasonic.distance_centimeters by 10 to get actual centimeters. A simple test case:

```
robot@ev3dev:/usr/lib/python3/dist-packages/ev3dev$ python3
Python 3.4.2 (default, Oct  8 2014, 14:47:30)
[GCC 4.9.1] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import ev3dev.ev3 as ev3
>>> ultrasonic = ev3.UltrasonicSensor('in4')
>>> ultrasonic.distance_centimeters
326
```

While the display shows 32.6 cm (which corresponds with real world distance from the sensor). I would expect to get either a floating point like 32.6 for distance_centimeters method or a changed method name like distance_milimeters or something. This seems incorrect, or am I doing something wrong? Workaround is easy enough and I'm having lots of fun programming my EV3 with python but thought I'd at least report this.

ddemidov:
Fixes #329

### DIFF
--- a/ev3dev/core.py
+++ b/ev3dev/core.py
@@ -1613,6 +1613,7 @@ class Sensor(Device):
         self._bin_data_format = None
         self._bin_data_size = None
         self._bin_data = None
+        self._mode_scale = {}
 
     __slots__ = [
 # ~autogen generic-class-slots classes.sensor>currentClass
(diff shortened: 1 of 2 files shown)


ddemidov:
Not yet tested on the EV3.

ddemidov:
So I did test this on the EV3, which resulted in 271f02a. But it seems I don't posses a single sensor that would both be listed in special sensors section *and* have a floating point value with non-zero decimals. All I could test is that `Sensor._scale(mode)` does work correctly (it returns 1 for my sensors). @marnovdm, would you mind testing this with your Ultrasonic sensor?

marnovdm:
Thanks for the update, I will test this when I get home later today and update this comment.

ddemidov:
Sorry, did not see the updated comment (github does not send notifications on updates, so its better to create new comments). Thanks for testing!

Issue #315, opened 2017-03-15
Motor synchronization


WasabiFan:
There's discussion that has been happening in https://github.com/ev3dev/ev3dev/issues/454 and https://github.com/ev3dev/ev3dev/issues/853 around synchronization of motors, mainly for wheeled robots. While kernel support for "proper" synchronization is not going to be arriving for a while, I think it's reasonable to implement this as best as possible in this library and adapt to the kernel model later.

I'm thinking that a `MotorPair` class which operates on given motor objects and/or port names would be reasonable as a Python implementation. It would be nice to provide functions for steering, potentially similar to the EV3-G functions.

Thoughts?

ddemidov:
Something like [Tank](https://github.com/rhempel/ev3dev-lang-python/blob/develop/ev3dev/helper.py#L170) class in helper.py?

ddemidov:
Hmm, that class is not what I thought it was. I thought @dwalton76 had something like this implemented.

WasabiFan:
In particular, one of the features of this class is that it would be as efficient as it can be for a library written in Python. That primarily means we'd set all the motor properties together and then send the commands sequentially to start them at the same time, which makes a surprisingly large difference in my brief testing.

dwalton76:
I can't think of any builds I've done where I needed a pair of motors to start at the exact same time...I don't think it was me :)

dwalton76:
@ddemidov maybe you are thinking of TankWebHandler in https://github.com/rhempel/ev3dev-lang-python/blob/develop/ev3dev/webserver.py , specifically around line 500 where it is starting two motors but with different speeds in order to turn the tank?  This was the tank that uses a web interface with a little "joystick" dot that you move around inside a larger circle to control the speed/direction of the tank.

ddemidov:
So what would be the interface?
```py
class MotorPair(object):
  def __init__(self, m1, m2):
    ...
  def drive(self, power, direction):
    ...
```
What else?

WasabiFan:
The ones I see are:

- The ability to control them in "tank" mode (individual powers for the two motors) as opposed to the current "steering" mode
- Options for both power and speed control
- A method to stop both motors
- An interface to set arbitrary properties on both motors

Additionally, it would be interesting to see if we could implement it such that the interface was interchangeable with a normal motor, allowing you to pair `MotorPair`s (e.g., if you have four motors for each corner of a 'bot). That would be a stretch goal though.

[...]

PR #336, opened 2017-09-02
Added class MotorSet

dwalton76:
The MotorSet class can be used to start/stop more than one motor at once as close together as we possibly can.  You can set different params for different motors in the set.  Here is a small script I used to try a few things.

```
#!/usr/bin/env python3
from ev3dev.auto import OUTPUT_A, OUTPUT_B, OUTPUT_C, OUTPUT_D, Motor
from ev3dev.helper import LargeMotorPair
import argparse
import logging
import time
import sys

# logging
logging.basicConfig(level=logging.INFO,

log.info("start")
tank = LargeMotorPair(OUTPUT_A, OUTPUT_C)
tank.reset()
#tank.run_to_abs_pos(speed_sp=240, position_sp=360, stop_action='brake')
#tank.wait_until_not_moving()

motor_A = tank.motors[OUTPUT_A]
motor_C = tank.motors[OUTPUT_C]
tank.set_args(motors=[motor_A], speed_sp='120')
tank.set_args(motors=[motor_C], speed_sp='240')

tank.run_forever(stop_action='brake')
time.sleep(2)
tank.stop()
log.info("end")
```


dwalton76:
For issue #315

ev3dev/helper.py
```diff
@@ -146,6 +147,171 @@ class MediumMotor(ev3dev.auto.MediumMotor, MotorMixin):
...
+class MotorSet(object):
+    """
+    motor_specs is a dictionary such as
+    {
+        OUTPUT_A : LargeMotor,
+        OUTPUT_B : MediumMotor,
+    }
+    """
+
+    def __init__(self, motor_specs, desc=None):
```
WasabiFan, line 159: Each motor object can tell you what port it's on... why does this need to take a dictionary?

ev3dev/helper.py
```diff
@@ -146,6 +147,171 @@ class MediumMotor(ev3dev.auto.MediumMotor, MotorMixin):
...
+    def _run_command(self, **kwargs):
+        motors = kwargs.get('motors', self.motors.values())
+
+        for motor in motors:
+            for key in kwargs:
+                if key not in ('motors', 'commands'):
+                    log.info("%s: %s set %s to %s" % (self, motor, key, kwargs[key]))
+                    setattr(motor, key, kwargs[key])
+
+        for motor in motors:
```
WasabiFan: I like this process (setting the properties then sending commands).

ev3dev/helper.py
```diff
@@ -146,6 +147,171 @@ class MediumMotor(ev3dev.auto.MediumMotor, MotorMixin):
...
+            motor.wait_until(s, timeout)
+
+    def wait_while(self, s, timeout=None, motors=None):
+        motors = motors if motors is not None else self.motors.values()
+
+        for motor in motors:
+            motor.wait_while(s, timeout)
+
+
+class LargeMotorPair(MotorSet):
```
WasabiFan: I think we should figure out exactly what our goal is. In my mind, the most common scenario will be a pair of large motors which you want to "steer": even if internally we're using a generic `MotorSet`, I'd like to see if we can make the "pair" classes more specialized so they legitimately feel easy to use and work "out of the box".

ev3dev/helper.py
dwalton76: The motor objects haven't been created yet though, we need to know what type of objects to make and which ports each of those objects is associated with.

ev3dev/helper.py
```diff
@@ -146,6 +147,171 @@ class MediumMotor(ev3dev.auto.MediumMotor, MotorMixin):
...
+
+        for motor_port in motor_specs.keys():
+            if motor_port not in OUTPUTS:
+                log.error("%s in an invalid motor, choices are %s" % (motor_port, ', '.join(OUTPUTS)))
+                sys.exit(1)
+
+        self.motors = OrderedDict()
+        for motor_port in sorted(motor_specs.keys()):
+            motor_class = motor_specs[motor_port]
+            self.motors[motor_port] = motor_class(motor_port)
```
dwalton76, line 169: The motor objects are created here

ev3dev/helper.py
WasabiFan: Sorry, I read through the rest of the file then forgot to come back and delete this. I've got it now.

ev3dev/helper.py
dwalton76: yeah....to me part of it depends on what the user input is to control the steering.  If it is the IR remote we have that implemented in the RemoteControlledTank class in helper.py...we could pull some of that code into a Mixin class that could be inherited here in LargeMotorPair.  I also did a web interface for a tank, that could be pulled in too but loading the www modules makes the program take a while to load on an ev3. What other user input methods might we see for a MotorPair?

WasabiFan: The scenarios I had in mind were:
- "Run the left motor at 50% speed forward and the right motor at 10% speed backward"
- "Run the two motors in a proportion of 3:1" (ratio, differential, etc.)

The second is a special case of the first. Note that the units are nonspecific and just an illustration. The idea is that this class is what's used whenever you have a driving robot... if you want to go straight, turn, spin, etc., you use this.

ev3dev/helper.py
```diff
@@ -146,6 +147,171 @@ class MediumMotor(ev3dev.auto.MediumMotor, MotorMixin):
...
+                    log.info("%s: %s set %s to %s" % (self, motor, key, kwargs[key]))
+                    setattr(motor, key, kwargs[key])
+
+        for motor in motors:
+            motor.command = kwargs['command']
+            log.info("%s: %s command %s" % (self, motor, kwargs['command']))
+
+    def run_forever(self, **kwargs):
+        kwargs['command'] = ev3dev.auto.LargeMotor.COMMAND_RUN_FOREVER
+        self._run_command(**kwargs)
```
ddemidov: One thing I am not sure about is the logic of these `run_*` functions. Here you are pushing command into kwargs, and then filter it out in [`_run_command` function](https://github.com/dwalton76/ev3dev-lang-python/blob/f928075be9f7079d093cc44c04fb0caeb3fff2f3/ev3dev/helper.py#L198-L202). This cost a string comparison per key in kwargs, which could be negligible with respect to file acesses, but still is not free.

Why not create `_set_args(self, **kwargs)` function, and replace the two lines above with
```python
self._set_args(**kwargs)
self._run_command(ev3dev.auto.LargeMotor.COMMAND_RUN_FOREVER)
```
?

dwalton76: There is a set_args() (see line 187) and the user can do it this way if they want but they don't have to.

The price of the string compare feels worth it to me, it let me clean up this code a good bit by moving almost all of the logic to _run_command().

ddemidov: Well, I don't have strong feelings about this, so feel free to merge.

ev3dev/helper.py
```diff
@@ -146,6 +147,217 @@ class MediumMotor(ev3dev.auto.MediumMotor, MotorMixin):
...
+        (self.left_motor, self.right_motor) = self.motors.values()
+        self.max_speed = self.left_motor.max_speed
+
+    def set_speed_ratio(self, left_motor_ratio, right_motor_ratio, set_right=True):
+        """
+        Set the speeds of the left_motor vs right_motor by ratio
+        """
+
+        # Set the right_motor speed relative to the left_motor_speed
+        if set_right:
```
WasabiFan: Huh, this isn't the behavior I had in mind :laughing: I was thinking you would specify a speed and ratio, and it would run the fastest motor at that speed and the other one at a speed defined by the supplied ratio. My thinking was that we'd mimic the "steering" behavior of EV3-G.

ev3dev/helper.py
```diff
@@ -146,6 +147,217 @@ class MediumMotor(ev3dev.auto.MediumMotor, MotorMixin):
...
+            self.right_motor.speed_sp = int((self.left_motor.speed_sp * left_motor_ratio) / right_motor_ratio)
+        else:
+            self.left_motor.speed_sp = int((self.right_motor.speed_sp * right_motor_ratio) / left_motor_ratio)
+
+    def set_speed_percentage(self, left_motor_percentage, right_motor_percentage):
+        """
+        Set the speeds of the left_motor vs right_motor by percentage of their maximum speed
```
WasabiFan: We should specify that the units are -100 to 100. The term "percentage" is often  used to refer to [-1,1] or [0,1] by some programmers.

ev3dev/helper.py
ddemidov: Something like this?
https://github.com/ddemidov/ev3dev-lang-python/blob/master-deprecated/ev3dev/__init__.py#L85-L121

dwalton76: I haven't touched EV3-G since the day I found ev3dev :) Does its steering do what ddemidov points to or something else?

ev3dev/helper.py
dwalton76: Ack...will also add some asserts to verify the inputs are within the range

ev3dev/helper.py
WasabiFan: Looking at the "synchronized steer" code [here](https://github.com/mindboards/ev3sources/blob/master/lms2012/d_pwm/Linuxmod_AM1808/d_pwm.c#L1963), I think I can say that they _don't_ match but @ddemidov's implementation would be preferred. LEGO seems to designate one motor as a "master", drive that one at the requested speed, and then run the other at a rate defined by the steering. I am no expert though so I may be mis-reading it.

ev3dev/helper.py
```diff
@@ -311,28 +311,56 @@ def __init__(self, motor_specs, desc=None):
...
         """
-        Set the speeds of the left_motor vs right_motor by ratio
+        Set the speed_sp for each motor in a pair to achieve the specified steering
+
+        direction [-100, 100]:
+            * -100 means turn left as fast as possible,
+            *  0   means drive in a straight line, and
+            *  100 means turn right as fast as possible.
+
+        power: the power that should be applied to the outmost motor (the one
```
WasabiFan: _power_? I think this should probably be "speed"

ev3dev/helper.py
```diff
@@ -311,28 +311,56 @@ def __init__(self, motor_specs, desc=None):
...
         else:
-            self.left_motor.speed_sp = int((self.right_motor.speed_sp * right_motor_ratio) / left_motor_ratio)
+            left_power *= speed
+
+        left_power = int(left_power)
+        right_power = int(right_power)
+        log.debug("%s: direction %d, power %d, speed %d, left_power %d, right_power %d" %
+            (self, direction, power, speed, left_power, right_power))
+
+        self.left_motor.speed_sp = int(left_power)
```
WasabiFan: We don't actually run a command here; is the expectation that you call `set_speed_steering` and then set the command in a loop? Note that updating the setpoint won't have an effect until a command is sent AFAIK.

dwalton76: yep it will be up to the user to call run_forever, run_timed, etc after this....all this function does is set them up with the desired speed_sp setting.  Same goes for set_speed_percentage().

ev3dev/helper.py
dwalton76: Caught that in the initial cut-n-paste....will change all of the power references to speed

ev3dev/helper.py
WasabiFan: Hmm... In that case, we need to make sure we document the behavior!

dwalton76:
Is 1.0.0 out...shall we release the merging floodgates?

WasabiFan:
Can you add a line to the documentation for `set_speed_steering` and the other related function explaining you need to call a `run_*` function afterward?

Also, is there anything we want to do here in response to #348 (e.g. renaming or changing the behavior to match EV3-G)?

dwalton76:
yep will add a line about needing to call `run_XYZ`

For #348 lets tackle any of that via a separate pull request. I need to take a look at what blocks EV3-G has, I'm not sure if the MotorSet class would cover one of them or not.

Issue #147, opened 2016-03-14
bad assumption on port path in motor test


dlech:
See https://github.com/rhempel/ev3dev-lang-python/commit/7cc7e116c0618c0dc8c0b3a2e632b2895c7289d3#commitcomment-16671433

dwalton76:
@ddemidov @WasabiFan do either of you remember how to run the tests? I was going to work on this but I've never actually run the tests and couldn't find docs on it.

WasabiFan:
Woah, we have tests? :laughing: But seriously, I think you just have to run the `api_tests.py` file. I'm not 100% sure they've been confirmed working since we upgraded to Python3 though.

ddemidov:
We do have tests and they are run on every commit and every pull request here:
https://travis-ci.org/rhempel/ev3dev-lang-python/builds

The steps needed to run the tests are 'documented' here:
https://github.com/rhempel/ev3dev-lang-python/blob/develop/.travis.yml#L8

WasabiFan:
I was joking! 😆 Well, mostly. I totally forgot we ran them in CO builds though.

dwalton76:
ah I did not have fake-sys that was part of my problem.  @ddemidov we create a directory called lego-sensor/sensor0 and cp the contents of /sys/class/lego-sensor/sensor0 to that directory
```
$ node=lego-sensor/sensor0
$ mkdir -p ./${node}
$ cp -P --copy-contents -r /sys/class/${node}/* ./${node}/
```
So this has to run on a device running ev3dev?  I'm not following how we are running the tests in github on each commit unless we have an ev3dev vm that spins up to run the tests

When I run it I get
```
dwalton@laptop ~/l/e/tests> ./api_tests.py 
	0	in1	/home/dwalton/lego/ev3dev-lang-python/tests/fake-sys/arena/lego-sensor/sensor0
	0	outA	/home/dwalton/lego/ev3dev-lang-python/tests/fake-sys/arena/tacho-motor/motor0
.	0	in1	/home/dwalton/lego/ev3dev-lang-python/tests/fake-sys/arena/lego-sensor/sensor0
/usr/lib/python3.5/unittest/case.py:600: ResourceWarning: unclosed file <_io.FileIO name='/home/dwalton/lego/ev3dev-lang-python/tests/fake-sys/arena/lego-sensor/sensor0/address' mode='rb+' closefd=True>
  testMethod()
[...]

PR #337, opened 2017-09-02
Remove hard coded outA on port4 assumption

dwalton76:
For issues #147


WasabiFan:
Why do we need to discover the port here? Isn't this only run from dry test data files? I must be tired...

dwalton76:
He code assumed motor A was on port4 which isn't necessarily the case.

dwalton76:
I'll wait until 1.0 is out and then merge this

PR #342, opened 2017-09-05
Throw friendly errors

### LINKED ISSUE #331: OSError when setting too large speed_sp
- **ev3dev version:** 4.4.68-20-ev3dev-ev3

One of my students ran into an issue where they set `motor.run_forever` for a large motor via a formula. Once the result reached 1050 they got a `OSError: InvalidArgument`. 1050 is the max_speed. I think the library should cap the input at 1050 (or the max_speed of the corresponding tacho-motor) or throw a `ValueError` instead.
The error message, in any way, is very cryptic considering how easy it is to trigger.

speed_R was set to something larger than the max_speed.

```
  File "./braitenberg4b.py", line 26, in <module>
    motor_R.run_forever(speed_sp = speed_R)
  File "/usr/lib/python3/dist-packages/ev3dev/core.py", line 793, in run_forever
    setattr(self, key, kwargs[key])
  File "/usr/lib/python3/dist-packages/ev3dev/core.py", line 583, in speed_sp
    self._speed_sp = self.set_attr_int(self._speed_sp, 'speed_sp', value)
  File "/usr/lib/python3/dist-packages/ev3dev/core.py", line 217, in set_attr_int
    return self._set_attribute(attribute, name, str(int(value)))
  File "/usr/lib/python3/dist-packages/ev3dev/core.py", line 206, in _set_attribute
    attribute.write(value.encode())
OSError: [Errno 22] Invalid argument
```

WasabiFan:
Fixes #331 

It would be nice if we were able to adapt the message when you attempt to run for distance with a negative sign, but commands are read-only and I didn't see an easy way to pipe that through. Any suggestions for other properties?

```
robot@ev3dev:~$ python3
Python 3.5.3 (default, Jan 19 2017, 14:11:04)
[GCC 6.3.0 20170118] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import ev3dev.ev3 as ev3
>>> m = ev3.Motor()
>>> m.connected
True
>>> m.speed_sp = -100000
  File "/usr/lib/python3/dist-packages/ev3dev/core.py", line 210, in _set_attribute
    attribute.write(value.encode())
OSError: [Errno 22] Invalid argument

The above exception was the direct cause of the following exception:

  File "/usr/lib/python3/dist-packages/ev3dev/core.py", line 604, in speed_sp
    self._speed_sp = self.set_attr_int(self._speed_sp, 'speed_sp', value)
  File "/usr/lib/python3/dist-packages/ev3dev/core.py", line 238, in set_attr_int
    return self._set_attribute(attribute, name, str(int(value)))
  File "/usr/lib/python3/dist-packages/ev3dev/core.py", line 213, in _set_attribute
    self._raise_friendly_access_error(ex, name)
  File "/usr/lib/python3/dist-packages/ev3dev/core.py", line 229, in _raise_friendly_access_error
    raise ValueError("The given speed value was out of range. Max speed: +/-" + str(max_speed)) from driver_error
ValueError: The given speed value was out of range. Max speed: +/-1050
>>> m.speed_sp = 1000000
  File "/usr/lib/python3/dist-packages/ev3dev/core.py", line 210, in _set_attribute
    attribute.write(value.encode())
OSError: [Errno 22] Invalid argument

The above exception was the direct cause of the following exception:

  File "/usr/lib/python3/dist-packages/ev3dev/core.py", line 604, in speed_sp
    self._speed_sp = self.set_attr_int(self._speed_sp, 'speed_sp', value)
  File "/usr/lib/python3/dist-packages/ev3dev/core.py", line 238, in set_attr_int
    return self._set_attribute(attribute, name, str(int(value)))
  File "/usr/lib/python3/dist-packages/ev3dev/core.py", line 213, in _set_attribute
    self._raise_friendly_access_error(ex, name)
  File "/usr/lib/python3/dist-packages/ev3dev/core.py", line 229, in _raise_friendly_access_error
    raise ValueError("The given speed value was out of range. Max speed: +/-" + str(max_speed)) from driver_error
ValueError: The given speed value was out of range. Max speed: +/-1050
```

### DIFF
--- a/ev3dev/core.py
+++ b/ev3dev/core.py
@@ -49,6 +49,7 @@ import select
 import shlex
 import stat
 import time
+import errno
 from os.path import abspath
 from struct import pack, unpack
 from subprocess import Popen, check_output, PIPE
@@ -209,12 +210,12 @@ class Device(object):
                 attribute.write(value.encode())
                 attribute.flush()
             except Exception as ex:
-                _raise_friendly_access_error(ex, name)
+                self._raise_friendly_access_error(ex, name)
             return attribute
         else:
             raise Exception('Device is not connected')
 
-    def _raise_friendly_access_error(driver_error, attribute):
+    def _raise_friendly_access_error(self, driver_error, attribute):
         if not isinstance(driver_error, OSError):
             raise driver_error
 
@@ -223,10 +224,10 @@ class Device(object):
                 try:
                     max_speed = self.max_speed
                 except (AttributeError, Exception):
-                    raise ValueError("The given speed value was out of range")
+                    raise ValueError("The given speed value was out of range") from driver_error
                 else:
-                    raise ValueError("The given speed value was out of range. Max speed: " + str(max_speed))
-            raise ValueError("One or more arguments were out of range or invalid")
+                    raise ValueError("The given speed value was out of range. Max speed: +/-" + str(max_speed)) from driver_error
+            raise ValueError("One or more arguments were out of range or invalid") from driver_error
         raise driver_error
 
     def get_attr_int(self, attribute, name):
--- a/ev3dev/core.py
+++ b/ev3dev/core.py
@@ -223,10 +223,10 @@ class Device(object):
                 try:
                     max_speed = self.max_speed
                 except (AttributeError, Exception):
-                    raise Exception("The given speed value was out of range")
+                    raise ValueError("The given speed value was out of range")
                 else:
-                    raise Exception("The given speed value was out of range. Max speed: " + str(max_speed))
-            raise Exception("One or more arguments were out of range or invalid")
+                    raise ValueError("The given speed value was out of range. Max speed: " + str(max_speed))
+            raise ValueError("One or more arguments were out of range or invalid")
         raise driver_error
 
     def get_attr_int(self, attribute, name):
--- a/ev3dev/core.py
+++ b/ev3dev/core.py
@@ -204,11 +204,30 @@ class Device(object):
                 attribute = self._attribute_file_open( name )
             else:
                 attribute.seek(0)
-            attribute.write(value.encode())
-            attribute.flush()
+
+            try:
+                attribute.write(value.encode())
+                attribute.flush()
+            except Exception as ex:
+                _raise_friendly_access_error(ex, name)
             return attribute
         else:
             raise Exception('Device is not connected')
+
+    def _raise_friendly_access_error(driver_error, attribute):
+        if not isinstance(driver_error, OSError):
+            raise driver_error
+
+        if driver_error.errno == errno.EINVAL:
+            if attribute == "speed_sp":
+                try:
+                    max_speed = self.max_speed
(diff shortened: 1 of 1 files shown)


WasabiFan:
Another potential candidate for friendlyifying:

```
  File "core.py", line 487, in position
  File "core.py", line 210, in get_attr_int
  File "core.py", line 192, in _get_attribute
OSError: [Errno 19] ENODEV
```

(instantiate motor, disconnect motor, access attribute... BAM! error.)

ev3dev/core.py
```diff
@@ -204,12 +205,31 @@ def _set_attribute(self, attribute, name, value):
...
+
+        if driver_error.errno == errno.EINVAL:
+            if attribute == "speed_sp":
+                try:
+                    max_speed = self.max_speed
+                except (AttributeError, Exception):
+                    raise ValueError("The given speed value was out of range") from driver_error
+                else:
+                    raise ValueError("The given speed value was out of range. Max speed: +/-" + str(max_speed)) from driver_error
+            raise ValueError("One or more arguments were out of range or invalid") from driver_error
```
ddemidov: I am still not convinced a value error is more friendly than OSError to be fair. OSError is the thing that happens in the background, why do we want to 'protect' the user from this?

May be it would be better to at least re-raise the original (OSError) exception instead of generic 'One or more arguments were out of range or invalid' for unsupported attributes?

ddemidov: I mean, what would make the debugging easier?

WasabiFan: > I am still not convinced a value error is more friendly than OSError to be fair. OSError is the thing that happens in the background, why do we want to 'protect' the user from this?

That the nice Python functions happen to be calling an operating system routine is an implementation detail. As far as the user is concerned, they called the function and provided an argument that was out-of-range; a `ValueError` is the Python concept that describes such an issue, and the message explains what went wrong. If the user is interested in the `OSError`, they can see it as the `__cause__` of the `ValueError` (hence `raise ... from`).

For most casual programmers, the message `OSError: [Errno 22] Invalid argument` will be confusing -- it isn't natural to expect an `OSError` for this sort of scenario, and the message certainly isn't clear. Unless I'm directly consuming OS functions, I would generally expect an `OSError` to indicate an issue with the implementation of a library, not my own mistake.

> May be it would be better to at least re-raise the original (OSError) exception instead of generic 'One or more arguments were out of range or invalid' for unsupported attributes?

See above; the inner exception is available, just within the user-friendly error that we throw.

WasabiFan: > I mean, what would make the debugging easier?

The error tells you exactly what you did wrong; no question is left as to what was going on when the error occurred or what caused it.

dwalton76: `what would make the debugging easier` for me if the exception text always included
- the name of the attribute we are trying to set
- the value we are trying to set it to
- the min/max or acceptable inputs for that attribute

that would be very useful

PR #345, opened 2017-09-05
Remove unused files

### LINKED ISSUE #344: Unused files?
I just noticed that there are two files which don't seem to be referenced anywhere. [`framebuffer.py`](https://github.com/rhempel/ev3dev-lang-python/blob/develop/ev3dev/framebuffer.py) and [`virtualterminal.py`](https://github.com/rhempel/ev3dev-lang-python/blob/develop/ev3dev/virtualterminal.py), the two files in question, were both created in 2016 but it seems that we've since reimplemented the functionality (for the framebuffer, at least) in `core.py`. Are these files still needed/used?

WasabiFan:
Fixes #344


dwalton76:
Looks good....I'm assuming we are holding off on all PRs until 1.0 is out yes?

WasabiFan:
It seems that's what we are doing. I would like to release v1.0.0 and then merge the outstanding PRs, do some testing, and release another version relatively quickly given that we have so many.

PR #346, opened 2017-09-05
Support micropython

WasabiFan:
This is my first crack at supporting micropython. Note that I haven't made any packaging changes, just rearranged things to make code run. I think it's a reasonable first step, but isn't useful yet without packaging.

The question is: do we want to do it this way? I chose to separate the two "optional" classes into their own file to give some separation, but that isn't necessary.

My eventual preference would be to split each module into its own separately-importable file: motors, sensors, LEDs, screen, sound, etc. That would both decrease load time (the interpreter doesn't need to parse a bunch of stuff that isn't used) and make it easier to navigate the file structure during development. However, I think these changes will be useful regardless of whether we decide to restructure or not.


ev3dev/core.py
```diff
@@ -145,8 +143,8 @@ def __init__(self, class_name, name_pattern='*', name_exact=False, **kwargs):
 
         def get_index(file):
             match = Device._DEVICE_INDEX.match(file)
-            if match:
-                return int(match.group('idx'))
```
WasabiFan: MicroPython's implementation of the `re` module is barebones and doesn't have support for named groups. I don't see any reason why we need to use a named group anyway.

ev3dev/core.py
```diff
@@ -146,7 +143,7 @@ def __init__(self, class_name, name_pattern='*', name_exact=False, **kwargs):
         def get_index(file):
             match = Device._DEVICE_INDEX.match(file)
             if match:
-                return int(match.group('idx'))
```
WasabiFan: MicroPython's implementation of the `re` module is barebones and doesn't have support for named groups. I don't see any reason why we need to use a named group anyway.

ddemidov:
Do you have timing tests from the actual EV3 for this?

WasabiFan:
Here are some timing tests, although I'm moderately suspicious of them:

```
robot@ev3dev:~$ python3 -m timeit 'import ev3dev.ev3 as ev3'
10 loops, best of 3: 365 usec per loop
robot@ev3dev:~$ micropython -m timeit 'import ev3dev.ev3 as ev3'
10 loops, best of 3: 242 usec per loop
robot@ev3dev:~$ python3 -m timeit 'import ev3dev.ev3 as ev3; m = ev3.Motor(); m.connected'
10 loops, best of 3: 7.35 msec per loop
robot@ev3dev:~$ micropython -m timeit 'import ev3dev.ev3 as ev3; m = ev3.Motor(); m.connected'
10 loops, best of 3: 15.8 msec per loop
robot@ev3dev:~$ micropython -m timeit 'import ev3dev.ev3 as ev3; m = ev3.Motor(); m.connected'
10 loops, best of 3: 16.2 msec per loop
robot@ev3dev:~$ python3 -m timeit 'import ev3dev.ev3 as ev3; m = ev3.Motor(); m.connected'
10 loops, best of 3: 7.38 msec per loop
```

ddemidov:
Yes, I think python caches imports. Also, you are probably missing the time it takes to load the interpreter as well. May be something like
```
$ time python3 hello.py
$ time micropython hello.py
```
would be more representative. Here hello.py could be
```py
import ev3dev.ev3 as ev3
m = ev3.Motor()
print(m.connected)
```

ev3dev/optional.py
```diff
@@ -0,0 +1,692 @@
+"""
```
dwalton76: If we are moving them now and long term we want them to be in sound.py and screen.py files (I like that idea) why not go ahead and put them there now?

dwalton76:
```
My eventual preference would be to split each module into its own separately-importable file: motors, sensors, LEDs, screen, sound, etc. That would both decrease load time (the interpreter doesn't need to parse a bunch of stuff that isn't used) and make it easier to navigate the file structure during development. 
```

+1

ev3dev/optional.py
WasabiFan: To be honest, I put them in the same file because the existing `framebuffer.py` was annoying me :laughing: Although, I am not entirely sure how I want to organize the platform check; I would prefer that the logic to compare implementation names is centralized in one location. Thoughts?

WasabiFan:
Good procedure, @ddemidov!

```
robot@ev3dev:~$ cat hello.py
import ev3dev.ev3 as ev3
m = ev3.Motor()
print(m.connected)
robot@ev3dev:~$ time python3 hello.py
True

real    0m10.202s
user    0m6.420s
sys     0m0.530s
robot@ev3dev:~$ time python3 hello.py
True

real    0m10.112s
user    0m6.480s
sys     0m0.420s
robot@ev3dev:~$ time python3 hello.py
True

real    0m10.375s
user    0m6.610s
sys     0m0.350s
robot@ev3dev:~$ time micropython hello.py
True

real    0m3.740s
user    0m2.230s
sys     0m0.360s
robot@ev3dev:~$ time micropython hello.py
True

real    0m3.981s
user    0m2.270s
sys     0m0.400s
robot@ev3dev:~$ time micropython hello.py
True

real    0m3.782s
user    0m2.290s
sys     0m0.320s
```

Mean 10.2 seconds for `python3`, 3.8 seconds for `micropython`. I figure that could decrease significantly further if we rearranged files so that we only loaded what we needed (at the moment it still parses the structure of the unsupported classes).

ev3dev/optional.py
```diff
@@ -0,0 +1,678 @@
...
+A set of optional modules for interfacing with peripherals (screen and sound)
+"""
+
+import sys
+optional_enabled = sys.implementation.name is not 'micropython'
+
```
ensonic, line 7: what are the features needed here missing on micropython? Can those be added as a comment? I guess it is the subprocess module, but most of the sound stuff should work with os.system().

ddemidov:
Some more details here: ev3dev/ev3dev#937
I think what we miss are mmap and ctypes (both needed for framebuffer/Screen), and subprocess (for Sound).

dwalton76:
For subprocess looks like we can use os.pipe() as a work-around
https://forum.micropython.org/viewtopic.php?t=851

ddemidov:
Part of the documented Sound interface is that users may call `.wait()` on operations like `Sound.play()`. Can we keep same functionality with `os.popen()` or `os.system()`?

dlech:
> I think what we miss are mmap and ctypes (both needed for framebuffer/Screen)

More importantly, PIL (pillow) will be missing for micro python.

Maybe this will help?

https://micropython.org/resources/docs/en/latest/wipy/library/framebuf.html

dlech:
Oh, and you can `write` to `/dev/fb0` instead of using `mmap` if that is not available.

dwalton76:
Not sure on a `.wait()` equivalent...I've never actually used os.pipe() just found that post when googling about micropython and subprocess.

ensonic:
Well os.system() would be with an implicit .wait() otherwise we need to use a thread.

WasabiFan:
Yes, the here are certainly ways to reimplement this functionality on micropython. I removed the impacted classes here so that we had a starting point to look into packaging and releasing it, and then we could modify these implementations as desired to use micropython's embedded interfaces or alternative APIs in the future.

WasabiFan:
I'm going to close this and break out the only independent portion of these changes for a different PR. The PR in its current form is really just a partial split of some classes into their own modules, which is covered in full by #375; micropython support didn't require any other major changes. That will be all that's necessary to "support" micropython at a minimum, excluding packaging.

WasabiFan:
#376

Issue #353, opened 2017-09-19
EV3-G API: motors


dwalton76:
![motor-01](https://user-images.githubusercontent.com/468309/30606466-ce662b7a-9d3f-11e7-9367-9875b5b1adb8.png)
![motor-02](https://user-images.githubusercontent.com/468309/30606469-ce6ac464-9d3f-11e7-89e3-bb460593444d.png)
![motor-03](https://user-images.githubusercontent.com/468309/30606467-ce674b86-9d3f-11e7-92a5-e456766595cd.png)
![move-steering-01](https://user-images.githubusercontent.com/468309/30606468-ce67e6a4-9d3f-11e7-8b22-a56ab9b6fb87.png)
![move-tank-01](https://user-images.githubusercontent.com/468309/30606470-ce6be934-9d3f-11e7-918d-9f4f25771987.png)

ddemidov:
We could merge various `on_for_*` methods into 
```py
m.on_for(rotations=3, brake=True)
m.on_for(degrees=45, power=75)
m.on_for(seconds=1)
```
Here `on_for` could be defined as
```py
def m.on_for(rotations=None, degrees=None, seconds=None, **kwargs):
  # set attributes from kwargs
  ...
  if rotations is not None:
    ...
  elif degrees is not None:
    ...
```

dwalton76:
Does EV3-G not provide a way to specify `STOP_ACTION_HOLD` when using their Motor, MoveTank, MoveSteering blocks?  It doesn't look like they do which seems a little odd.

I am trying to get EV3-G running via wine so that I can check for this kind of stuff but no luck yet

ddemidov:
I think `hold` is ev3dev specific.

dlech:
`brake` is ev3dev-specific. `hold` in ev3dev == Brake in EV3-G.

http://docs.ev3dev.org/projects/lego-linux-drivers/en/ev3dev-stretch/motors.html#sysfs

laurensvalk:
I came across this thread after being tagged in something so hopefully I'm not taking this too far off topic...

Quoting from the table for the yellow blocks:
> current_power = m.measure_current_power()

I understand the idea here is to obtain functionality close to EV3-G, and I could live with power and speed being used interchangeably for the green blocks, but this one is quite exceptional.

Measuring "current power" in EV3-G gives you the speed (no relation to current or power at all).

[...]

PR #360, opened 2017-09-20
EV3-G API: motors

dwalton76:


dwalton76:
For issue #353 
I'm still testing everything (so far so good) but figured I would go ahead and post it so everyone can start giving feedback before I spend too much time testing.

ev3dev/core.py
```diff
@@ -955,6 +958,94 @@ def wait_while(self, s, timeout=None):
...
+        self.speed_sp = power
+
+    def set_position_rotations(self, power, rotations):
+        if power > 0:
+            self.position_sp = self.position + int(rotations * self.count_per_rot)
+        else:
+            self.position_sp = self.position - int(rotations * self.count_per_rot)
+
+    def set_position_degrees(self, power, degrees):
+        rotations = float(degrees / self.count_per_rot)
```
WasabiFan: This codepath will read from `count_per_rot` twice. Can we restructure this to avoid that?

ev3dev/core.py
```diff
@@ -955,6 +958,94 @@ def wait_while(self, s, timeout=None):
         """
         return self.wait(lambda state: s not in state, timeout)
 
+    def set_power(self, power):
+        self.speed_sp = power
+
+    def set_position_rotations(self, power, rotations):
```
WasabiFan: What are these methods (setting the `position_sp`, `time_sp`, etc.) for? They don't seem to match up with the EV3-G blocks in the spec image.

ev3dev/core.py
```diff
@@ -955,6 +958,94 @@ def wait_while(self, s, timeout=None):
...
+            self.stop_action = self.STOP_ACTION_COAST
+
+    def set_time(self, ms):
+        self.time_sp = ms
+
+    def on_for_rotations(self, power, rotations, brake):
+        """
+        Rotate the motor at 'power' for 'rotations'
+        """
+        self.set_power(power)
```
WasabiFan: These powers will be in tacho counts per second. Do we not want to take input on a [-100,100] scale as LEGO does?

ev3dev/core.py
```diff
@@ -955,6 +958,94 @@ def wait_while(self, s, timeout=None):
...
+    def get_rotations(self):
+        return float(self.position / self.count_per_rot)
+
+    def get_degrees(self):
+        return self.get_rotations() * 360
+
+    def get_power(self):
+        return self.speed
+
+    def rotations_greater_than(self, threshold):
```
WasabiFan: I think we should leave the users to do these comparisons themselves. I feel like this obfuscates the logic without any real gain.

ev3dev/core.py
dwalton76: Originally in `on_for_degrees`, etc I just had a `run_to_abs_pos()` call where I passed in all of needed kwargs.  This worked fine for a single motor but for MoveTank and MoveSteering if I called
```
    self.left_motor.on_for_degrees(...)
    self.right_motor.on_for_degrees(...)
```
it added delay between when the two motors start.  We want to follow the model of "set as many parameters as you can for both motors, then start the motors" so that they start as close together as possible.  So that is why set_power(), etc are broken out like this, it is so I can set all of the parameters first when working with pairs of motors.

ev3dev/core.py
dwalton76: I could cache it here or we could do something in the Device class (or maybe the Motor class) where it is aware of what properties will never change and cache it there.  If we want to do the latter I'll work on that as a separate PR.

ev3dev/core.py
dwalton76: Shouldn't the scale be `[-1 * self.max_speed, self.max_speed]`?

ev3dev/core.py
dwalton76: Yeah they seem pretty silly to me too but David had screenshots for these blocks so I coded them up.  I am fine with chopping them.

ev3dev/core.py
ddemidov: The stock Lego interface measures speed/power in percentage of `max_speed`, and the `max_speed` is never exposed.

dwalton76: ah I see now...will clean this up

dlech: I share the same sentiments as @laurensvalk that we should be using scientific units where possible in places where LEGO uses 0..100 or -100..100

https://github.com/ev3dev/ev3dev/issues/668#issuecomment-224670770

So, in this case, our speed parameter should be in degrees per second or RPM.

ev3dev/core.py
dlech: How about making them private (i.e. leading underscore)?

ev3dev/core.py
dlech: RE: properties that never change. See:

* https://github.com/ev3go/ev3dev/issues/53
* https://github.com/ev3go/ev3dev/issues/54
* https://github.com/ev3go/ev3dev/issues/55

ev3dev/core.py
```diff
@@ -955,6 +958,94 @@ def wait_while(self, s, timeout=None):
...
+        rotations = float(degrees / self.count_per_rot)
+        self.set_position_rotations(power, rotations)
+
+    def set_brake(self, brake):
+        if brake:
+            self.stop_action = self.STOP_ACTION_BRAKE
+        else:
+            self.stop_action = self.STOP_ACTION_COAST
+
+    def set_time(self, ms):
```
dlech: Function calls in python are expensive. Here and in several other places, your are just setting an attribute. It would be more efficient to just set the attribute directly.

ev3dev/core.py
```diff
@@ -955,6 +958,94 @@ def wait_while(self, s, timeout=None):
...
+        else:
+            self.position_sp = self.position - int(rotations * self.count_per_rot)
+
+    def set_position_degrees(self, power, degrees):
+        rotations = float(degrees / self.count_per_rot)
+        self.set_position_rotations(power, rotations)
+
+    def set_brake(self, brake):
+        if brake:
+            self.stop_action = self.STOP_ACTION_BRAKE
```
dlech: As mentioned elsewhere, brake in EV3-G == hold in ev3dev

ev3dev/core.py
dlech: Also, I wouldn't mined a default value for the brake parameter here as `True` since this is what gets used most of the time.

ev3dev/core.py
```diff
@@ -955,6 +958,94 @@ def wait_while(self, s, timeout=None):
...
+    def on_for_degrees(self, power, degrees, brake):
+        """
+        Rotate the motor at 'power' for 'degrees'
+        """
+        self.set_power(power)
+        self.set_position_degrees(power, degrees)
+        self.set_brake(brake)
+        self.run_to_abs_pos()
+
+    def on_for_seconds(self, power, seconds, brake):
```
dlech: All of the `on_for_*` methods should wait for completion before returning.

e.g., this method should not use `self.run_time()`, it should

    self.run_forever()
    time.sleep(seconds)
    self.stop()

ev3dev/core.py
```diff
@@ -955,6 +958,94 @@ def wait_while(self, s, timeout=None):
...
+        Rotate the motor at 'power' for forever
+        """
+        self.set_power(power)
+        self.run_forever()
+
+    def off(self, brake):
+        self.set_brake(brake)
+        self.stop()
+
+    def get_rotations(self):
```
dlech: How about using python properties for these getters?

ev3dev/core.py
dlech: On second thought, my suggestion would not be a good implementation when there are multiple threads.

Unfortunately, we don't currently have a way to be notified by the motor driver that the current operation was interrupted.

ev3dev/core.py
dwalton76: yep can make them private

ev3dev/core.py
dwalton76: Filed issued #365 to track this

ev3dev/core.py
dwalton76: ack will change this

ev3dev/core.py
dwalton76: ack will chop `set_power()` and `set_time()`

ev3dev/core.py
dwalton76: wait if the EV3-G on_for_rotations block only takes power as a percentage of max_speed and we take power as the scientific unit then our attempt at a 1:1 API isn't going to match

ev3dev/core.py
dwalton76: We could have a `wait_until_stopped` arg and if it is True we call `self.wait_until_not_moving()`.  But then our API would no longer match the EV3-G API exactly.  We could just document how to use `self.wait_until_not_moving()`

ev3dev/core.py
dwalton76: I hate decorators :)  Since we do use `@property` elsewhere though I guess it does make sense to use them here....will fix

dlech:
```python
def rotations(self):
    pass

rotations = property(rotations)
```

No decorator. :laughing:

ev3dev/core.py
WasabiFan: Why can't we wait until `'running' not in self.state`? Does the driver consider itself "running" while holding?

dwalton76: In EV3-G if you tell a motor to run for 3 rotations does it start the motor and move on to the next block or does it wait for the three rotations to finish?  If it moves on then I think we are good with what we have now, it if blocks though then we need to add something like
```
    self.wait_until('running', timeout=100)
    self.wait_until_not_moving()
```

WasabiFan: It's a blocking call. It starts the motors and only returns to the caller once the motor has turned to the specified position. EV3-G programs are very commonly just a single sequence of motor blocks with some branching (drive forward, turn, drive forward, check sensor, etc.).

dwalton76: ok lets have it block by default but have an optional arg that can be passed to not block

ev3dev/core.py
```diff
@@ -958,93 +962,84 @@ def wait_while(self, s, timeout=None):
...
-
-    def set_position_rotations(self, power, rotations):
+    def _set_position_rotations(self, power, rotations):
         if power > 0:
             self.position_sp = self.position + int(rotations * self.count_per_rot)
         else:
             self.position_sp = self.position - int(rotations * self.count_per_rot)
 
-    def set_position_degrees(self, power, degrees):
+    def _set_position_degrees(self, power, degrees):
```
WasabiFan: Can we have `rotations` call `degrees` instead? It makes more sense to me to do everything in terms of the "smaller" units and convert from the "larger" ones as needed. Or, it might be optimal to just make them operate independently so we avoid unnecessary nesting for execution speed.

ev3dev/core.py
```diff
@@ -958,93 +962,84 @@ def wait_while(self, s, timeout=None):
...
-            self.stop_action = self.STOP_ACTION_BRAKE
+            self.stop_action = self.STOP_ACTION_HOLD
         else:
             self.stop_action = self.STOP_ACTION_COAST
 
-    def set_time(self, ms):
-        self.time_sp = ms
-
-    def on_for_rotations(self, power, rotations, brake):
+    def on_for_rotations(self, power, rotations, brake=True, block=True):
```
WasabiFan: What do others think about calling it "speed" instead of "power"? This would only apply if we switched to using scientific units.

ev3dev/core.py
dwalton76: 1 rotation, 360 degrees....rotation is the smaller unit :)

ev3dev/core.py
dwalton76: yeah the use of `power` is some places and `speed` in other is already making my head spin :)  I think the EV3-G blocks use `power`?

ev3dev/core.py
WasabiFan: If you need more degrees than rotations to represent the same quantity, the quantity represented by one degree is smaller than the quantity represented by one rotation.

dwalton76: fixed

ev3dev/core.py
```diff
@@ -955,6 +962,91 @@ def wait_while(self, s, timeout=None):
         """
         return self.wait(lambda state: s not in state, timeout)
 
+    def _set_position_rotations(self, max_speed_pct, rotations):
```
WasabiFan: Why `max_speed_pct`? I read that to mean "the maximum speed (percentage)" as opposed to "percentage of the max speed". TBH, I think it's an unnecessary distinction as most people would just consider it "speed percentage".

WasabiFan:
We also need to consider when we're updating docs+demos for these changes. I want to make sure that we don't lose track of that work.

P.S. I'm hoping I can find time to take on a few of the pending features; we'll see if I can make that work over the weekend. I don't want to take something then delay it if I get busy.

ev3dev/core.py
dwalton76: How about 'speed_pct'?

WasabiFan: I like that.

ev3dev/core.py
```diff
@@ -44,6 +44,7 @@
 import array
 import mmap
 import ctypes
+import logging
```
ddemidov: I think you should put logging into a separate PR. Every import makes startup time a bit longer, so we should only do this if necessary.

ev3dev/core.py
```diff
@@ -64,6 +67,10 @@
 INPUT_AUTO = ''
 OUTPUT_AUTO = ''
 
+# The number of milliseconds we wait for the state of a motor to
+# update to 'running' in the "on_for_XYZ" methods of the Motor class
+WAIT_RUNNING_TIMEOUT = 100
```
ddemidov, line 61: Do we really need it? Should not it be solved by #241 and the related changes in kernel drivers (ev3dev/lego-linux-drivers#57)?

ddemidov:
I wonder if introducing properties for various units would make a lot of these functions easier (and may be even unnecessary). For example, we could have
```py
@property
def degrees_sp(self):
    return self.position_sp * self.POS2DEG_FACTOR
@degrees_sp.setter
def degrees_sp(self, value):
    self.position_sp = value * self.DEG2POS_FACTOR
```
and then same thing for `rotations_sp`, `power_sp` (percentage of max_speed_sp), etc.

ev3dev/core.py
dwalton76: Fair point, I'll back out the logging parts that are in here and file an issue to chat about what logging we want in core.py.

ev3dev/core.py
dwalton76: It is, what I found is that if you do
```
    m.run_to_abs_pos(...)
    m.wait_until_not_moving()
```
The run_to_abs_pos will update the `command` file but there is some finite amount of time before the kernel updates the `state` file to `running`.  So you really have to wait until the motor starts moving before you can wait for it to stop moving.

dwalton76:
I think the only thing left in this PR to address is how to specify speed_pct vs speed_dps vs speed_rps so I'm going to go ahead and merge this one in and open a new issue for the speed thing.

Issue #361, opened 2017-09-20
EV3-G API: touch sensor


dwalton76:
Need to get screenshots of the touch sensor blocks

dlech:
![image](https://user-images.githubusercontent.com/963645/30651706-21a04a58-9dec-11e7-8b54-ce355ca002f0.png)

dwalton76:
@dlech wrong image :)

dlech:
Buttons are a touch sensors of sorts :wink: 

Here is the correct screenshot:

![image](https://user-images.githubusercontent.com/963645/30656253-19319108-9df9-11e7-98a2-dad21965a7a4.png)

laurensvalk:
---> [This](https://github.com/ev3dev/lms2012-compat/issues/51) may be relevant when looking to replicate EV3-G, including or excluding original bugs.

If that post doesn't make sense, ask away :)

dwalton76:
@dlech does ev3dev support the `bumped` value?  I've been `cat value0` over and over while touching the sensor but I only ever see 0 or 1

WasabiFan:
@dwalton76 ev3dev's built-in touch sensor driver doesn't support `bumped`; it only reports whether or not the button is pressed (which I think is the correct behavior). If we want to support that logic, we'll have to implement it in Python. I'd say we should just omit it for now.

dwalton76:
works for me

PR #367, opened 2017-09-20
EV3-G API: touch sensor add is_released

dwalton76:


dwalton76:
So I added this for TouchSensor and was about to start on `wait_for_pressed()` and `wait_for_released()` (because those would be handy) but then noticed that a huge chunk of the sensors code is autogenerated from the templates...so the code in this PR will be over-written the next time some uses the templates to build core.py.

On our call the other night we kicked around the idea of life without the templates and autogen.  I wasn't around when it was written but it sounds like things were changing often enough back then to make it worth it.  Today though it doesn't feel worth it to me, the reasons being:
- There have only been 5 commits to it all year, a good indication that things are not churning at nearly the rate that they used to
- I find it difficult to use.  I tried running it again today and it failed despite there not being any checkins since the last time I ran it and it worked.
- It clutters the code base with `# ~autogen generic-class classes.motor>currentClass` and `# ~autogen` comments
- It makes it way more complicated than it needs to be to add code like I have in this PR

The only thing we have to do to move away from it in ev3dev-lang-python is
- remove the `# ~autogen` lines
- remove the `ev3dev-lang` submodule

It could still be used for other ev3dev-lang-XYZ repos if they wish to use it.

Thoughts? Objections? +1s? :)

dwalton76:
Making sure @ddemidov @dlech @WasabiFan @ensonic see this

ddemidov:
I am ok with this, just make sure to run the autogen one last time before you strip it off (I think your stretch-related commits in ev3dev-lang have not yet been applied).

WasabiFan:
I think @dwalton76 was having issues running the script and I haven't sat down to debug it yet.

ddemidov:
I just tried to do this myself and noticed that @dwalton76 already did this :). Sorry for the noise.

dwalton76:
cool, removal of ev3dev-lang and autogen via PR #368

ddemidov (approved):

dwalton76:
Am holding off on merging this in...will implement the following this weekend
```
wait_for_pressed()
wait_for_released()
wait_for_bumped()
```

ev3dev/core.py
```diff
@@ -1779,24 +1779,20 @@ class TouchSensor(Sensor):
...
     SYSTEM_DEVICE_NAME_CONVENTION = Sensor.SYSTEM_DEVICE_NAME_CONVENTION
 
-    def __init__(self, address=None, name_pattern=SYSTEM_DEVICE_NAME_CONVENTION, name_exact=False, **kwargs):
-        super(TouchSensor, self).__init__(address, name_pattern, name_exact, driver_name=['lego-ev3-touch', 'lego-nxt-touch'], **kwargs)
-        self.auto_mode = True
-
-
     #: Button state
     MODE_TOUCH = 'TOUCH'
+    MODES = ('TOUCH',)
```
WasabiFan: Remove trailing comma

WasabiFan: Also, can we use the other constant as the value here?

ev3dev/core.py
```diff
@@ -1814,6 +1810,46 @@ def is_pressed(self):
...
+
+        if self._poll is None:
+            self._value0 = self._attribute_file_open('value0')
+            self._poll = select.poll()
+            self._poll.register(self._value0, select.POLLPRI)
+
+        while True:
+            self._poll.poll(timeout_ms)
+
+            if wait_for_press:
```
WasabiFan: Can we reduce this to a single Boolean expression?

ev3dev/core.py
dwalton76: The trailing comma is needed for single entry tuples
```
>>> MODES = ('TOUCH',)
>>> MODES[0]
'TOUCH'
>>> 
>>> MODES = ('TOUCH')
>>> MODES[0]
'T'
>>> 
```

dwalton76: Will change it to `MODES = (MODE_TOUCH,)`

ev3dev/core.py
dwalton76: Meaning `if (wait_for_press and self.is_pressed) or (not wait_for_press and self.is_released)`?  I could go either way but I think the current layout is a tiny bit easier to read.

ev3dev/core.py
WasabiFan: 😒 Python is... a _special_ language.

ev3dev/core.py
WasabiFan: `wait_for_press ^ self.is_released` or `self.is_pressed == wait_for_press`? I think either should work, feel free to switch inversions for readability if that's useful.

ev3dev/core.py
```diff
@@ -1810,6 +1806,51 @@ def is_pressed(self):
...
+
+    def wait_for_bump(self, timeout_ms=None):
+        """
+        Wait for the touch sensor to be pressed down and then released.
+        Both actions must happen within timeout_ms.
+        """
+        start_time = time.time()
+
+        if self.wait_for_pressed(timeout_ms):
+            timeout_ms -= int((time.time() - start_time) * 1000)
```
ddemidov: What happens here is timeout_ms is None? I think something like:
```py
>>> None - 4
TypeError: unsupported operand type(s) for -: 'NoneType' and 'int'
```

dwalton76: yep, good catch! Fixed now

ev3dev/core.py
dwalton76: ah I follow you now, fixed.

dwalton76:
ok I think that is everything from the code review comments so am going to merge this one and we can do any further tweaking here via a new PR/issue.

This is for issue #361

Issue #355, opened 2017-09-19
EV3-G API: color sensor


dwalton76:
![color-sensor-01](https://user-images.githubusercontent.com/468309/30606514-f08eb9d8-9d3f-11e7-85d4-d53b7ca1546d.png)

dwalton76:
For the two calibrate methods per
http://docs.ev3dev.org/projects/lego-linux-drivers/en/ev3dev-stretch/sensor_data.html

_[28] | This mode is not usable. When in COL-CAL mode, the color sensor does not respond to the keep-alive sent from the EV3 brick. As a result, the sensor will time out and reset._

so will skip those unless the quote above is no longer accurate.


I'm not sure what all we should add here...`wait_for_color(...)` might be handy.  What methods would make line followers easier to code?

dwalton76:
Do you guys see the RGB values that come from the sensor as very very dark?  I can point mine at the white side of a rubiks cube and it comes back with numbers like (34, 58, 46) on 0-255 scale.  That is basically dark green:
https://www.w3schools.com/colors/colors_converter.asp

I'm wondering if this thing is broken.

dwalton76:
Prior to this commit `ColorSensor.rgb()` took the raw RGB values and scaled them to 0-255 based on a 0-1020 scale (since that is the official range of the values returned by the sensor).  The color sensor never returns anything close to 1020 though so the end result was that we got very dark RGB values out of `rgb()`.  A typical MINDCUBER scan of a rubiks cube would look like this:

![color-sensor-before](https://user-images.githubusercontent.com/468309/31440882-67b59556-ae5f-11e7-8226-2a0d15e76932.png)

Now we have a `calibrate_white()` function the user can call (to be called when the color sensor is looking at something white) to figure out what are the maximum RGB values we can expect to see. `ColorSensor.rgb()` will then scale the results to 0-255 based on those max values.  The result is that you get much brighter/accurate data:

![color-sensor-after](https://user-images.githubusercontent.com/468309/31440982-c743af9e-ae5f-11e7-9490-1932c4990eb5.png)

Note that the two pics above are of different rubiks cubes but you get the idea, the colors in the second one are much more accurate than the first.

dlech:
FYI, converting to HSV or some other colorspace would be even more accurate.

dwalton76:
[...]

PR #371, opened 2017-09-23
ColorSensor: Move rgb() from helper.py to core.py

dwalton76:


dwalton76:
For issue #355

ev3dev/core.py
```diff
@@ -2362,6 +2362,17 @@ def raw(self):
         return self.value(0), self.value(1), self.value(2)
 
     @property
+    def rgb(self):
+        """
+        Same as raw() but RGB values are scaled to 0-255
+        """
+        (red, green, blue) = self.raw
+
+        return (int((red * 255) / 1020),
```
WasabiFan: Why integers? It seems like we should return floats given that the sensor theoretically returns 4x the precision as can be represented in this range.

(yes, this is old code, but I'd like to take the chance to make improvements)

dwalton76: Displaying RGB values in html or passing them to a lot of color libraries need them to be integers.

WasabiFan: If the sensor's accuracy is low enough that 1/4th the resolution isn't a detriment, I have no problem with this; otherwise, I think we either need to heavily document the behavior or add a new accessor/method.

dwalton76: The accuracy of mine is pretty horrible, I think we are ok dropping the 1/4th resolution

WasabiFan: That certainly sounds familiar 😁 These LEGO sensors are never very accurate.

dlech: It would probably be useful to normalize the values here as well. For example, the blue is always significantly less than red and green. Also, you will never get anywhere near 1020, so the scaled values currently will never be anywhere near 255.

Related discussion: https://www.facebook.com/groups/legomindstorms/permalink/662009207280642/

dwalton76: Brainstorming....we could add a `calibrate_to_white()` method where we note the RGB values and then `rgb()` could scale from 0-255 based on the RGB values we stored when `calibrate_to_white()` ran.

dwalton76: That video is interesting, I don't get colors anywhere close to that bright.  Let me flip back to the stock ev3 software and try calibrating my sensor.

ev3dev/core.py
```diff
@@ -2361,6 +2367,38 @@ def raw(self):
...
+    @property
+    def rgb(self):
+        """
+        Same as raw() but RGB values are scaled to 0-255
+        """
+        (red, green, blue) = self.raw
+
+        return (min(int((red * 255) / self.red_max), 255),
+                min(int((green * 255) / self.green_max), 255),
+                min(int((blue * 255) / self.blue_max), 255))
```
dwalton76, line 2408: Scaling to red_max, green_max, blue_max instead of 1020 makes a HUGE difference in the accuracy of the RGB values returned.  I scanned my rubiks cube and displayed the colors returned from rgb() in a web page and they looked pretty good.  Before they all looked super super dark.

ev3dev/core.py
```diff
@@ -2351,6 +2353,10 @@ def color(self):
         return self.value(0)
 
     @property
+    def color_name(self):
+        return self.COLORS[self.color]
+
+    @property
     def raw(self):
         """
         Red, green, and blue components of the detected color, in the range 0-1020.
```
WasabiFan: Should we add a comment here pointing to the `calibrate_white` method?

ev3dev/core.py
```diff
@@ -2351,6 +2353,10 @@ def color(self):
         return self.value(0)
 
     @property
+    def color_name(self):
```
WasabiFan, line 2356: doc comment?

WasabiFan (approved):
Minor suggestions.

ev3dev/core.py
dwalton76: will add one

ev3dev/core.py
dwalton76: `raw()` doesn't do anything with it though...it only impacts `rgb()`

WasabiFan: Something like:

> The colors returned by this method may be unexpectedly dark. If this is an issue, check out the rgb() and calibrate_white() methods.

dwalton76: They scale from 0-1020 though so if you displayed them as if they were 0-255 they would appear fairly bright (assuming none went over the 255 limit).  What I'm trying to say is whether they appear dark or not is really up to how the user is using the returned values.  If the user is scaling them much like what we do in rgb() then they wouldn't appear that dark.

WasabiFan: My issue is that the doc comment says the range is "0-1020" yet the values will never get up that high. I think that expectation should be laid out in the comment here and in the `rgb` method.

dwalton76: ah I follow you now, adding this
```
+        Red, green, and blue components of the detected color, officially in the
+        range 0-1020 but the values returned will never be that high. We do not
+        yet know why the values returned are low, but pointing the color sensor
+        at a well lit sheet of white paper will return values in the 250-400 range.
+
+        If this is an issue, check out the rgb() and calibrate_white() methods.
```

WasabiFan: Perfect, I'm happy with this PR.

PR #373, opened 2017-09-23
Throw friendly error when device is disconnected

WasabiFan:


ev3dev/core.py
```diff
@@ -225,6 +225,8 @@ def _raise_friendly_access_error(self, driver_error, attribute):
                 else:
                     raise ValueError("The given speed value was out of range. Max speed: +/-" + str(max_speed)) from driver_error
             raise ValueError("One or more arguments were out of range or invalid") from driver_error
+        elif driver_error.errno == errno.ENODEV:
+            raise Exception("The device is no longer connected") from driver_error
```
ddemidov: Why 'no longer'? Do we know it was ever connected?

WasabiFan: Currently, if it was never connected, the `connected` property is false and we throw an exception at a higher level. There's an `if` somewhere around  here which handles that case.

dwalton76: How about:
```
raise Exception("%s is no longer connected" % self) from driver_error
```
that way the exception has the `__str__` of the device that was disconnected

ev3dev/core.py
```diff
@@ -197,12 +197,12 @@ def _get_attribute(self, attribute, name):
     def _set_attribute(self, attribute, name, value):
         """Device attribute setter"""
         if self.connected:
-            if None == attribute:
-                attribute = self._attribute_file_open( name )
-            else:
-                attribute.seek(0)
-
             try:
+                if None == attribute:
```
dwalton76: Do we follow PEP-8?  If so we should do `if attribute is None` here.

ev3dev/core.py
WasabiFan: I haven't paid attention to how inheritance works in Python... Will that have the expected behavior? I know that `__class__` is the type the method was defined in and not the runtime type of the enclosing object.

ev3dev/core.py
WasabiFan: Yeah, this is existing code. I'll update it.

ev3dev/core.py
dwalton76: The Device class has a `__str__` that will be used to determine what to print whenver you have to print the object in a string

WasabiFan: The question I was asking was a technical one about the language. I'll figure it out and update as needed.

dwalton76 (approved):
Looks good

PR #374, opened 2017-09-24
EV3-G API infrared sensor

dwalton76:


dwalton76:
We have the InfraredSensor functionality split into three classes
- InfraredSensor
- RemoteControl
- BeaconSeeker

but the EV3-G block do all of this via on InfraredSensor block. I like the way we did it much more, it feels much cleaner.  Anyway,  I moved the functionality from RemoteControl and BeaconSeeker up to InfraredSensor but did so in a way to make RemoteControl and BeaconSeeker backwards compatible.

ev3dev/core.py
```diff
@@ -2028,6 +2030,15 @@ def driver_name(self):
         return value
 
     @property
+    def auto_mode(self):
+        return True
```
WasabiFan: Why? I can't imagine we have anyone using this property... I think we should just remove it.

ev3dev/core.py
```diff
@@ -2038,7 +2049,17 @@ def mode(self):
...
+        This will be called often so we cache the current mode in
+        self._mode_value to avoid some file IO.
+        """
+
+        # If 'mode' is already set to 'value' there is nothing to do.
+        if value == self._mode_value:
+            return
+
         self._mode = self.set_attr_string(self._mode, 'mode', value)
+        self._mode_value = value
```
WasabiFan: Don't you need to update the getter as well?

WasabiFan (commented):
Would it be possible to re-arrange the classes as they were before so the diff isn't so weird?

ev3dev/core.py
dwalton76: We don't use it in any of the demos...I would be ok with chopping it

ev3dev/core.py
dwalton76: True I could just have it return self._mode_value

dwalton76:
Looking into the api_test failure...

ev3dev/core.py
```diff
@@ -1936,6 +1936,7 @@ class Sensor(Device):
     '_decimals',
     '_driver_name',
     '_mode',
+    '_mode_value',
```
ddemidov: The PR title says 'Infrared sensor', but this looks like a move to introduce caching for the mode value. Should this be split into a couple of PRs?

dwalton76: Will pull this part out into a new PR

dwalton76:
> Would it be possible to re-arrange the classes as they were before so the diff isn't so weird?

yep will put them back where they were, we can move them around later via another PR

dwalton76:
Should be easier to read now

ev3dev/core.py
```diff
@@ -2584,8 +2584,8 @@ def rate_and_angle(self):
 
         return self.value(0), self.value(1)
 
-class InfraredSensor(Sensor):
 
+class InfraredSensor(Sensor):
```
WasabiFan: Should I be worried?

ev3dev/core.py
```diff
@@ -2615,27 +2610,121 @@ def __init__(self, address=None, name_pattern=SYSTEM_DEVICE_NAME_CONVENTION, nam
...
     MODE_IR_CAL = 'IR-CAL'
 
-
     MODES = (
-      'IR-PROX',
-      'IR-SEEK',
-      'IR-REMOTE',
-      'IR-REM-A',
-      'IR-CAL',
+      MODE_IR_PROX,
```
WasabiFan, line 2668: :+1:

ev3dev/core.py
```diff
@@ -2615,27 +2610,121 @@ def __init__(self, address=None, name_pattern=SYSTEM_DEVICE_NAME_CONVENTION, nam
...
+      MODE_IR_SEEK,
+      MODE_IR_REMOTE,
+      MODE_IR_REM_A,
+      MODE_IR_CAL
     )
 
+    _BUTTON_VALUES = {
+            0: [],
```
WasabiFan: Could we get a comment explaining what these indicies are?

ev3dev/core.py
```diff
@@ -2615,27 +2610,121 @@ def __init__(self, address=None, name_pattern=SYSTEM_DEVICE_NAME_CONVENTION, nam
...
+            6: ['red_up', 'blue_down'],
+            7: ['red_down', 'blue_up'],
+            8: ['red_down', 'blue_down'],
+            9: ['beacon'],
+            10: ['red_up', 'red_down'],
+            11: ['blue_up', 'blue_down']
+            }
+
+    #: Handles ``Red Up`` events.
+    on_red_up = None
```
WasabiFan, line 2695: Are the buttons color-coded? I have never actually used the remote after taking it out of the box :laughing:

ev3dev/core.py
```diff
@@ -2615,27 +2610,121 @@ def __init__(self, address=None, name_pattern=SYSTEM_DEVICE_NAME_CONVENTION, nam
...
         A measurement of the distance between the sensor and the remote,
         as a percentage. 100% is approximately 70cm/27in.
         """
+        self.mode = self.MODE_IR_PROX
+        return self.value(0)
 
-        if self.auto_mode:
-            self.mode = self.MODE_IR_PROX
+    # BeaconSeeker
+    def heading(self, channel=1):
```
WasabiFan: What's the idea behind making these methods instead of property getters?

ev3dev/core.py
```diff
@@ -2615,27 +2610,121 @@ def __init__(self, address=None, name_pattern=SYSTEM_DEVICE_NAME_CONVENTION, nam
...
+        """
+        Returns heading (-25, 25) to the beacon on the given channel.
+        """
+        self.mode = self.MODE_IR_SEEK
+        channel = max(1, min(4, channel)) - 1
```
WasabiFan: Warn/error if the channel is out-of-bounds?

ev3dev/core.py
```diff
@@ -2615,27 +2610,121 @@ def __init__(self, address=None, name_pattern=SYSTEM_DEVICE_NAME_CONVENTION, nam
...
+        return self.value(channel * 2)
+
+    def distance(self, channel=1):
+        """
+        Returns distance (0, 100) to the beacon on the given channel.
+        Returns -128 when beacon is not found.
```
WasabiFan: Well duh, _of course_ it returns -128 :laughing: Do we want to catch that and return `None` instead? I think no one actually reads our docs :wink:

ev3dev/core.py
```diff
@@ -2615,27 +2610,121 @@ def __init__(self, address=None, name_pattern=SYSTEM_DEVICE_NAME_CONVENTION, nam
...
+
+    def heading_and_distance(self, channel=1):
+        """
+        Returns heading and distance to the beacon on the given channel as a
+        tuple.
+        """
+        return (self.heading(channel), self.distance(channel))
+
+    # RemoteControl
+    def red_up(self, channel=1):
```
WasabiFan: `is_red_up_pressed`? or `red_up_pressed`? I originally read this thinking that this was asking whether "red" was released.

ev3dev/core.py
```diff
@@ -3046,100 +3135,49 @@ def buttons_pressed(self):
...
-
-    #: Handles ``Blue Up`` events.
-    on_blue_up = None
-
-    #: Handles ``Blue Down`` events.
-    on_blue_down = None
-
-    #: Handles ``Beacon`` events.
-    on_beacon = None
+    def __init__(self, sensor=None, channel=1):
```
WasabiFan: Is this the order we want it in? Or should `channel` be first? I'd imagine that providing a custom sensor will be quite rare. Not that it _really_ matters as you can reference it by name anyway.

ev3dev/core.py
dwalton76: I had to look at this for a minute to figure out what in the world this change was.  It is because I deleted the blank line between the `class InfraredSensor` and the `"""`.  It is odd that it displayed it that way though

ev3dev/core.py
dwalton76: How about
```
The following are all of the various combinations of button presses for the remote
control.  The key/index is the number that will be written in the attribute file to
indicate what combination of buttons are currently pressed. 
```

WasabiFan: Looks good!

ev3dev/core.py
dwalton76: They are all gray but there is a red mark on the remote between the two buttons on the left and a blue mark between the two on the right.

ev3dev/core.py
dwalton76: It is so you can pass the `channel` that you want...the EV3-G block has this.

WasabiFan: Oh, duh. Makes sense.

ev3dev/core.py
dwalton76: yeah I'll put an assert in here...it would be really confusing behavior if someone passed in channel 5

ev3dev/core.py
dwalton76: Not sure...what does the EV3-G block do in this scenario?

ev3dev/core.py
dwalton76: yeah the up/down naming convention is a little odd.  I don't think we should change this though, we use the RemoteControl a fair amount in the demos, I imagine these are used a lot.

ev3dev/core.py
dwalton76: It was that way before so we should probably leave it as-is

ev3dev/core.py
```diff
@@ -2615,27 +2610,130 @@ def __init__(self, address=None, name_pattern=SYSTEM_DEVICE_NAME_CONVENTION, nam
...
+            9: ['beacon'],
+            10: ['red_up', 'red_down'],
+            11: ['blue_up', 'blue_down']
+            }
+
+    #: Handles ``Red Up`` events.
+    on_red_up = None
+
+    #: Handles ``Red Down`` events.
+    on_red_down = None
```
ddemidov, line 2698: These are user-defined event handling functions that should fire up during call to `process()` method (and by the way, I don't see the method here). As in
```py
def beep_when_pressed(state):
    if state:
      Sound.beep()

rc.on_red_up = beep_when_pressed

while True:
  rc.process()
```

I don't see how these would work as is, since we are merging `RemoteControl` with the `InfraredSensor`, and there is no way to set the channel here (we used to set channel in the constructor of `RemoteControl`).

We should probably change these into dictionaries: `on_press[1] = beep_when_pressed` (not lists because channel numbers start at 1).

ev3dev/core.py
ddemidov: Returning 'None' makes sense.

ev3dev/core.py
```diff
@@ -3046,100 +3144,49 @@ def buttons_pressed(self):
         return pressed
 
 
-class RemoteControl(ButtonBase):
+class RemoteControl(ButtonBase, InfraredSensor):
```
ddemidov: I think we should either drop `RemoteControl` and `BeakonSeeker` or keep everything as is. The classes do not do anything that `InfraredSensor` already does not, so why keep them?.

ev3dev/core.py
ddemidov: The PR makes the class a subclass of `InfraredSensor`. Why do we need the `sensor` parameter (and `_sensor` field) here at all?

ev3dev/core.py
ddemidov: Re `process()` method: in the `RemoteControl` class it was inherited from `ButtonBase`. If we do want to merge `RemoteControl` with `InfraredSensor`, we should probably inherit from `ButtonBase` as well.

dwalton76: gotcha, will test/work on this part this evening

ev3dev/core.py
dwalton76: ack

ev3dev/core.py
dwalton76: agreed they aren't doing much but we need them for backwards compatibility

ev3dev/core.py
dwalton76: for backwards compatibility

ddemidov: I mean, the `BeaconSeeker` class here already is `InfraredSensor` (since it inherits from it). So why store another instance of `InfraredSensor` in `_sensor`?

dwalton76: Because our current API allows the user to pass in a InfraredSensor object to the BeaconSeeker and use that as the _sensor.  If we are taking the stance that the new API is not backwards compatible with the old API then we can certainly clean all of this up a good bit. My understanding was that we wanted to keep the old API around for backward compatibility.

ddemidov: Then we should not derive from `InfraredSensor`.

WasabiFan: I think we should simply be deliberate about it. If there's a good reason to make a breaking change, I would rather do it now for the first Stretch version than later on.

dwalton76: Given how much code/API we can clean up if we are not backwards compatible I think it would be worth it.  ok just so we are all on the same page, we are saying drop the `RemoteSensor` and `BeaconSeeker` classes and put all functionality in the `InfraredSensor` class yes?

WasabiFan: I think that seems reasonable; how does this map to EV3-G?

dwalton76: EV3-G has the equivalent of our `InfraredSensor`, `RemoteSensor`, and `BeaconSeeker` all in a single block.  So we put all of the logic in `InfraredSensor`, delete `RemoteSensor` and `BeaconSeeker` and then we are 1:1 with the EV3-G block.

WasabiFan: :+1: I support this

ev3dev/core.py
dwalton76: Now that we are not worried about being backwards compatible do we want to change the names of `red_up`, `red_down`, etc?  IMO they are pretty confusing. Brainstorming....
- We could label them `top_left`, `bottom_left`, `top_right`, `bottom_right`
- We could rename `red_up()` to `is_top_left_pressed()`, etc 
- We could rename `on_red_up` to `on_top_left_press`, etc

ev3dev/core.py
dlech: python zen: "flat is better than nested"

having a single InfraredSensor class sounds like a good call.

ev3dev/core.py
dwalton76: This is working now

ev3dev/core.py
WasabiFan: We should start keeping track of all breaking changes; do you guys want to do that in the changelog file as we go or add a "changelog" section to the README? Or maybe create an issue thread to post into? I'm up for whatever's easiest for you guys.

ddemidov: We can start a draft of new release [here](https://github.com/rhempel/ev3dev-lang-python/releases). It seems README could grow too long too fast if we start putting changes there. Another idea is we could add another section in the documentation for this.

WasabiFan: Interesting idea! I just created a release where we can try out tracking this; as we merge PRs we should make sure to update it. If this doesn't go well we can try out a more conventional file approach. https://github.com/rhempel/ev3dev-lang-python/releases/tag/untagged-f43144149d22a7d3cc09

Note that the link might  break once we add a version tag.

ev3dev/core.py
ddemidov: How exactly does it work now? Where do we specify channel to use with the handles?

ddemidov:
> Now that we are not worried about being backwards compatible do we want to change the names of red_up, red_down, etc? IMO they are pretty confusing. Brainstorming....

I like `left_top` or `red_top`, side (left/red or right/blue) in front. I see the side being more important, because there is builtin color coding, and the whole side (red/left or blue/right) is usually dedicated to a motor.

ev3dev/core.py
dwalton76: The `channel` has to be passed to the `process()` call.  Here is `RemoteControlledTank` from `helper.py`
```
class RemoteControlledTank(MoveTank):

    def __init__(self, left_motor_port, right_motor_port, polarity='inversed', speed=400, channel=1):
        MoveTank.__init__(self, left_motor_port, right_motor_port)
        self.set_polarity(polarity)

        left_motor = self.motors[left_motor_port]
        right_motor = self.motors[right_motor_port]
        self.speed_sp = speed
        self.remote = InfraredSensor()
        self.remote.on_red_up = self.make_move(left_motor, self.speed_sp)
        self.remote.on_red_down = self.make_move(left_motor, self.speed_sp* -1)
        self.remote.on_blue_up = self.make_move(right_motor, self.speed_sp)
        self.remote.on_blue_down = self.make_move(right_motor, self.speed_sp * -1)
        self.channel = channel

    def make_move(self, motor, dc_sp):
        def move(state):
            if state:
                motor.run_forever(speed_sp=dc_sp)
            else:
                motor.stop()
        return move

    def main(self):

        try:
            while True:
                self.remote.process(self.channel)
                time.sleep(0.01)

        # Exit cleanly so that all motors are stopped
        except (KeyboardInterrupt, Exception) as e:
            log.exception(e)
            self.stop()
```

dwalton76: Speaking of `helper.py`, we are not too far from being able to get rid of it
- All of the motor functionality that I put in there now exists in core.py (it didn't at the time that I wrote it).  I need to update MINDCUB3R to not use the helper.py and then all of the motor code in there can go.
- The wheel classes could be moved into a `wheel.py`, BALANC3R and GYROBOY demos will need to be updated.
- We need to find a home for `RemoteControlledTank`
- `ColorSensorMixin` only added the `rgb()` call and there weren't any objections to adding that method to core.py

dwalton76:
What do you guys say we go ahead and merge this one and we can kick around renaming `red_up`, etc via another PR?  If we rename them we will only be able to do so in certain parts of the code because the kernel will still use `red_up`, `red_down` etc.  I'm worried we might make things more confusing...I didn't think about this until after we had already started kicking around names though that is why I didn't bring it up earlier.

Thank you for all of the feedback on this PR, I think this is the most comments I've seen on a single PR here :)  I like where we ended up, it looks much cleaner now and is 1:1 with the EV3-G block.

ev3dev/core.py
ddemidov: > The channel has to be passed to the process() call. 

So we loose the ability to have different handlers for different channels? I understand that this is the restriction of how `ButtonBase.process()` is implemented, but it seems an important feature (why have several channels, when they behave in the same way?). :thinking:

dwalton76: > So we loose the ability to have different handlers for different channels?

We did not have that capability before either. Prior to this PR you would need to create a RemoteControl object for each channel that you wished to define handlers for. With this PR you need to create an InfraredSensor object for each channel that you wish to define handlers for.

If we think that supporting per-channel handlers in a single object is something folks will use let's add that via another PR.  It shouldn't be too bad, we just have the handler mappings be a dictionary where the channel is the key.

ddemidov: > Prior to this PR you would need to create a RemoteControl object for each channel that you wished to define handlers for. 

Yes, that is correct. But the new implementation of infrared sensor allows to (or claims to) be channel-independent: channel is passed to methods like `heading()` or `process()` instead of being assigned at construction. But for RC handlers to work with different channels users would have to create an instance of `InfraredSensor` per channel, which does not seem logical. I think we could drop `ButtonBase` from parents, and reimplement the functionality so that handlers would take not just state, but also channel, or something to the same effect?

ddemidov: > If we think that supporting per-channel handlers in a single object is something folks will use 

I used this ability for the big shooting robot (EV3STORM?). One channel for leg movement, the other for shooting. It does come handy from time to time.

dwalton76: Good point. Are you ok if we merge this PR and file a new issue to add per-channel capabilities to the handlers?  We are at 10 commits and 60+ comments on this PR and short of the per-channel handlers I think it is in pretty good shape.

ddemidov: Let's do that, it seems in good shape otherwise.

ddemidov (approved):

WasabiFan (approved):
Remember to add this to the draft release for VNext.

ev3dev/core.py
dwalton76: Filed #380 for this part

PR #376, opened 2017-09-25
Use index instead of name for device ID regex group

WasabiFan:
This is the most important change from #346. With this change, the core motors work on the Micropython platform. This opens the door for official Micropython support down the road.

### DIFF
--- a/ev3dev/core.py
+++ b/ev3dev/core.py
@@ -109,7 +109,7 @@ class Device(object):
 
     DEVICE_ROOT_PATH = '/sys/class'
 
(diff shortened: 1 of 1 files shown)


dwalton76 (commented):
I'm missing something, this value returned will be the same it just doing so without using a named group?

WasabiFan:
@dwalton76 Correct, no functional changes. The minimal implementation of `re` that Micropython provides doesn't support named groups.

dwalton76:
Ah makes sense now

dwalton76 (approved):

Issue #365, opened 2017-09-20
Cache properties that never change


dwalton76:
We should cache properties that can never change...things like a motor's max_speed, counts_per_rot, etc

- https://github.com/ev3go/ev3dev/issues/53
- https://github.com/ev3go/ev3dev/issues/54
- https://github.com/ev3go/ev3dev/issues/55

WasabiFan:
I recall discussing this in detail previously... Did we really never implement it? Was there a PR somewhere which was closed?  Or did we decide to cut it and save the related decisions for later?

dwalton76:
Not that I remember but it may have been before my time

WasabiFan:
I could've sworn I was discussing it on a PR with @ddemidov within the last month or two... I can't remember if it was part of said PR and we removed it, or if we were just including snippets in comments.

ddemidov:
> I could've sworn I was discussing it on a PR with @ddemidov within the last month or two...

This does not sound like anything to me :). 

But caching the attributes on motor construction makes sense.

WasabiFan:
Am I going crazy? I'm convinced we discussed this recently. We had a cache object, and if the property value existed in the cache it was used otherwise it was added to the cache... And we ran into some sort of decision to be made but didn't want to block the larger PR so we removed it. 

Maybe I _am_ insane.

ddemidov:
May be I am just forgetting things :). Now that you mentioned it, I do remember some discussion involving python dictionaries, but I still can not remember when/where was that.

ddemidov:
Was it about this function (#329, #330)? https://github.com/rhempel/ev3dev-lang-python/blob/06ae05faf27fc5d33dc5d26742a11ed2bc82e937/ev3dev/core.py#L1664-L1674

WasabiFan:
Ooh, that looks like it! I guess we implemented a specialized version of this but not the generic solution.

kortschak:
I'm considering the design for ev3go now and I'd like input from you all. The approach that I'm leaning towards capturing all the invariant attributes on opening the device (I'll add struct fields to the device types). This differs from the approach above in that there is an upfront cost to getting the values, but it means that I can make the methods to return the values later non-errorable. It also simplifies the handling of memoisation in that I don't need to retain and validity state marker.

Opinions?

WasabiFan:
[...]

PR #382, opened 2017-09-26
Deprecate auto_mode

dwalton76:


dwalton76:
For issue #378

ev3dev/core.py
```diff
@@ -2033,12 +2035,26 @@ def mode(self):
...
 
     @mode.setter
     def mode(self, value):
+        """
+        This will be called often so we cache the current mode in
+        self._mode_value to avoid some file IO.
+        """
+
+        # If 'mode' is already set to 'value' there is nothing to do.
+        if value == self._mode_value:
```
WasabiFan: Do we want a way to bypass this cache logic? Previously, the sysfs accessors were always guaranteed to perform a write operation, whereas now there's no way to handle the case where you intentionally want multiple processes interfacing with the same device. This is a result of the fact that caching was moved down a level from the friendly functions to the getters/setters.

dwalton76: Part of me is wondering if we need it at all.  We didn't have it before and I've never noticed any perf issues with setting the modes.  We aren't setting the mode any more often with the deprecation of auto_mode (assuming nobody ever set this to false).  Maybe we just chop the `mode_value`/caching part of this patch?

dwalton76: @WasabiFan what say you on the `mode_value`...keep or chop?  I'm leaning towards chop and address it as part of #365

WasabiFan: Agreed. I don't like the inconsistency this introduces, especially in light of cases such as the ultrasonic sensor where the only way to send a ping is to set the mode again. We should cut it from here and address it elsewhere.

dwalton76:
Removed `_mode_value`, merging

PR #383, opened 2017-09-28
Use /sys/class/board-info/ to detect platform type

dwalton76:


ev3dev/auto.py
```diff
@@ -1,41 +1,37 @@
...
-        elif machine == 'armv6l':
-            return 'brickpi'
-        else:
-            return 'unsupported'
+                    if key == 'BOARD_INFO_MODEL':
+                        if value == 'FatcatLab EVB':
+                            return 'evb'
+                        elif value == 'LEGO MINDSTORMS EV3':
+                            return 'ev3'
+                        elif value == 'TBD':
```
dwalton76: @dlech if you can send me the BOARD_INFO_MODEL values for brickpi, pistorms, etc I'll go ahead and add those

ev3dev/evb.py
```diff
@@ -0,0 +1,128 @@
+
+"""
+An assortment of classes modeling specific features of the EV3 brick.
```
dwalton76: Will s/EV3/EVB/

ddemidov:
Awesome!

ev3dev/auto.py
```diff
@@ -1,41 +1,37 @@
...
-"""
 import os
 
-filename = '/tmp/current_platform'
-current_platform = None
 
-if os.path.exists(filename):
-    with open(filename, 'r') as fh:
-        current_platform = fh.read().strip()
+def get_current_platform():
```
WasabiFan: This might be useful as a core library method

ev3dev/auto.py
dlech: All the info you need should be on this page: http://docs.ev3dev.org/projects/lego-linux-drivers/en/ev3dev-stretch/board-info.html

dwalton76: perfect

dwalton76:
I need to test the PiStorms part but that is what was needed back in #170 (was closed because we had no way to detect if we were running on a pistorms)

dwalton76:
Any objections to merging this one in?  I need it to go in so I can post the PR for the buttons API

ev3dev/core.py
```diff
@@ -60,6 +60,42 @@
 # update to 'running' in the "on_for_XYZ" methods of the Motor class
 WAIT_RUNNING_TIMEOUT = 100
 
+
+def get_current_platform():
+    """
+    Look in /sys/class/board-info/ to determine the platform type
```
WasabiFan: Document the possible return values?

ev3dev/evb.py
```diff
@@ -0,0 +1,128 @@
...
+        This handler is called by `process()` whenever state of 'right' button
+        has changed since last `process()` call. `state` parameter is the new
+        state of the button.
+        """
+        pass
+
+    @staticmethod
+    def on_enter(state):
+        """
+        This handler is called by `process()` whenever state of 'enter' button
```
WasabiFan, line 74: What is this for?

ev3dev/evb.py
```diff
@@ -0,0 +1,128 @@
+
+"""
+An assortment of classes modeling specific features of the EVB brick.
```
WasabiFan: EVB _brick_? I would suggest "EVB BeagleBone cape" or whatever those things are called.

WasabiFan (approved):

ev3dev/evb.py
dwalton76: I'll chop the ` brick` part...I normally here it referred to as just an `EVB`

ev3dev/evb.py
dwalton76: This is so you can do things like
```
def foobar():
    print("The user pressed ENTER")

btn = Button()
btn.on_enter = foobar

while True:
    btn.process()
```

I'm working on making this event driven as part of the buttons API work...today you have to call it in a while loop like that to make it do what you want.

ev3dev/core.py
dwalton76: k

ev3dev/evb.py
WasabiFan: Oh, I see. Damn, Python is weird :laughing: Is that a common idiom? Is there not an event framework of some sort?

Either way, I don't think it's blocking.

WasabiFan:
LGTM

ev3dev/evb.py
dwalton76: This project is the only place I've seen this approach used...but I've only been writing python for 4 years so who knows maybe it is wildly popular and I just haven't been exposed to it.

Issue #354, opened 2017-09-19
EV3-G API: buttons


dwalton76:
![buttons-01](https://user-images.githubusercontent.com/468309/30606495-e1333fa4-9d3f-11e7-8ac4-10fbb8e1fca5.png)

dwalton76:
What we have right now for `class Button` in `ev3dev/ev3.py` looks pretty good to me.  Is there anything to be done for this one?

PS
How does one use the on_left(), etc methods?

dwalton76:
@WasabiFan @dlech can you spot anything that we are missing to be EV3-G 1:1?  If not lets close this one.

dlech:
How do you wait for a button to be pressed/released/bumped?

dwalton76:
We don't have a way to do that but from looking at the EV3-G block it doesn't either?  It sounds useful though (and we have it for TouchSensor), we can add it.

dlech:
And what about bumped?

dwalton76:
I'll add pressed, released and bumped

dlech:
Here is a more fleshed out view of the EV3-G buttons blocks:

![image](https://user-images.githubusercontent.com/963645/30870787-6b2cc56c-a2aa-11e7-963c-25fb31fd971d.png)

![image](https://user-images.githubusercontent.com/963645/30870801-76beeb9e-a2aa-11e7-8f8f-5b5c8d4b0a0e.png)

![image](https://user-images.githubusercontent.com/963645/30870836-91f28466-a2aa-11e7-940a-8dd17e548fbd.png)

![image](https://user-images.githubusercontent.com/963645/30870853-a0fae494-a2aa-11e7-9008-c8436b413065.png)

![image](https://user-images.githubusercontent.com/963645/30870877-b1a907c6-a2aa-11e7-99e7-2314e0506326.png)

dwalton76:
@dlech for EVB what should its _buttons be now?  This is what we have for EV3
```
    _buttons = { 
            'up': {'name': '/dev/input/by-path/platform-gpio_keys-event', 'value': 103},
            'down': {'name': '/dev/input/by-path/platform-gpio_keys-event', 'value': 108},
            'left': {'name': '/dev/input/by-path/platform-gpio_keys-event', 'value': 105},
            'right': {'name': '/dev/input/by-path/platform-gpio_keys-event', 'value': 106},
[...]

PR #386, opened 2017-09-29
EV3-G API for Button

dwalton76:


dwalton76:
For issue #354

dlech:
Might I suggest using python-evdev to simplify things a bit here?

http://python-evdev.readthedocs.io/en/latest/

dwalton76:
That looks handy for Buttons but what about Sensors?  I could just have any `_wait()` for a Sensor class drop into a `while True` loop and constantly check to see if anything changed and then someday if there is POLL support for Sensor clean it up.

dlech:
For sensors, the only thing you can do for now is busy wait as you suggested.

debian/control
```diff
@@ -3,7 +3,7 @@ Maintainer: Ralph Hempel <rhempel@hempeldesigngroup.com>
 Section: python
 Priority: optional
 Standards-Version: 3.9.5
-Build-Depends: python3-setuptools (>= 0.6b3), python3-all (>= 3.4), debhelper (>= 9), dh-python
+Build-Depends: python3-setuptools (>= 0.6b3), python3-all (>= 3.4), debhelper (>= 9), dh-python, python-evdev
```
dlech: This is build deps. You don't need to add anything here.

Debian automatically figures out python package dependencies, so you don't need to add it in the binary package either.

(And this is the python2 version of the package)

ev3dev/core.py
```diff
@@ -137,6 +138,22 @@ def matches(attribute, pattern):
             if all([matches(path + '/' + k, kwargs[k]) for k in kwargs]):
                 yield f
 
+
+def file_open(path):
```
dlech: Private?

debian/control
dwalton76: reverted

ev3dev/core.py
dwalton76: Actually, I can undo this change, it is no longer needed.

dwalton76:
How do I update the CI test to install python-evdev?

dlech:
Like this?

https://github.com/rhempel/ev3dev-lang-python/blob/develop/.travis.yml#L6

dlech:
Or you can change `sudo` to `true` and add an install line for `sudo apt-get install --yes python3-evdev`

(probably needs `sudo apt-get update` first)

.travis.yml
```diff
@@ -1,9 +1,10 @@
 language: python
 python:
 - 3.4
-sudo: false
+sudo: true
```
WasabiFan: I'd prefer not to do this... a) we already use pip for the other dependency and b) this switches the builds from container-based to full VMs which they claim are slower.

dwalton76:
It looks like it is installing but fails to import in api_tests.py....will dig more tomorrow

ddemidov:
Sorry for being late to the train, but is there `python-evdev` for micropython?

dwalton76:
I can't find where there is micropython support for python-evdev.  For micropython we could always drop into a loop instead of being event driven...not great but would be a work-around.

dlech:
No, micropython does not support anything that is not in the standard library and only a fraction of the standard library. We can easily implement something very similar to the parts of evdev that we need in micropython, but not in a way that is compatible with regular python (e.g. you need to use uctypes which is not api compatible with ctypes).

dwalton76:
I need to get this in to finish the work on #388 and I think this one is in pretty good shape (no comments recently) so I am going to go ahead and merge it.

Issue #380, opened 2017-09-26
InfraredSensor - add support for on_red_up, on_red_down, etc bindings per channel


dwalton76:
This came up in the discussion PR #374 . Today you have to create an InfraredSensor object per channel to accomplish this.

PR #388, opened 2017-09-30
InfraredSensor on_red_up, etc support for multiple channels

dwalton76:
For issue #380 

@ddemidov look good for EVERSTORM?

### DIFF
--- a/ev3dev/core.py
+++ b/ev3dev/core.py
@@ -2789,49 +2789,49 @@ class InfraredSensor(Sensor, ButtonBase):
     # pressed.
     _BUTTON_VALUES = {
             0: [],
-            1: ['red_up'],
-            2: ['red_down'],
-            3: ['blue_up'],
-            4: ['blue_down'],
-            5: ['red_up', 'blue_up'],
-            6: ['red_up', 'blue_down'],
-            7: ['red_down', 'blue_up'],
-            8: ['red_down', 'blue_down'],
+            1: ['top_left'],
+            2: ['bottom_left'],
+            3: ['top_right'],
+            4: ['bottom_right'],
+            5: ['top_left', 'top_right'],
+            6: ['top_left', 'bottom_right'],
+            7: ['bottom_left', 'top_right'],
+            8: ['bottom_left', 'bottom_right'],
             9: ['beacon'],
-            10: ['red_up', 'red_down'],
-            11: ['blue_up', 'blue_down']
+            10: ['top_left', 'bottom_left'],
+            11: ['top_right', 'bottom_right']
             }
 
-    _BUTTONS = ('red_up', 'red_down', 'blue_up', 'blue_down', 'beacon')
+    _BUTTONS = ('top_left', 'bottom_left', 'top_right', 'bottom_right', 'beacon')
 
     # See process() for an explanation on how to use these
     #: Handles ``Red Up``, etc events on channel 1
-    on_red_up1 = None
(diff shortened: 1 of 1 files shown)


ev3dev/core.py
```diff
@@ -2728,20 +2730,36 @@ class InfraredSensor(Sensor, ButtonBase):
...
-    #: Handles ``Blue Down`` events.
-    on_blue_down = None
-
-    #: Handles ``Beacon`` events.
-    on_beacon = None
+    _BUTTONS = ('red_up', 'red_down', 'blue_up', 'blue_down', 'beacon')
+
+    # See process() for an explanation on how to use these
+    #: Handles ``Red Up``, etc events on channel 1
+    on_red_up1 = None
```
ddemidov: Would `channel1_red_up` (`C1_red_up`, `ch1_red_up`) be easier to understand? Also, would now be a good time to follow up on https://github.com/rhempel/ev3dev-lang-python/pull/374#discussion_r140795846 and change `red_up` to `top_red` or `top_left`?

dwalton76: @dlech wouldn't you also need to rename the buttons in the kernel too?

Regarding the on `on_red_up1` I think maybe we look at making these normal methods for the InfraredSensor class and then in the documents show how to make a child class where you override those methods.  To me that would be far easier to follow, the current approach is pretty unique and isn't something most folks have seen.  I like the `channel1_red_up` model for naming.

dlech: > wouldn't you also need to rename the buttons in the kernel too?

If these names are in the kernel, it is only in the documentation. What you call them in python doesn't depend on what the kernel calls them.

dwalton76: ah, I see now. 

changing `red_up` to either `top_left` or `top_red` sounds fine, of the two `top_left` would be my preference.

WasabiFan: `red_top_pressed`?

dwalton76: Having `pressed` in the name would be confusing though because the button may not be pressed. In `red_up` that doesn't mean the button is up, it is referring to the upper button on the red side.

Let's do this, I'll commit a patch with `top_left`, `channel1_top_left`, etc and we can see what it looks like and then go from there.

ev3dev/core.py
```diff
@@ -2787,33 +2789,49 @@ class InfraredSensor(Sensor, ButtonBase):
...
-    #: Handles ``Blue Down`` events.
-    on_blue_down = None
-
-    #: Handles ``Beacon`` events.
-    on_beacon = None
+    _BUTTONS = ('top_left', 'bottom_left', 'top_right', 'bottom_right', 'beacon')
+
+    # See process() for an explanation on how to use these
+    #: Handles ``Red Up``, etc events on channel 1
+    channel1_top_left = None
```
ddemidov: I just noticed that this drops `on_` prefix from the handlers. I think its ok given that the names differ from `top_left`. `top_right`, ... properties, and are already quite verbose with `channel*_` prefixes. Just wanted to make sure this is intentional. It could be confusing for users (e.g. `channel1_top_left` could be easily misunderstood to be a boolean property for top left button on channel 1).

dwalton76: I'll put the `on_` back that way the format is consistent with the `on_left`, `on_right`, etc for the buttons

dwalton76:
I think we are good on this one...merging

Issue #379, opened 2017-09-26
Deprecate helper.py


dwalton76:
- All of the motor functionality that I put in there now exists in core.py (it didn't at the time that I wrote it). I need to update MINDCUB3R to not use the helper.py and then all of the motor code in there can go.
- The wheel classes could be moved into a wheel.py, BALANC3R and GYROBOY demos will need to be updated.
- We need to find a home for RemoteControlledTank
- ColorSensorMixin only added the rgb() call and there weren't any objections to adding that method to core.py ColorSensor class

Issue #375, opened 2017-09-25
Split core.py into individual files


WasabiFan:
As discussed in ev3dev/ev3dev#937, it would be helpful for many reasons to split each class or area into its own file.

The primary difficulty here is that we want to wait until all outstanding work is merged so that we don't accidentally lose anything. Naturally, git will treat all of these changes as deletions and creations rather than a clean delta.

dwalton76:
Once the Infrared PR is in I will not have any more EV3-G APIs in progress so I could pause on that front so we can split up core.py.

dwalton76:
We should go ahead and figure out what we want the `ev3dev` directory structure to be.  Today we have
```
dwalton@laptop ~/l/ev3dev-lang-python> ls -l ev3dev/
total 212
-rw-rw-r-- 1 dwalton dwalton   1033 Sep 22 15:21 auto.py
-rw-rw-r-- 1 dwalton dwalton   2713 Sep 22 15:23 brickpi.py
-rw-rw-r-- 1 dwalton dwalton 135150 Sep 25 10:32 core.py
-rw-rw-r-- 1 dwalton dwalton   6118 Sep 22 15:23 ev3.py
drwxrwxr-x 2 dwalton dwalton  12288 Sep 20 15:08 fonts/
-rw-rw-r-- 1 dwalton dwalton  18404 Sep 22 15:23 GyroBalancer.py
-rw-rw-r-- 1 dwalton dwalton   6637 Sep 25 10:28 helper.py
-rw-rw-r-- 1 dwalton dwalton      0 Sep 22 15:21 __init__.py
-rw-rw-r-- 1 dwalton dwalton  22396 Sep 22 15:23 webserver.py
dwalton@laptop ~/l/ev3dev-lang-python> 
```
and we are about to add evb.py and pistorms.py.  Brainstorming....

## Platforms
Maybe it is time for a `platforms` sub-dir to hold
- brickpi.py
- ev3.py
- evb.py
- pistorms.py

Moving ev3.py would make me a little nervous because so many people import it instead of auto.py.  But if we were ever going to move it this would be the time.

## Sensors
Have a `sensors` sub-dir? `class Sensor` could go in its __init__.py and then all of the other sensor classes could go in individual files.  We could move all of the code that is in GyroBalancer.py to a gryo.py that holds `class Gyro`

## Motors 
[...]

PR #391, opened 2017-10-03
File musical chairs

dwalton76:
For issue #375 

@dlech @ddemidov @WasabiFan @ensonic take a look, this is a first stab at getting everything split up and moved around.  TouchSensor, Motors, ColorSensor, Buttons, InfraredSensor, and LEDs are all working.


dwalton76:
Ignoring the fonts directory, this is the current structure
```
robot@evb[ev3dev]# tree .
.
├── _button.py
├── button.py
├── control
│   ├── GyroBalancer.py
│   ├── __init__.py
│   ├── rc_tank.py
│   └── webserver.py
├── display.py
├── fonts
│   ├── charB08.pbm


│   ├── timR18.pil
│   ├── timR24.pbm
│   └── timR24.pil
├── helper.py
├── __init__.py
├── _led.py
├── led.py
├── motor.py
├── _platform
│   ├── brickpi.py
│   ├── ev3.py
│   ├── evb.py
│   ├── __init__.py
│   └── pistorms.py
├── port.py
├── sensor
│   ├── __init__.py
│   └── lego.py
├── sound.py
└── version.py

4 directories, 423 files
robot@evb[ev3dev]# 
```

dwalton76:
Can we make travis have a `/sys/class/board-info/`? See CI failure...

WasabiFan:
Why are there two files for buttons?

dlech:
version could be put in the root `__init__.py`, so there is no need for a module for just that.

dlech:
fonts probably shouldn't be in the source code tree.

dlech:
Is there anything left in `helper.py` after moving stuff to `control`?

dlech:
It seems like it would be cleaner to do some import magic in `_platform/__init__.py` rather than having `button`/`_button` and `led`/`_led`

dwalton76:
- `version.py` is generated via `python3 setup.py install`
- `fonts` not sure what the scoop is there, I've never used them
- `helper.py` I'll go ahead and chop it, I just need to redo the MINDCUBER demo but at this point we need to redo all of the demos so no point holding off

dwalton76:
yeah button and led need some work that was just what I came up with to get the importing working.  These are both things where a platform specific class must live in `_platform` and that class must inherit from some parent class.  So button for instance
- we know we want the user to `from ev3dev.button import Button`
- The Button class will live in `_platform` but needs `ButtonEVIO`
- So where to put `ButtonEVIO`? It can't be in `button.py` else you get into a circular import situation between it and the `_platform` file.

dwalton76:
@ddemidov thoughts on `fonts`?

dlech:
Why would we have platform-specific button classes? As far as I can tell, the only thing that should be different is the path to the device node.

dwalton76:
There is no guarantee that all platforms will have the same buttons...it just happens that EV3 and EVB do.  And since pistorms doesn't have any it shouldn't really have a Button class.

dwalton76:
Wait I think pistorms has a `back` button but that is it and brickpi doesn't have any buttons.

dlech:
How will documentation work if the same class is not the same between platforms? It seems to me like it would be better to make them all the same and throw a `RuntimeError` if something is missing (and the documentation will say something like "raises RuntimeError on BrickPi" if a class or method is not supported on that platform).

P.S. PiStorms has a Go button. (And one can add buttons to BrickPi using extra gpios)

dwalton76:
I'll give you it would certainly clean up the importing. What about LEDs though they differ a good deal from EV3 vs pistorms?

ddemidov:
Out of interest, what does `time python3 -c 'import ev3dev.ev3'` (before the PR) and `time python3 -c 'import ev3dev.motor'` (after the PR) show?

ddemidov:
> thoughts on fonts?

I don't see why we could not leave them to be as is, except they are only usable with python3/PIL.

ddemidov:
> Can we make travis have a /sys/class/board-info/? See CI failure...

We could put a `board-info` file into [ddemidov/ev3dev-lang-fake-sys](https://github.com/ddemidov/ev3dev-lang-fake-sys) (btw, should we move that repo under ev3dev?). Then we need to make the [get_current_platform](https://github.com/rhempel/ev3dev-lang-python/blob/799920eef3d8f34528243a40ccbb9d234782e9f6/ev3dev/core.py#L65) function to respect [Device.DEVICE_ROOT_PATH](https://github.com/rhempel/ev3dev-lang-python/blob/799920eef3d8f34528243a40ccbb9d234782e9f6/ev3dev/core.py#L150) constant (or, better yet, move the constant somewhere else). This way the travis tests [will know](https://github.com/rhempel/ev3dev-lang-python/blob/799920eef3d8f34528243a40ccbb9d234782e9f6/tests/api_tests.py#L14) where to look for the file.

ddemidov:
See #225 for some background on fonts.

ddemidov:
@dlech, 

> fonts probably shouldn't be in the source code tree

Do you mean we should put them into a separate package?

ddemidov:
Re platform specific buttons: platform file could provide both path for the event file and the list of supported buttons. Since python is a dynamic language, we could use that list (during import) to create properties corresponding to the buttons inside `Button` class.

dwalton76:
## Before
```
robot@ev3dev[~]# time python3 -c 'import ev3dev.ev3'

real	0m14.938s
user	0m11.340s
sys	0m0.560s
robot@ev3dev[~]# 
```

## After
```
robot@ev3dev[~]# time python3 -c 'import ev3dev.motor' 

real	0m11.645s
user	0m9.090s
sys	0m0.510s
robot@ev3dev[~]# 
```

dwalton76:
Latest commit removes `_button.py` with the assumption that all platforms have the same buttons.  I'll test later on a pistorms and have it raise some NoButton exception if the user tries to use buttons on a device that doesn't have them.  Do you guys like the way this looks more?  If so I'll do the same for LEDs.

dwalton76:
> btw, should we move that repo under ev3dev?

yes

dwalton76:
> We could put a board-info file into ddemidov/ev3dev-lang-fake-sys

cool, @ddemidov can you add that? I haven't done much with ev3dev-lang-fake-sys

ddemidov:
What should be in there? Should `FAKE-SYS` work?

dwalton76:
Works for me

dwalton76:
Should we move 'PowerSupply' out of '__init__.py' and into a 'power.py'?

WasabiFan:
> Should we move 'PowerSupply' out of 'init.py' and into a 'power.py'?

I'd say "yes", because it's pretty rare that someone would use the power supply class. In my mind it's just clutter for most use cases. (In fact, I think we should try to minimize the classes in _init.py_ overall)

dlech:
> What about LEDs though they differ a good deal from EV3 vs pistorms?

How so? PiStorms just has one extra color. I wouldn't call that a big difference.

dlech:
> @dlech,
> 
>     fonts probably shouldn't be in the source code tree
> 
> Do you mean we should put them into a separate package?

I'm not sure what is the usual way for distributing data files with python packages, but is just seems odd to have data files mixed in with source code files in the source repository. Maybe better to save this discussion for later though.

ev3dev/_platform/ev3.py
```diff
@@ -39,6 +38,8 @@
 INPUT_3 = 'in3'
 INPUT_4 = 'in4'
 
+buttons_filename = '/dev/input/by-path/platform-gpio_keys-event'
```
dlech: All caps for constants?

dwalton76:
I was thinking EV3 had more LEDs than pistorms but now that I look at http://www.ev3dev.org/docs/platform-comparison/ I see that it doesn't.  I'll clean up LEDs to follow the same model as buttons.

ev3dev/_platform/ev3.py
dwalton76: k

ev3dev/button.py
```diff
@@ -28,14 +28,283 @@
 if sys.version_info < (3,4):
     raise SystemError('Must be using Python 3.4 or higher')
 
+import array
+import time
+import evdev
```
dlech: `evdev` is probably a linux-specific module like `fcntl`

dlech:
> Since python is a dynamic language, we could use that list (during import) to create properties corresponding to the buttons inside Button class.

(Warning, getting a bit off topic here)

:+1: for taking advantage of dynamic nature of python to adapt to platforms, but I think I would rather see a single function that works for all buttons rather than identical functions/properties for each button. For example, I have a USB keypad that I can use for BrickPi/PiStorms that has the essential 6 EV3 buttons plus a bunch more. If I wanted to take advantage of other buttons, say the 10 numeric digits, then I would have to write 10 functions, one to handle each button. Whereas, if the button key code is just passed as a parameter, then I just have to write one function with a bunch of `elif`s.

ev3dev/button.py
dwalton76: I was meaning to ask about the try/except around fcntl...do we have folks running `api_test.py` on windows?

dlech: Probably not, but any editor with code completion will load the modules.

dwalton76:
Working on LEDs now...

dwalton76:
LEDs are cleaned up, I tested them on EV3 and tried testing on pistorms but ran into https://github.com/ev3dev/ev3dev/issues/978

I think this is far enough along to merge and then we can tackle anything else via separate issues.  Will merge tomorrow unless folks want me to hold off.

dwalton76:
Also resolves #379

ddemidov:
>> We could put a board-info file into ddemidov/ev3dev-lang-fake-sys

> cool, @ddemidov can you add that? I haven't done much with ev3dev-lang-fake-sys

Done in https://github.com/ddemidov/ev3dev-lang-fake-sys/commit/61d8dedecd55614331aa0e6c1d6f180c4a4d949c

WasabiFan:
OK, I understand what's going on. Inversion is now only done [here](https://github.com/gregcowell/ev3dev-lang-python/blob/4bc425f90270bd1fbdfbbc2daaf8c1e88728056d/ev3dev/motor.py#L729).

PR #394, opened 2017-10-04
Fix steering issue

### LINKED ISSUE #393: MoveTank on_for_rotations method problem
- **ev3dev version:** 4.4.87-22-ev3dev-ev3
- **ev3dev-lang-python version:** develop branch

If I run the code below, the wheels/tracks do not rotate in opposite directions as expected:
```
tank = ev3.MoveTank('outB', 'outC')
tank.on_for_rotations(left_speed_pct=5, right_speed_pct=-5, rotations=1.35)
```
This also affects the MoveSteering on_for_rotations method since it uses the MoveTank equivalent. I will submit a proposed fix as a pull request.

gregcowell:
Fixes #393 (MoveTank on_for_rotations method problem).


WasabiFan:
Is this confirmed to fix the problem? The motor drivers will ignore the speed as per the note on [this page](http://www.ev3dev.org/docs/tutorials/tacho-motors/#run-to-rel-pos):

> Note: Using a negative value for position_sp will cause the motor to rotate in the opposite direction. The sign of speed_sp is ignored like in run-to-abs-pos command.

In other words, we need to invert `rotations` rather than `speed`.

In regards to the sample from https://github.com/rhempel/ev3dev-lang-python/issues/393, is that the correct calling convention? [I don't think we have individual parameters for the two sides' speeds for that function.](https://github.com/rhempel/ev3dev-lang-python/blob/799920eef3d8f34528243a40ccbb9d234782e9f6/ev3dev/core.py#L1944) I think the expected call would use the `steering` parameter instead (which _does_ exist), in which case the correct distances will be calculated with the sign maintained as needed.

gregcowell:
There appears to be individual parameters for the MoveTank on_for_rotations method but not the MoveSteering on_for_rotations method. I believe it is the correct calling convention because I read it directly from the develop branch code. 

Yes, I tested this on my own EV3 and it fixed the problem. 

I noticed the issue when I tried this code:
```
move_steer = ev3.MoveSteering('outB', 'outC')

# Move robot in a square path
for i in range(4):
    # Move robot forward for 4 seconds
    move_steer.on_for_seconds(steering=0, speed_pct=50, seconds=3)
    # Turn robot left 90 degrees (adjust rotations for your particular robot)
    move_steer.on_for_rotations(steering=-100, speed_pct=5, rotations=1.35)
```
Then I tried the code below to confirm the issue:
```
tank = ev3.MoveTank('outB', 'outC')

tank.on_for_rotations(left_speed_pct=5, right_speed_pct=-5, rotations=1.35)
```
After making the change proposed in my pull request, both scripts worked fine. That is, the robot turned 90 degrees as expected. Apologies if I have misinterpreted how it is meant to work.

WasabiFan (approved):
Ah, I see how this works now. I was certainly looking at the wrong motor pair class. This looks good to me! (although, I'm still a bit confused as to why the sign is handled as expected.)
 
**Note before merging**: @dwalton76 Do you want to hold off on this until #391 is merged? I expect this'd just mess up our "musical chairs" and we'll need to re-apply this patch manually after that one is merged.

WasabiFan:
Side-note: @dwalton76 Why do we [use the absolute run command with a manually-added relative position](https://github.com/rhempel/ev3dev-lang-python/blob/799920eef3d8f34528243a40ccbb9d234782e9f6/ev3dev/core.py#L1002) instead of just using the relative run mode? Did that make implementation of something else easier?

dwalton76:
#391 is now in, @gregcowell can you resolve the conflicts, basically just make the same change but to motor.py instead

> Why do we use the absolute run command with a manually-added relative position instead of just using the relative run mode? Did that make implementation of something else easier?

We could have done it either way, I tend to work with absolute positions out of habit so did it that way.

dwalton76:
It took me a minute to wrap my head around how it was breaking, we were getting burned by a double negative (speed and rotations).  Fix looks good.

WasabiFan:
@dwalton yeah, the negative was introduced twice and as a result both motors have the same sign. But shouldn't the sign of the speed be ignored by the motor drivers? If we're now taking the absolute value of the rotation distance, where is the sign being reintroduced?

dwalton76:
This output is without the fix in place, this is for a `tank.on_for_rotations(-100, 20, 2)` call

`get_speed_steering()` shows the following, looks good
```
2017-10-04 12:50:36,154 DEBUG: MoveSteering: steering -100, LargeMotor(outA) speed -20, LargeMotor(outC) speed 20
```

`MoveTank.on_for_rotations()` has
```    
1538     def on_for_rotations(self, left_speed_pct, right_speed_pct, rotations, brake=True, block=True):
1539         """
1540         Rotate the motor at 'left_speed & right_speed' for 'rotations'
1541         """
1542         assert rotations >= 0, "rotations is %s, must be >= 0" % rotations
1543         self._validate_speed_pct(left_speed_pct, right_speed_pct)
1544         left_speed = int((left_speed_pct * self.max_speed) / 100)
1545         right_speed = int((right_speed_pct * self.max_speed) / 100)
1546 
1547         if left_speed > right_speed:
1548             left_rotations = rotations
1549             right_rotations = float(right_speed / left_speed) * rotations
1550         else:
1551             left_rotations = float(left_speed / right_speed) * rotations
1552             right_rotations = rotations
1553 
1554         log.debug("%s: left_speed_pct %s, right_speed_pct %s, rotations %s" % (self, left_speed_pct, right_speed_pct, rotations))
1555         log.debug("%s: %s speed %s rotations %s, %s speed %s rotations %s" % (self, self.left_motor, left_speed, left_rotations, self.right_motor, right_speed, right_rotations))
```

and debugs show
```
2017-10-04 12:50:36,158 DEBUG: MoveSteering: left_speed_pct -20, right_speed_pct 20, rotations 2
2017-10-04 12:50:36,161 DEBUG: MoveSteering: LargeMotor(outA) speed -210 rotations -2.0, LargeMotor(outC) speed 210 rotations 2
```

so we determine left_speed and right_speed one of which is negative so then when we calculate left_rotations and right_rotations one of those will also end up negative and we are in the double negative scenario.

dwalton76:
> But shouldn't the sign of the speed be ignored by the motor drivers

na they will treat a negative speed as rotate in reverse

dwalton76:
BTW we really need to enable logging, it makes troubleshooting so much easier

ddemidov:
> BTW we really need to enable logging, it makes troubleshooting so much easier

I don't mind if whether to log or not is configurable (and off by default). Otherwise logs tend to eat up disk space.

dwalton76:
If the user program importing the libraries doesn't enable logging then the log calls don't use any disk space.  If the user program does enable logging then whether or not it uses disk space depends on where the user tells python to write the logs, default is to print them to STDOUT.

WasabiFan:
> na they will treat a negative speed as rotate in reverse

This is contrary to the info I quoted above from the motor docs. @dlech has this behavior changed in Stretch?

dwalton76:
Hmm yeah I see what you mean about the docs.  It certainly seems to honor positive vs negative speed in terms of which direction the motor rotates.

dlech:
> This is contrary to the info I quoted above from the motor docs. @dlech has this behavior changed in Stretch?

Currently the `run-to-*-pos` commands should ignore the sign of the `speed_sp` - although I am considering changing this in stretch (just had a similar conversation in ev3go a week ago).

dwalton76:
>  It certainly seems to honor positive vs negative speed in terms of which direction the motor rotates.

I was wrong there, more debugging shows the negative sign from the speed was used to create a negative position_sp for one motor and positive for the other
```
robot@evb[ev3dev-lang-python]# ~/bin/tank.py 
2017-10-04 21:32:13,515 DEBUG motor.py: 1695 -        get_speed_steering(): MoveSteering: steering -100, LargeMotor(outA) speed -20, LargeMotor(outC) speed 20
2017-10-04 21:32:13,518 DEBUG motor.py: 1570 -          on_for_rotations(): MoveSteering: left_speed_pct -20, right_speed_pct 20, rotations 2
2017-10-04 21:32:13,522 DEBUG motor.py: 1571 -          on_for_rotations(): MoveSteering: LargeMotor(outA) speed -210 rotations 2.0, LargeMotor(outC) speed 210 rotations 2
2017-10-04 21:32:13,527 DEBUG motor.py:  750 -   _set_position_rotations(): LargeMotor(outA): current position 0, rotations 2.0, count-per-rot 360, final position -720
2017-10-04 21:32:13,534 DEBUG motor.py:  750 -   _set_position_rotations(): LargeMotor(outC): current position 0, rotations 2, count-per-rot 360, final position 720
robot@evb[ev3dev-lang-python]# 
```

WasabiFan:
OK, I understand what's going on. Inversion is now only done [here](https://github.com/gregcowell/ev3dev-lang-python/blob/4bc425f90270bd1fbdfbbc2daaf8c1e88728056d/ev3dev/motor.py#L729).

Issue #390, opened 2017-10-03
motors POLLPRI: need to read mode file at least once


dwalton76:
> Another thing is that if you have never read a file, poll won't block. So you have to read at least once before calling poll if you want to see it block.

dwalton76:
Won't fix...up to the user to make sure they read it one time at the start (very likely)

PR #395, opened 2017-10-04
Misc

dwalton76:
No squashing when we merge this PR, will leave these as individual commits.  The commit that adds the asserts in the motor class will catch issues like #393

### DIFF
--- a/ev3dev/motor.py
+++ b/ev3dev/motor.py
@@ -31,7 +31,7 @@ if sys.version_info < (3,4):
 import select
 import time
 from os.path import abspath
-from ev3dev import get_current_platform, Device
+from ev3dev import get_current_platform, Device, list_device_names
 
 # The number of milliseconds we wait for the state of a motor to
 # update to 'running' in the "on_for_XYZ" methods of the Motor class
--- a/ev3dev/motor.py
+++ b/ev3dev/motor.py
@@ -56,6 +56,9 @@ elif platform == 'brickpi':
 elif platform == 'brickpi3':
     from ev3dev._platform.brickpi3 import OUTPUT_A, OUTPUT_B, OUTPUT_C, OUTPUT_D
 
+elif platform == 'fake':
+    from ev3dev._platform.fake import OUTPUT_A, OUTPUT_B, OUTPUT_C, OUTPUT_D
+
 else:
     raise Exception("Unsupported platform '%s'" % platform)
 
--- a/ev3dev/motor.py
+++ b/ev3dev/motor.py
@@ -1549,9 +1549,9 @@ class MoveTank(MotorSet):
 
         if left_speed > right_speed:
             left_rotations = rotations
-            right_rotations = float(right_speed / left_speed) * rotations
+            right_rotations = abs(float(right_speed / left_speed)) * rotations
         else:
-            left_rotations = float(left_speed / right_speed) * rotations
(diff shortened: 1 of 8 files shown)


dwalton76:
The fix for #390 is also in here

WasabiFan:
Can you add the issue numbers to the individual commit messages (or, even better, `fixes` annotations)?

dwalton76:
Hmmm CI is not happy about fake-sys....will dig some this evening after work

ev3dev/motor.py
```diff
@@ -112,36 +219,10 @@ def __init__(self, address=None, name_pattern=SYSTEM_DEVICE_NAME_CONVENTION, nam
...
-    '_state',
-    '_stop_action',
-    '_stop_actions',
-    '_time_sp',
-    '_poll',
-    ]
+        # We use poll() to monitor the state of the motor so we can react when
+        # the state changes but if you have never read the file, poll() won't
+        # block. Read 'state' once so that poll() will block.
+        self.state
```
dlech: This is only necessary if you are not calling poll in a loop, otherwise the loop does exactly the same thing.

ddemidov:
> Hmmm CI is not happy about fake-sys....will dig some this evening after work

For this to work `board_info_dir` needs to be configurable, and not a hard-coded path here:
https://github.com/rhempel/ev3dev-lang-python/blob/aab7a1cd6b21851ed3aa46e4f1ce00feed22d717/ev3dev/__init__.py#L43-L51

Similar to what is done inside `Device`:
https://github.com/rhempel/ev3dev-lang-python/blob/aab7a1cd6b21851ed3aa46e4f1ce00feed22d717/ev3dev/__init__.py#L123-L128

Then tests may change the root to point to fake-sys as it is done here:
https://github.com/rhempel/ev3dev-lang-python/blob/aab7a1cd6b21851ed3aa46e4f1ce00feed22d717/tests/api_tests.py#L16

ev3dev/motor.py
dwalton76: Seems worth keeping though, it is only one read and if the user does their own thing with POLLPRI they won't get tripped up on this.

dwalton76:
@ddemidov any idea what is causing the CI
```
The command "eval git submodule update --init --recursive " failed. Retrying, 2 of 3.
```
failure?

ev3dev/motor.py
dlech: The user will still have to read once themselves so that they know the initial state

dlech: Also, I just realized that polling the state attribute will only work on EV3 (and EVB since they share hardware drivers), but it won't work on any other platform (BrickPi, BrickPi3, PiStorms).

WasabiFan:
@dwalton76 I think the submodule in this repo currently points to a fork of `fake-sys` instead of the original. Not sure why.

ev3dev/motor.py
dwalton76: will back this part out

Any ideas on how to get poll() working on BrickPi, etc?

dwalton76:
@WasabiFan I reverted my fake-sys change and will tackle it via a different PR.  Our fake-sys is pointing at https://github.com/rhempel/ev3dev-lang-fake-sys/tree/master which is a fork of https://github.com/ddemidov/ev3dev-lang-fake-sys, we want Denis's.

Issue #358, opened 2017-09-19
EV3-G API: Sound


dwalton76:
![sound-01](https://user-images.githubusercontent.com/468309/30606561-16a38a0e-9d40-11e7-9803-830261c4714b.png)

dwalton76:
Per this video
https://www.youtube.com/watch?v=D2cGnTxsdy4

- play_type 0 is wait for completion
- play_type 1 is "play once"...I assume do not wait for completion
- play_type 2 is "play on a loop"

dwalton76:
```
class Sound:
    """  
    Sound-related functions. The class has only static methods and is not
    intended for instantiation. It can beep, play wav files, or convert text to
    speech.
```

Unless anyone objects, I am going to make the Sound class follow the model of all of our other classes where it is expected that you create an instance of that class.

ddemidov:
I think you can already create an instance of the Sound class and used it that way.

dwalton76:
`self` isn't expected in any of the methods though...I don't think it would work.  Either way will clean it up and make it consistent.

ddemidov:
That's because the methods are static. The class is stateless, so self is not necessary. But the following works for me:
```py
>>> from ev3dev.core import Sound
>>> s = Sound()
>>> s.beep()
```

dwalton76:
The only musical instrument I can play is the radio.  My musical ignorance is making `play_note()` challenging to implement. @EricPobot you wrote a lot of the code in this area earlier this year, would you be willing to implement that part of the EV3-G API for the `Sound` class?

EricPobot:
@dwalton76 I'd be glad to help, but I'm not sure to see clearly what is expected to be done. I understand that it has some relation with an EV3-G connection but nothing more.  I confess having been a bit away from ev3dev these last months, and thus have not followed its latest evolutions. Could you give me some input please ?

WasabiFan:
Related: https://github.com/ev3dev/ev3dev/issues/970

EricPobot:
@WasabiFan I'm a bit confused. The related topic is about gathering sounds and graphics to be used with ev3dev. Not about software.

dlech:
I think the question is how do you convert a musical note to a frequency. Here is a table from the official LEGO firmware.

```python
_NOTES = {
  'C4': 262,
  'D4': 294,
  'E4': 330,
  'F4': 349,
  'G4': 392,
  'A4': 440,
  'B4': 494,
  'C5': 523,
[...]

PR #396, opened 2017-10-04
EV3-G API Sound

dwalton76:


dwalton76:
For issue #358

ev3dev/sound.py
```diff
@@ -44,11 +44,9 @@ def _make_scales(notes):
     return res
 
 
-class Sound:
+class Sound(object):
```
WasabiFan, line 47: Why do we want to do this? Can't you already instantiate the class if desired?

dwalton76: For consistency with all of our other classes.

ev3dev/sound.py
```diff
@@ -75,8 +73,21 @@ class Sound:
...
+    # play_types
+    PLAY_WAIT_FOR_COMPLETE = 0
+    PLAY_NO_WAIT_FOR_COMPLETE = 1
+    PLAY_LOOP = 2
+
+    PLAY_TYPES = (
+        PLAY_WAIT_FOR_COMPLETE,
+        PLAY_NO_WAIT_FOR_COMPLETE,
+        PLAY_LOOP
+    )
```
ddemidov, line 85: @WasabiFan  this may answer your question above: @dwalton introduces these parameters, so the `Sound` class is now stateful. But, @dwalton, why are these capitalized? These should be user settings, not constants, right?

ddemidov: Ah, sorry. These are not parameters, these are types of operation.

Issue #235, opened 2016-10-23
Odometry library


dwalton76:
This might be fun/useful to port to Python 

https://github.com/ilovenetruinos/ev3g-odometry/tree/de917e2c8c96936ef03a8bbbdf3e13e930bfca4b

WasabiFan:
Yeah, that seems interesting; and it could definitely be useful for beginners (it is, after all, intended for that audience). I find it funny how it was inspired by ev3dev projects, then implemented in EV3-G, and now we are going to translate it to ev3dev code again :laughing:

dwalton76:
It has been almost 2 years and I haven't even started on this one.  I am going to close it and we can re-open if someone wants to work on it.

dwalton76:
Now that MoveDifferential and DistanceValue classes are in, work can begin on odometry. @WasabiFan @ddemidov thoughts on what the API for this should look like?  I think most changes we make will need to be in the MoveDifferential class as that is the only place we know the wheel dimensions and spacing needed to do the calculations.

It could be as simple as passing a ``use_gyro_for_accuracy`` bool to
- on_for_distance
- turn_right
- turn_left
- on_arc_right/on_arc_left I need to give this one some thought

Is there anything we want to do for the case where a gyro is not available?  ``MoveDifferential`` is already providing some basic odometry.  I can update it to use the new DistanceValue classes but beyond that I can't think of anything else to add for a non-gyro scenario.

dwalton76:
One big use case for this could be First Lego League.

WasabiFan:
I think that makes sense. The main things you want to do which are odometry-backed are going straight and doing a circular turn.

dwalton76:
Doing some research on this...the ev3dev project that was the source of inspiration for the EV3-G implementation was https://www.ev3dev.org/projects/2016/08/07/Mapping/

This chunk of code reads the motor positions periodically and sends those (plus the timestamp) to some other program via a UDP socket
https://github.com/bmegli/ev3dev-mapping-modules/blob/master/ev3odometry/main.cpp

Not sure yet what code it is sending that data to though

dwalton76:
Great source here:
http://seattlerobotics.org/encoder/200610/Article3/IMU%20Odometry,%20by%20David%20Anderson.htm

WasabiFan:
[...]

PR #398, opened 2017-10-05
Added MoveJoystick

dwalton76:


dwalton76:
I'm digging MoveTank...makes it easy to make child classes like this one.

I tested this with the mobile web-interface.  I have a PS3 controller and plan on doing a `ev3dev/control/ps3_tank.py` but am not there yet.

dwalton76:
Any reservations on this one?  It is almost all code we had before just cleaned up to be a child class of `MoveTank`

ev3dev/motor.py
```diff
@@ -1695,3 +1699,199 @@ def on_for_seconds(self, steering, speed_pct, seconds, brake=True, block=True):
...
         (left_speed_pct, right_speed_pct) = self.get_speed_steering(steering, speed_pct)
         MoveTank.on(self, left_speed_pct, right_speed_pct)
+
+
+class MoveJoystick(MoveTank):
+    """
+    Used to control a pair of motors via a joystick
+    """
+
+    def angle_to_speed_percentage(self, angle):
```
WasabiFan, line 1702: I think we should be referring to this as "arcade drive".

ev3dev/motor.py
```diff
@@ -1695,3 +1699,199 @@ def on_for_seconds(self, steering, speed_pct, seconds, brake=True, block=True):
...
+        for example. We look at how the motors transition from 0 degrees to 45 degrees:
+        - the left motor is 1 so that is easy
+        - the right motor moves from -1 to 0
+
+        We determine how far we are between 0 and 45 degrees (11.25 is 25% of 45) so we
+        know that the right motor should be 25% of the way from -1 to 0...so -0.75 is the
+        percentage for the right motor at 11.25 degrees.
+        """
+
+        if angle >= 0 and angle <= 45:
```
WasabiFan: Is there really no less verbose way of doing this?

dwalton76: Not that I can think of off the top of my head.

ev3dev/motor.py
dwalton76: But it takes an angle and returns speed percentages for the two motors...the current name seems much more descriptive to me.

ev3dev/motor.py
ddemidov: Looks like `0 <= angle <= 45` should work:
https://stackoverflow.com/questions/13628791/how-do-i-check-whether-an-int-is-between-the-two-numbers

dwalton76: oh I thought wasabi was asking if the entire function could be made less verbose but if it is just the angle comparisons yep I can clean that up to a single operation for each if/elif

ev3dev/motor.py
WasabiFan: I'm proposing that you rename the class and any other references to the concept. There are many ways to drive some motors with joysticks, so being specific to make it clear what algorithm is being used would be helpful.

dwalton76: How about `MoveJoystickArcade`?

ev3dev/motor.py
ddemidov: May be you are right. I tried to plot the left and right motor powers from your comment against the angle to see if there is a simple mathematical expression lurking there, but it makes for some weirdly interesting plots.

ev3dev/motor.py
ddemidov: Where does 'Arcade' come from? Is it a common term for this? Frankly, neither of the names make sense to me and I would have to read the comments anyway to understand the purpose of this function.

ev3dev/motor.py
dwalton76: I remember when I was working on this I would take the joystick and move it around the outer edge of the circle.  Initially I got some really crazy movements but then figured out that there are 8 sections of the circle to deal with thus the big if/elif tree. How the motor speeds adjust as you move from one section to the next varies a lot depending on which two sections you are dealing with.

ev3dev/motor.py
dwalton76: It is just because a lot of arcade machines have joysticks that you use to control vehicles, characters, etc.  What if we included the name Tank in the class....`MoveTankJoystick`?  Are there other ways one might use a joystick to move a tank?

ev3dev/motor.py
ddemidov: The plots: http://nbviewer.ipython.org/745ed4980597830194766f8e865dba69

ddemidov: Here is the plot for old `steering` function.
http://nbviewer.jupyter.org/gist/anonymous/d2750f5923b3523f8c5e25812adb7a20
It does look the same as the top part of your circle, which I think is a good sign. Not sure though that the bottom part of your circle is correct. Should not there be some kind of symmetry?

ddemidov: Nah, I think your function is correct.

dwalton76: Cool plots

ddemidov: Does this make more sense?
http://nbviewer.jupyter.org/gist/anonymous/bc23dcefc6da4f0008df620d33e4d171

EDIT: ignore the comments in the steering function definition, it should do the same as your function.

dwalton76: I would have to build a tank robot and test it to say for sure but my gut says no just because it looks so different from what I have and I know that what I have works (and took me a while to get right).  It would be worth testing though.

ddemidov: The keypoints have same values as on your circular plot, and I like the last version better because there is symmetry between top and bottom halves.

dwalton76: I'll test it out sometime in the next week....I just built MINDCUBER yesterday so all of my large motors are tied up right now.  Anyone know of a tank base that uses medium motors?

ev3dev/motor.py
dwalton76: I cleaned up the angle compares, haven't changed the name of the class yet but I am game for changing it, we just need to pick a name.

dwalton76:
@ddemidov I built a quick tank with medium motors and tried
http://nbviewer.jupyter.org/gist/anonymous/bc23dcefc6da4f0008df620d33e4d171

When moving the joystick up it works but when I pull the joystick down the left vs. right is backwards.  For example if I pull the joystick down and to the left the robot should back up while make a left turn but instead it is making a right turn (it does move backwards though).

ddemidov:
After some thought, what you observe makes sense. A thought experiment:

* I start with joystick in full forward, and slowly move it to the right. I end up with the tank rotating clockwise.
* Now I start with joystick in full backwards, and slowly move it to the right again. I end up with the tank rotating counterclockwise.

Which means there is a discontinuity at the left and right positions of the joystick, and the function should look like this: 
http://nbviewer.jupyter.org/gist/anonymous/d0d656ee5748ab3d2147634bbfb4fa5b.

dwalton76:
The gotcha there is right when you go past 180 both the left and right motors will suddenly switch directions and you get a very jerky movement

dlech:
Why not use MoveSteering instead of MoveTank? Then just feed the X-axis to steering and the Y-Axis to speed.

WasabiFan:
> Why not use MoveSteering instead of MoveTank? Then just feed the X-axis to steering and the Y-Axis to speed.

Then moving the joystick to `(1, 0)` wouldn't apply any power, whereas the intention with the current code is that it spins at full speed.

dwalton76:
The vector_length
```
vector_length = hypot(x, y)
```
could be used to determine the power to apply though.

Testing now...if this works this will be drastically cleaner.

dwalton76:
Well...it kinda works but the behavior is pretty different from what we have today and doesn't feel like what you would expect out of a joystick.   Today if you move the joystick 1/10 to (100, 0) the robot will rotate clockwise at 10% speed, if you move all the way to (100, 0) it rotates clockwise at 100% speed.  

With the MoveSteering approach if I move the joystick 1/10 to (100, 0) we no longer rotate clockwise instead we make a very gradual right turn.  If I move all the way to (100, 0) then we would have comparable behavior.

I think we should stick with what we have, the function might be a tiny bit long for what it is doing (but as we've proven with the brainstorming here replacing it isn't a quick task) but it does what we need.

ddemidov:
Ok, what we have is probably the best one can do with joystick and a simple tank drive.

Offtopic: it would be cool to introduce an accelerometer sensor to the tank, and change the behavior: when user pulls the joystick to some angle, the tank reorients itself to face the selected direction, and then drives forward. At least that's what I would expect from a video game :).

dwalton76:
> it would be cool to introduce an accelerometer sensor to the tank, and change the behavior: when user pulls the joystick to some angle, the tank reorients itself to face the selected direction, and then drives forward

Sounds like a combination of #235 and the controls in this robot http://www.bartneck.de/2017/08/27/holonomic-lego-mindstorms-robot/
:)

dwalton76:
Am going to merge this one, we can rename the class via another PR if we come up with a more descriptive name.

Issue #381, opened 2017-09-26
EV3-G API - breaking changes


dwalton76:
Use this issue to track breaking changes that need to be documented.

PR #374 removed these classes
- RemoteControl
- BeaconSeeker

dlech:
If we are going to make breaking changes can we do the following:

1. Release a 1.x version on ev3dev-stretch that includes 06ae05faf27fc5d33dc5d26742a11ed2bc82e937 - this should get OpenRoberta working on stretch - it is currently broken (someone just asked about this on IRC this morning).
2. Rename the root namespace and package names so that 1.x and 2.x can be installed in parallel (e.g. `ev3dev` > `ev3dev2`). This way, 2.x can be released so people can use it and OpenRoberta can keep working without constant maintenance while 2.x stabilizes.

ddemidov:
For reference: https://github.com/rhempel/ev3dev-lang-python/issues/339#issuecomment-327022503

ddemidov:
> Rename the root namespace and package names so that 1.x and 2.x can be installed in parallel (e.g. ev3dev > ev3dev2). 

I understand the package name change, but why the root namespace?

dlech:
It doesn't have to be the *root* namespace, but if both version are installed at the same time there has to be something different in the namespace, does there not? Otherwise, how would imports work?

dwalton76:
`auto_mode` for Sensors has been removed

dwalton76:
For `InfraredSensor` we changed the following names
- `red_up` is now `top_left`
- `red_down` is now `bottom_left`
- `blue_up` is now `top_right`
- `blue_down` is now `bottom_right`

and the `on_red_up`, etc handlers now support multiple channels and are named `on_channel1_top_left`, etc

dwalton76:
Importing core.py and auto.py will no longer work

ddemidov:
Which is better: rename stable package (and namespace) for OpenRoberta to `ev3dev-v1` or `ev3dev-jessie`, or rename future packages to `ev3dev-v2`?

ensonic:
I'd appreciate if we keep things stable for jessie. If we break things for stretch anyway I'll need to do some porting and depend on the right version. So please collect a list of things that are changed.

dwalton76:
Can we keep just one package name but with different versions?

WasabiFan:
@ensonic see above. This thread is devoted to tracking breaking changes to make sure we can adequately document it.

[...]

PR #366, opened 2017-09-20
Stretch fixes

dwalton76:

Issue #339, opened 2017-09-04
ev3dev-stretch


dlech:
Following up on https://github.com/rhempel/ev3dev-lang-python/issues/338#issuecomment-327018121

> > ev3dev-lang-python is already broken in ev3dev-stretch
>
> How so? That seems pretty important. Can you open a new issue for it?

https://github.com/ev3dev/ev3dev-kpkg/blob/ev3dev-stretch/ev3dev-ev3/changelog

> * Breaking change: LEDs are renamed to mainline kernel names
>     * Names will be similar to "led0:red:brick-status"

> * Breaking change: EV3/EVB input/output port drivers are merged
>     * EV3 port names will be similar to "ev3-ports:in1".

dlech:
Important info for package maintainer: 

in `debian/changelog`, `stable` in the line

    python-ev3dev (0.8.1) stable; urgency=medium

will upload to jessie.

To upload to stretch, use `testing`.

    python-ev3dev (0.8.1) testing; urgency=medium

dlech:
For all of the ev3dev packages I maintain, I have created an `ev3dev-jessie` branch and an `ev3dev-stretch` branch. I've basically adopted (something like) the Debian policy for stable and testing, which is only critical bug/security fixes go into stable (jessie in our case), no new features. You can do what you like, including new features and breaking changes.

WasabiFan:
> https://github.com/ev3dev/ev3dev-kpkg/blob/ev3dev-stretch/ev3dev-ev3/changelog

Oh, damn... when I originally read that I thought all the changes were on the EVB/BrickPi side. Looks like we'll need to fix those.

> For all of the ev3dev packages I maintain, I have created an ev3dev-jessie branch and an ev3dev-stretch branch.

I think we've gotten to the point where that convention would work well here. We'll need to figure out whether we want to maintain a "dev" branch or not.  List of things that will be impacted (not necessarily complete):
- CI builds (I think we have some configuration which points to branch names, or at least we _did_)
- Documentation versions and labels
- Contribution info (somewhere we added an explanation of `develop` vs. `master`)

dwalton76:
So branches and branch names...some options

option 1
- rename `master` to `ev3dev-lang-python-jessie`
- create a child branch called `ev3dev-lang-python-stretch`
- create a `ev3dev-lang-python-stretch-develop` child branch off of `ev3dev-lang-python-stretch`
- change the default branch to `ev3dev-lang-python-stretch`
[...]

PR #400, opened 2017-10-06
Stable stretch release

ddemidov:
Branched `stretch-stable` off of `jessie`, bumped it to 06ae05faf27fc5d33dc5d26742a11ed2bc82e937 and added a changelog entry.

See https://github.com/rhempel/ev3dev-lang-python/issues/381#issuecomment-332241734

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -19,9 +19,9 @@ python-ev3dev (1.0.0) stable; urgency=medium
   [Daniel Walton]
   * Add wait_until_not_moving for motors
 
- -- TODO
+ -- Denis Demidov <dennis.demidov@gmail.com>  Tue, 05 Sep 2017 16:50:20 +0300
 
- python-ev3dev (0.8.1) stable; urgency=medium
+python-ev3dev (0.8.1) stable; urgency=medium
 
   [Kaelin Laundry]
   * Documentation updates


WasabiFan:
We need to update the files in `debian/` to point to the new OS version. Also, don't we want to update the package name and module name here to ev3dev2 since this is the first Stretch release?

debian/changelog
```diff
@@ -1,3 +1,10 @@
+python-ev3dev (1.1.0) testing; urgency=medium
```
ddemidov: @WasabiFan, as I understood from https://github.com/rhempel/ev3dev-lang-python/issues/381#issuecomment-332241734, `testing` here is enough for the package to end up in the stretch repo.

ddemidov:
> Also, don't we want to update the package name and module name here to ev3dev2 since this is the first Stretch release?

This is intended as an OpenRoberta-compatible release, which means it is still v1 and we don't need to rename the package. v2 would be the future release.

debian/changelog
ddemidov: Wrong link. Should be https://github.com/rhempel/ev3dev-lang-python/issues/339#issuecomment-327022503

WasabiFan: I was thinking of `debian/compat`, but it looks like that's already pointing to Stretch...

WasabiFan:
Ok, that makes sense. Shouldn't we publish a Jessie release too to keep parity? Additionally, this branching feels wierd... Is your plan to cherry-pick the changelog commit to our develop branch later?

ddemidov:
> Shouldn't we publish a Jessie release too to keep parity?

Functionally, this is on par with our jessie release, so I don't see the need to update that.

> Additionally, this branching feels wierd... Is your plan to cherry-pick the changelog commit to our develop branch later?

This is actually far behind our develop branch, so I don't see how we could do that.

WasabiFan:
> Functionally, this is on par with our jessie release, so I don't see the need to update that

The current Jessie release lacks the motor pair classes, I believe. Same with the friendly errors.

> This is actually far behind our develop branch, so I don't see how we could do that.

That's the idea behind git cherry-pick: apply the patch from a commit or multiple commits to a separate commit history line (i.e., different parent, but same changes and author). I suppose the question I'm really asking is: was your plan to work off of this branch in the future? Or something else?

ddemidov:
> The current Jessie release lacks the motor pair classes, I believe. Same with the friendly errors.

Hmm, then I would just drop those commits from this branch as well. The idea is provide a version minimally compatible with stretch but otherwise identical to our last jessie release, to make @ensonic happy.

> That's the idea behind git cherry-pick: apply the patch from a commit or multiple commits to a separate commit history line (i.e., different parent, but same changes and author). I suppose the question I'm really asking is: was your plan to work off of this branch in the future? Or something else?

I understand how the cherry-pick works, but I would just start from scratch with the changelog when we are ready to release v2. Again, I consider this branch a dead-end only meant to keep OpenRoberta afloat while we are working on v2.

WasabiFan:
OK, that seems fine to me. I do believe that you'll want to drop some of the intermediate changes from this branch to keep parity with the current Jessie release.

ddemidov:
Kept #366, dropped the rest.

WasabiFan (approved):
Looks good to me. Not an expert on Debian packaging though.

dwalton76 (approved):

WasabiFan:
@ddemidov How'd releasing this go? Did you get it published or were there still some packaging issues?

ddemidov:
I am stuck with https://github.com/ev3dev/ev3dev-lang-python/issues/381#issuecomment-336144617.

dlech:
Following up to https://github.com/ev3dev/ev3dev-lang-python/issues/381#issuecomment-336144617

I have the following in `~/.pbuilderrc` (only the first line should be effective).

```
APTCACHEHARDLINK=no
#DEBBUILDOPTS="-sa"

#APTKEYRINGS="/usr/share/keyrings/ev3dev-archive-keyring.gpg"
```

Edit: I think the reason I have `APTCACHEHARDLINK=no` is because my `/home/` is on a separate file system (partition). It shouldn't make a difference otherwise.

dlech:
Here is how travis runs pbuilder-ev3dev: https://github.com/ev3dev/pbuilder-ev3dev/blob/master/.travis.yml#L15

Maybe there is a missing dependency in the package?

dlech:
Also, some other repositories use pbuilder-ev3dev in travis sucessfully: https://github.com/ev3dev/brickman/blob/ev3dev-stretch/.travis.yml

dlech:
Here is a successful log to compare: https://travis-ci.org/ev3dev/brickman/jobs/214739979#L1357

One thing I noticed is that `/etc/mailname` seems to be missing in your log. On my computer, this file contains the hostname.

After that is 

    E: gnupg, gnupg2 and gnupg1 do not seem to be installed, but one of them is required for this operation

 that we discussed already. Perhaps none of these commands are in `$PATH` for root/fakeroot?

dlech:
I have also updated http://www.ev3dev.org/docs/devtools/installing-the-ev3dev-archive/

There is a new PPA for getting ev3dev development tools. I'm not sure if it has a newer version of pbuilder-ev3dev or not.

dlech:
Missing gnupg was fixed in pbuilder-ev3dev v1.0.1: https://github.com/ev3dev/pbuilder-ev3dev/commit/e8c7da17d78b85f2515056bda1571d1b7d75bfa4

ddemidov:
It was an error on my side. I had an outdated version of `pbuilder-ev3dev` in my `~/bin` for some reason (I have probably set up my packaging image before `pbuilder-ev3dev` debian package was a thing). Removing that solved the issue with missing gnupg.

Sorry for the noise.

ddemidov:
Closed this by mistake. I've pushed the changes to `stretch-stable`, and uploaded the packages to the package repository.

Issue #399, opened 2017-10-05
Create an upgrading guide for Jessie to Stretch


WasabiFan:
We should look into writing a simple page that lists the changes that one needs to make to upgrade their scripts to our v2 release for Stretch. The most important change would be to imports (it's looking like we might rename the module to ev3dev2) but there are some sensor types which we also shifted around. There's an issue with a full list.

It would also be nice to outline the new features, specifically the new API surface for existing classes and the motor pairs.

WasabiFan:
With the `2.0.0-beta1` release, this is now a thing that exists. It's available here:

https://ev3dev-lang.readthedocs.io/projects/python-ev3dev/en/ev3dev-stretch/upgrading-to-stretch.html

If you don't have the V2 library installed already, you can do so yourself:

```
sudo apt-get update
sudo apt-get install python3-ev3dev2
```

PR #401, opened 2017-10-07
Enable using markdown for Sphinx docs

WasabiFan:
I'm going to admit it... reStructured Text is _driving me nuts_.

It turns out that there's an RTFD-supported extension that parses Markdown and enables it to be used with Sphinx by injecting the right RST-style metadata where necessary. This would certainly make writing docs easier for me :laughing:

I don't want to force a full translation task here, but I was thinking that if others are interested we could begin writing new pages in Markdown (e.g. #399). Thoughts? I figure this PR is a fine place to discuss, I'm not trying to force it.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -25,8 +25,10 @@ on_rtd = os.environ.get('READTHEDOCS', None) == 'True'
 if on_rtd:
     import pip
     pip.main(['install', 'sphinx_bootstrap_theme'])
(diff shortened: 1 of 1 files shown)


dwalton76 (approved):

dwalton76:
I haven't written many docs so if you say this makes things easier I trust you :)

ddemidov (approved):

PR #402, opened 2017-10-08
Remove "connected" attribute

### LINKED ISSUE #372: Rethink "connected" attribute
When we originally designed the ev3dev-lang libraries, we based them all on a concept where devices would expose a `connected` property to indicate whether they had successfully found a device matching the parameters from the constructor. This has led to a lot of code that looks like this:

```python
m = ev3.Motor(ev3.OUTPUT_A)
assert m.connected
```

...and lots of confused users as a result (especially when they don't know to include that assertion).

Do we want to change this behavior? I think it would make sense at this point to throw an exception from the constructor. This would be just for this library and wouldn't necessarily require a change to others.

WasabiFan:
Fixes #372

I haven't tested this on-device yet.

### DIFF
--- a/ev3dev/__init__.py
+++ b/ev3dev/__init__.py
@@ -124,13 +124,16 @@ def list_device_names(class_path, name_pattern, **kwargs):
                 yield f
 
 
+class DeviceNotFoundException(Exception):
+    pass
+
 # -----------------------------------------------------------------------------
 # Define the base class from which all other ev3dev classes are defined.
 
 class Device(object):
     """The ev3dev device base class"""
 
-    __slots__ = ['_path', 'connected', '_device_index', 'kwargs']
+    __slots__ = ['_path', '_device_index', 'kwargs']
 
     DEVICE_ROOT_PATH = '/sys/class'
 
@@ -158,7 +161,7 @@ class Device(object):
             d = ev3dev.Device('tacho-motor', address='outA')
             s = ev3dev.Device('lego-sensor', driver_name=['lego-ev3-us', 'lego-nxt-us'])
 
-        When connected succesfully, the `connected` attribute is set to True.
+        If there was no valid connected device, an error is thrown.
         """
 
         classpath = abspath(Device.DEVICE_ROOT_PATH + '/' + class_name)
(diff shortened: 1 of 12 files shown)


ev3dev/__init__.py
```diff
@@ -124,13 +124,16 @@ def matches(attribute, pattern):
                 yield f
 
 
+class DeviceNotFoundException(Exception):
```
dwalton76: would do `class DeviceNotFound(Exception)`

ev3dev/__init__.py
```diff
@@ -174,23 +177,24 @@ def get_index(file):
...
             try:
                 name = next(list_device_names(classpath, name_pattern, **kwargs))
                 self._path = classpath + '/' + name
                 self._device_index = get_index(name)
-                self.connected = True
             except StopIteration:
                 self._path = None
                 self._device_index = None
-                self.connected = False
+                raise DeviceNotFoundException("No device matching the given requirements was found.")
```
dwalton76: `"%s is not connected" % self` would let the user know exactly which device isn't connected

ev3dev/__init__.py
```diff
@@ -256,7 +252,7 @@ def _raise_friendly_access_error(self, driver_error, attribute):
             # We will assume that a file-not-found error is the result of a disconnected device
             # rather than a library error. If that isn't the case, at a minimum the underlying
             # error info will be printed for debugging.
-            raise Exception("%s is no longer connected" % self) from driver_error
+            raise DeviceNotFoundException("%s is no longer connected" % self) from driver_error
```
dwalton76: s/DeviceNotFoundException/DeviceNotFound/

tests/api_tests.py
```diff
@@ -21,25 +21,20 @@ def test_device(self):
...
-        self.assertTrue(not d.connected)
+        with self.assertRaises(ev3.DeviceNotFoundException):
+            d = ev3.Device('tacho-motor', 'motor*', address='outA', driver_name='not-valid')
 
         d = ev3.Device('lego-sensor', 'sensor*')
-        self.assertTrue(d.connected)
 
-        d = ev3.Device('this-does-not-exist')
-        self.assertFalse(d.connected)
+        with self.assertRaises(ev3.DeviceNotFoundException):
```
dwalton76: same for these two

dwalton76: other than that it looks good to me

WasabiFan: 😟 Why didn't our tests catch this? That's... quite worrying.

dwalton76:
`I haven't tested this on-device yet.` @WasabiFan have you had a chance to test it on a device?  If so it looks good to me.

WasabiFan:
@dwalton76 sorry, I'm quite busy right now so might not be able to take another look at these PRs for another few weeks. If I do get some time I'll report back.

WasabiFan:
I got a chance to test this on my EV3. I instantiated a motor with motors connected and without and it operated as expected.

ddemidov (approved):

Issue #406, opened 2017-10-10
motors API: passing something other than speed percentage


dwalton76:
Today in `on_for_rotations()`, etc the user passes the speed_pct because this is how the EV3-G block works.  I started updating the MINDCUBER demo to use the new API but I end up doing the following all over the place (two lines, one to cald the speed_pct and a second to call the API)
```
        speed_pct = self.turntable.dps_to_speed(MindCuber.rotate_speed)
        self.turntable.on_to_position(speed_pct, temp_pos)
```

which works just fine but doesn't look super clean (side note, `on_to_position()` is new).  What if the `on_to_XYZ()` methods accepted either speed_pct or `(speed_type, speed_value)` tuple such as the following
```
        self.turntable.on_to_position(('dps', MindCuber.rotate_speed), temp_pos)
```
Valid types there could be 'dps', 'dpm', 'rps', or 'rpm'.

WasabiFan:
Maybe this is just my non-Python background, but doesn't this style feel a bit unnatural? To an experienced programmer, it seems like there would be a more direct and obvious way of specifying this value; to an inexperienced one, it would be quite easy to misinterpret the grouping and be left completely confused. Why not add other named parameters for `speed_dps`, `speed_rpm`, etc.?

dwalton76:
we could do
```
   def on_for_rotations(self, speed_pct, rotations, brake=True, block=True, speed_rps=None, speed_rpm=None, speed_dps=None, speed_dpm=None)
```

where the user would then call via
```
    motor.on_for_rotations(None, 3, speed_rps=400)
```

but I was worried that might be confusing when comparing our API vs the EV3-G block since the speed parameter is out of order (in the block it is specified first).

WasabiFan:
That seems cleaner to me, but I'd be interested to hear what others think.

dwalton76:
Another option would be something like
```
class SpeedRPS(int):
    pass

class SpeedRPM(int):
    pass

class SpeedDPS(int):
    pass

class SpeedDPM(int):
    pass
```

and then the user would call via
```
    motor.on_for_rotations(SpeedRPS(400), 3)
```

That does look a little less hackish doesn't it

dwalton76:
For the `SpeedRPS` approach `on_for_rotations()` would need to call a function that uses isinstance to correctly convert the SpeedRPS value to speed percentage.

WasabiFan:
That seems like a pretty nice option as well.

[...]

PR #407, opened 2017-10-10
Add on_to_position(), on_for_XYZ accept additional speed inputs

dwalton76:


dwalton76:
Implements the idea floated in #406

dwalton76:
Will fix fake-sys to have `max_speed`

ev3dev/motor.py
```diff
@@ -67,6 +67,58 @@
...
+class SpeedDPS(SpeedInteger):
+    """
+    Speed in degrees-per-second
+    """
+
+    def __str__(self):
+        return ("%d dps" % self)
+
+    def get_speed_pct(self, motor):
+        return motor.dps_to_speed(self)
```
WasabiFan: Doesn't it seem a bit odd that we convert degrees-per-second into percentage of max speed just to convert it back? I think this is one of those cases where we should choose the native unit and base everything on that -- degrees per second are the reasonable "scientific" units whereas the percentage isn't.

ev3dev/motor.py
```diff
@@ -67,6 +67,58 @@
...
+class SpeedInteger(int):
+    pass
+
+
+class SpeedRPS(SpeedInteger):
+    """
+    Speed in rotations-per-second
+    """
+
+    def __str__(self):
```
WasabiFan, line 79: :+1:

ev3dev/motor.py
```diff
@@ -732,6 +792,46 @@ def wait_while(self, s, timeout=None):
         """
         return self.wait(lambda state: s not in state, timeout)
 
+    def rps_to_speed(self, rps):
+        """
+        Return the motor speed percentage to achieve rps (rotations-per-second)
+        """
+        assert abs(rps) <= self._max_rps, "%s maximum rotations-per-second is %s, %s was requested" % (self, self._max_rps, rps)
```
WasabiFan: What do you think about moving these conversions into the individual unit classes (or somewhere else)?

ev3dev/motor.py
dwalton76: A little...but the fact that we set the motor speed in terms of dps and EV3-G uses percentages means there is going to be some funkyness.  I can live with converting this one and then later having it converted back, since it means we treat all of these SpeedInteger classes consistently.

ev3dev/motor.py
dwalton76: sounds good...I am going to change `Motor._max_rps` to `Motor.max_rps` though since `SpeedRPS` will be accessing it to do the conversion

ev3dev/motor.py
```diff
@@ -80,7 +80,10 @@ def __str__(self):
         return ("%d rps" % self)
 
     def get_speed_pct(self, motor):
-        return motor.rps_to_speed(self)
+        """
+        Return the motor speed percentage to achieve desired rotations-per-second
+        """
+        return float(self/motor.max_rps) * 100
```
WasabiFan: Side-note: does this actually require the `float` conversion? As of Python 3 it'll do floating-point division anyway.

WasabiFan (approved):
Looks good to me! Last thing: docs on the methods that accept the new speed values?

ev3dev/motor.py
dwalton76: I still do this out of python2 habit...will drop the float call

PR #408, opened 2017-10-10
Add a wait_until_angle_changed_by() method to GyroSensor

gregcowell:
This adds a method to the GyroSensor class that simplifies the required code when using the gyro sensor to make accurate turns. Here is a short example:
```
motor_pair = MoveSteering(OUTPUT_B, OUTPUT_C)
gyro = GyroSensor()

# Spin robot to the left
motor_pair.on(steering=-100, speed_pct=5)

# Wait until angle changed by 90 degrees
gyro.wait_until_angle_changed_by(90)

# Stop motors
motor_pair.off()
```
Have discussed this with @dwalton76.

### DIFF
--- a/ev3dev/sensor/lego.py
+++ b/ev3dev/sensor/lego.py
@@ -430,6 +430,17 @@ class GyroSensor(Sensor):
         self.mode = self.MODE_GYRO_ANG
         self._direct = self.set_attr_raw(self._direct, 'direct', 17)
 
+    def wait_until_angle_changed_by(self, delta):
+        """
+        Wait until angle has changed by specified amount.
+        """
+        assert self.mode in (self.MODE_GYRO_G_A, self.MODE_GYRO_ANG,
+                             self.MODE_TILT_ANG),\
+            'Gyro mode should be MODE_GYRO_ANG, MODE_GYRO_G_A or MODE_TILT_ANG'
+        start_angle = self.value(0)
+        while abs(start_angle - self.value(0)) < delta:
+            time.sleep(0.01)
+
 
 class InfraredSensor(Sensor, ButtonBase):
     """


WasabiFan (approved):

ev3dev/sensor/lego.py
```diff
@@ -430,6 +430,14 @@ def reset(self):
         self.mode = self.MODE_GYRO_ANG
         self._direct = self.set_attr_raw(self._direct, 'direct', 17)
 
+    def wait_until_changed_by(self, delta):
+        """
+        Wait until angle has changed by specified amount.
+        """
+        start_angle = self.value()
```
dwalton76: - would call it something like `wait_until_angle_changed_by()`
- have it assert the mode is one of MODE_GYRO_ANG, MODE_GYRO_G_A, or MODE_TILT_ANG
- if it is MODE_GYRO_G_A you'll want to use `self.value(1)`, for the other two `self.value(0)`
- you could also do a sibling `wait_until_rate_changed_by()` method...or does that make any sense? Maybe `wait_until_rate()` instead?

WasabiFan: > you could also do a sibling wait_until_rate_changed_by() method...or does that make any sense? Maybe wait_until_rate() instead?

Yeah... that makes no sense 😀

gregcowell: Thanks, sounds great. I'll change the name and add the assert code. I think the equivalent EV3-G block has functionality similar to wait_until_rate_changed_by() and wait_until_rate(). With wait_until_rate(), I need to have a think about how that should work. In EV3-G, you have to specify a comparison operator like >=, <= etc. 

I'll have a good look at the EV3-G wait block functionality. I'm assuming that the goal of the high level API is to mirror how EV3-G works as closely as possible. 

Would it be okay to address wait_until_rate_changed_by() and wait_until_rate() in a separate pull request ? Or is it better to include it in this one ?

dwalton76: > Would it be okay to address wait_until_rate_changed_by() and wait_until_rate() in a separate pull request

yep that would be fine

gregcowell: Looks like MODE_GYRO_G_A actually also uses self.value(0). The rate seems to be at self.value(1).

The rate_and_angle property might need to be renamed angle_and_rate or the order of the return statement reversed.

ddemidov: > would call it something like wait_until_angle_changed_by()

> you could also do a sibling wait_until_rate_changed_by() method...

We have `Motor.wait()` with generic condition. Would it make sense to add some common predicates like `angle_changed_by(n)`, `rate_changed_by(n)`? To be used like
```py
m.wait(m.angle_changed_by(n))
```
The predicates could be implemented as
```py
def angle_changed_by(self, n):
    cur_angle = self.angle
    def changed_by():
      return abs(self.angle - cur_angle) >= n
```

gregcowell: I've updated the pull request with the new method name and assert code. Tested okay on my robot. I could not test MODE_TILT_ANG as my robot's gyro sensor does not support that. MODE_GYRO_ANG and MODE_GYRO_G_A worked fine.

gregcowell: >We have Motor.wait() with generic condition. Would it make sense to add some common predicates like angle_changed_by(n), rate_changed_by(n)? 

I'm thinking of wait_until_angle_changed_by() as not just being associated with motors. For example, it could be used to detect when the robot's angle has been changed by some external event. Also I think it works very similarly to the equivalent EV3-G block and would be easy to understand for users familiar with EV3-G.

ddemidov (approved):

ev3dev/sensor/lego.py
ddemidov: > wait_until_angle_changed_by() as not just being associated with motors.

Yes, please ignore my comment. Was made before the first coffee cup in the morning.

ev3dev/sensor/lego.py
```diff
@@ -430,6 +430,18 @@ def reset(self):
         self.mode = self.MODE_GYRO_ANG
         self._direct = self.set_attr_raw(self._direct, 'direct', 17)
 
+    def wait_until_angle_changed_by(self, delta):
+        """
+        Wait until angle has changed by specified amount.
+        """
+        mode = self.mode
+        assert mode == self.MODE_GYRO_G_A or mode == self.MODE_GYRO_ANG or\
+            mode == self.MODE_TILT_ANG,\
```
dwalton76: You can save a few CPU cycles via `assert mode in (self.MODE_GYRO_GA, self.MODE_GYRO_ANG, self.MODE_TILT_ANG)`

gregcowell: Thanks, I've made that change.

dwalton76 (approved):
Looks good

Issue #356, opened 2017-09-19
EV3-G API: display


dwalton76:
![display-01](https://user-images.githubusercontent.com/468309/30606535-02bac070-9d40-11e7-8437-94807f035fcf.png)

dwalton76:
Should we rename `display.py -> screen.py` since the class is `Screen` or do we want leave to the filename alone and change our class from `Screen` to `Display`?

dwalton76:
I am going to guess that we cannot include the screen `.bmp` files here due to copyright reasons?
https://sites.google.com/site/ev3basic/ev3-basic-programming/using-buttons-the-screen-and-the-leds/lego-ev3-standard-bitmaps

ddemidov:
> Should we rename

I vote to stick with `Screen` as class name and rename the file.

WasabiFan:
https://github.com/ev3dev/ev3dev/issues/970

> Should we rename...

When I say "screen" a few times in my head it loses all meaning... I feel like "display" is a bit more clear. I don't have a strong opinion though.

ensonic:
For the assents, we could start a collection of free replacement images. I don't think we can get LEGO to license them to us in a useful way, since content if often done by sub contracting companies.
For the bars, gauges and eyes, I'd love to have classes where one can properly animate them though.

dlech:
See https://github.com/ev3dev/ev3dev/issues/970

dwalton76:
Let's look to see what EV3-G calls this block (screen or display) and follow that...I'll take a look this evening.

dwalton76:
ok EV3-G calls their block `Display` to lets stick with display.py but change the name of our class from Screen to Display so we are 1:1

dwalton76:
Some references on display

- http://ev3lessons.com/translations/en-us/beginner/Display.pdf
- http://www.mainerobotics.org/uploads/8/3/4/4/8344007/using_the_ev3_display_block.pdf

PR #410, opened 2017-10-12
EV3-G API Display

dwalton76:
For issue #356

### DIFF
--- a/ev3dev/display.py
+++ b/ev3dev/display.py
@@ -335,11 +335,11 @@ class Display(FbMem):
             return self.draw.text((x, y), text, fill=text_color)
 
     def text_grid(self, text, clear_screen=True, x=0, y=0, text_color='black', font=None):
-        assert 0 < x < Display.GRID_COLUMNS,\
+        assert 0 <= x < Display.GRID_COLUMNS,\
             "grid columns must be between 0 and %d, %d was requested" %\
             ((Display.GRID_COLUMNS - 1, x))
 
-        assert 0 < y < Display.GRID_ROWS,\
+        assert 0 <= y < Display.GRID_ROWS,\
             "grid rows must be between 0 and %d, %d was requested" %\
             ((Display.GRID_COLUMNS - 1), y)
 
--- a/ev3dev/display.py
+++ b/ev3dev/display.py
@@ -341,7 +341,7 @@ class Display(FbMem):
 
         assert 0 <= y < Display.GRID_ROWS,\
             "grid rows must be between 0 and %d, %d was requested" %\
-            ((Display.GRID_COLUMNS - 1), y)
+            ((Display.GRID_ROWS - 1), y)
 
         return self.text_pixels(text, clear_screen,
                                 x * Display.GRID_COLUMN_PIXELS,
--- a/ev3dev/display.py
+++ b/ev3dev/display.py
@@ -181,13 +181,18 @@ class FbMem(object):
         )
 
 
-class Screen(FbMem):
+class Display(FbMem):
     """
     A convenience wrapper for the FbMem class.
     Provides drawing functions from the python imaging library (PIL).
     """
 
-    def __init__(self):
+    GRID_COLUMNS = 22
+    GRID_COLUMN_PIXELS = 8
+    GRID_ROWS = 12
+    GRID_ROW_PIXELS = 10
+
+    def __init__(self, desc='Display'):
         FbMem.__init__(self)
 
         self._img = Image.new(
@@ -196,6 +201,10 @@ class Screen(FbMem):
                 "white")
 
         self._draw = ImageDraw.Draw(self._img)
+        self.desc = desc
+
+    def __str__(self):
+        return self.desc
 
     @property
     def xres(self):
@@ -312,6 +321,20 @@ class Screen(FbMem):
             return self.draw.text((x, y), text, fill=text_color, font=fonts.load(font))
         else:
             return self.draw.text((x, y), text, fill=text_color)
+
+    def text_grid(self, text, clear_screen=True, x=0, y=0, text_color='black', font=None):
+        assert 0 < x < Display.GRID_COLUMNS,\
+            "grid columns must be between 0 and %d, %d was requested" %\
+            ((Display.GRID_COLUMNS - 1, x))
+
+        assert 0 < y < Display.GRID_ROWS,\
+            "grid rows must be between 0 and %d, %d was requested" %\
+            ((Display.GRID_COLUMNS - 1), y)
+
+        return self.text_pixels(text, clear_screen,
+                                x * Display.GRID_COLUMN_PIXELS,
+                                y * Display.GRID_ROW_PIXELS,
+                                text_color, font)
 
     def reset_screen(self):
         self.clear()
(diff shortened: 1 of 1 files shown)


ev3dev/display.py
```diff
@@ -228,18 +238,6 @@ def draw(self):
...
-    def image(self):
-        """
-        Returns a handle to PIL.Image class that is backing the screen. This can
-        be accessed for blitting images to the screen.
-
-        Example::
-
-            screen.image.paste(picture, (0, 0))
-        """
-        return self._img
```
WasabiFan: I think we should do what we can to keep this accessor in some form, as it's what all existing display code is based on.

dwalton76: I'm trying to remember why I removed this and can't come up with a good reason....will put this back and rename the other `def image` to `def image_filename`

ev3dev/display.py
```diff
@@ -267,3 +277,77 @@ def update(self):
...
+            self.clear()
+
+        filename_im = Image.open(filename)
+
+        if x2 is not None and y2 is not None:
+            return self._img.paste(filename_im, (x1, y1, x2, y2))
+        else:
+            return self._img.paste(filename_im, (x1, y1))
+
+    def shapes_line(self, clear_screen=True, x1=0, y1=0, x2=0, y2=0, line_color='black', width=1):
```
WasabiFan: Why not `draw_line`, `draw_circle`, etc.?

ev3dev/display.py
```diff
@@ -267,3 +277,77 @@ def update(self):
...
+
+        if clear_screen:
+            self.clear()
+
+        return self.draw.line((x1, y1, x2, y2), fill=line_color, width=width)
+
+    def shapes_circle(self, clear_screen=True, x=50, y=50, radius=40, fill_color='black', outline_color='black'):
```
WasabiFan: Document color options and how to make it transparent

ev3dev/display.py
```diff
@@ -267,3 +277,77 @@ def update(self):
...
+
+        if clear_screen:
+            self.clear()
+
+        x1 = x - radius
+        y1 = y - radius
+        x2 = x + radius
+        y2 = y + radius
+
+        return self.draw.ellipse((x1, y1, x2, y2), fill=fill_color, outline=outline_color)
```
WasabiFan, line 316: What would happen if you passed, say, `red` to this function in the EV3 (the EV3 has a monochrome display)? Does it convert the color brightness to a single greyscale channel, or do something else?

ev3dev/display.py
```diff
@@ -267,3 +277,77 @@ def update(self):
...
+        if clear_screen:
+            self.clear()
+
+        if font is not None:
+            assert font in fonts.available(), "%s is an invalid font" % font
+            return self.draw.text((x, y), text, fill=text_color, font=fonts.load(font))
+        else:
+            return self.draw.text((x, y), text, fill=text_color)
+
+    def text_grid(self, text, clear_screen=True, x=0, y=0, text_color='black', font=None):
```
WasabiFan, line 364: Document what these two text methods do... I'm a bit confused myself

dwalton76:
I just noticed the EV3-G display block for text allows you to specify font size in terms of
```
0 – small font
1 – small, bold font
2 – large font
```

@ddemidov any recommendations on what fonts we should use for those three choices?

ev3dev/display.py
dwalton76: I just went went the names the @dlech had in the image posted in https://github.com/rhempel/ev3dev-lang-python/issues/356  I'm not sure if those are official names of the EV3-G equivalents or not though.  I agree though `draw_line`, etc sounds better.

ev3dev/display.py
dwalton76: How do you make it transparent?  We are way beyond my experience level with PIL :)

Regarding supported color names, PIL says they support `Common HTML colour names`, I could never find a definitive list of the color names they support so didn't attempt to include that in the doc string.

WasabiFan: oh, I don't know... I've never used it :laughing: My guess is to pass `None`.

ev3dev/display.py
dwalton76: `red` it converted to `black` on my EV3. I don't know if there is a cutoff where something close to white translates to white instead of black or not.

ev3dev/display.py
dwalton76: k

dwalton76: Isn't there a page somewhere that shows what all of the different fonts look like on the EV3 screen?  I'm looking but can't find it...was going to include that in the doc string

ev3dev/display.py
WasabiFan: This is where it comes from:

image

I don't think calling them `draw_line`/`draw_circle`/etc. will be confusing. It might even work to just call it `line`/`circle`...

ev3dev/display.py
WasabiFan: http://python-ev3dev.readthedocs.io/en/latest/other.html#ev3dev.fonts.load (https://github.com/rhempel/ev3dev-lang-python/commit/e6a1ddef12819e599355261a1fd53c5616c5cdf2)

I couldn't find info on how this was generated... somewhere I think @ddemidov had a Jupyter notebook with some code for it.

dwalton76: found it
http://ev3dev-lang.readthedocs.io/projects/python-ev3dev/en/stable/other.html#bitmap-fonts

ev3dev/display.py
```diff
@@ -179,14 +181,18 @@ def _map_fb_memory(fbfid, fix_info):
...
     """
 
-    def __init__(self):
-        from PIL import Image, ImageDraw
+    GRID_COLUMNS = 22
+    GRID_COLUMN_PIXELS = 8
+    GRID_ROWS = 12
+    GRID_ROW_PIXELS = 10
+
+    def __init__(self, desc='Display'):
```
WasabiFan, line 195: I'm curious... what is the description for?

ev3dev/display.py
```diff
@@ -267,3 +277,120 @@ def update(self):
...
+        """
+        Draw a line from (x1, y1) to (x2, y2)
+        """
+
+        if clear_screen:
+            self.clear()
+
+        return self.draw.line((x1, y1, x2, y2), fill=line_color, width=width)
+
+    def circle(self, clear_screen=True, x=50, y=50, radius=40, fill_color='black', outline_color='black'):
```
WasabiFan, line 303: These defaults seem a bit random :laughing:

I swear, this is my last comment :wink: I need to be better about all-in-one reviews.

WasabiFan (approved):

ev3dev/display.py
dwalton76: It is just so if you print the object as a string the `__str__` has something to work with

ev3dev/display.py
dwalton76: yeah those are just so the user sees a full circle on the screen if they are kicking the tires and call `circle()` without passing any args

ev3dev/display.py
ddemidov: #225 links to the notebook with the code used to generate the image.

PR #411, opened 2017-10-13
Initial set of updates to documentation for Stretch

### LINKED ISSUE #399: Create an upgrading guide for Jessie to Stretch

WasabiFan:
This is an initial set of documentation updates for Stretch. I rearranged the README so it made a bit more sense to me, added information about the VSCode extension, and added an "upgrading guide".

I'd like a thorough check from others to make sure I didn't do anything stupid

TODOs:

- [ ] Update information on required kernel version
- [x] Unfinished area at the bottom of the new file (I expect to create an additional page for this; the idea is to have a page explaining the new features)
- [x] Documentation for special sensor classes is broken (time to learn how this rST stuff works)
- [x] (existing) "Module Index" link at bottom of homepage broken

I'm finding it a bit ugly to have to explain adding additional classes to the import list; read it over and tell me what you guys think.

Fixes #399


WasabiFan:
It looks like there's some messed-up formatting which I need to fix: http://ev3dev-lang.readthedocs.io/projects/python-ev3dev/en/stretch-documentation/

**If you aren't a maintainer, don't click this link; it isn't complete documentation.**

README.rst
```diff
@@ -36,109 +36,96 @@ your Python code to a file.
 Make sure that you look at the `User Resources`_ section as well for links
 to documentation and larger examples.
 
-Usage Examples
+Usage
 --------------
```
ddemidov: The underlines should be same length as the titles.

README.rst
```diff
@@ -36,109 +36,96 @@ your Python code to a file.
...
-.. code-block:: python
+You should use the ``.py`` extension for your file, e.g. ``my-file.py``.
 
-  import ev3dev.ev3 as ev3
+Important: Make your script executable
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-If you are using a BrickPi, use this line:
+To be able to run your Python file, **your program must be executable**. If
+you're using the `ev3dev Visual Studio Code extension`_, this step will be
```
ddemidov: 'you are'

ddemidov: I think "If you're using the `ev3dev Visual Studio Code extension`_, **you can skip this step**, since it will be automatically performed when you download your code to the brick." sounds better. Otherwise it sounds like you can skip this step in any case.

README.rst
```diff
@@ -36,109 +36,96 @@ your Python code to a file.
...
+          leds.set_color(leds.led_groups.LEFT, leds.led_colors.RED)
 
 Running a motor
 ~~~~~~~~~~~~~~~
 
-Now plug a motor into the ``A`` port and paste this code into the Python prompt.
-This little program will run the motor at 500 ticks per second, which on the EV3
-"large" motors equates to around 1.4 rotations per second, for three seconds
-(3000 milliseconds).
+This will run a LEGO Large Motor at 75% power for 5 rotations.
```
ddemidov: Is it power or speed?

README.rst
```diff
@@ -36,109 +36,96 @@ your Python code to a file.
...
-  m = ev3.LargeMotor('outA')
-  m.run_timed(time_sp=3000, speed_sp=500)
-
-The units for ``speed_sp`` that you see above are in "tacho ticks" per second.
-On the large EV3 motor, these equate to one tick per degree, so this is 500
-degress per second.
-
-
+  m = LargeMotor(OUTPUT_A)
+  m.on_for_rotations(75, 5)
```
ddemidov: This is offtopic, but looking at this line of code I feel it could be more readable. Is 75 here power or rotations? Why the power attribute is on the first place when the method is called `on_for_rotations`? Why not `m.on_for_rotations(5, power=75)`?

WasabiFan: We can use the new methodology from #407

WasabiFan:
@dwalton76 why's this closed?

dwalton76:
@WasabiFan weird it looks like when I deleted `develop` (since it was a duplicate of `stretch`) it closed any open PRs against it.

WasabiFan:
Note: we probably need to add a note somewhere in the Jessie docs to point people to Stretch, and maybe vice versa as well, because web search results might pick up both.

WasabiFan:
We should figure out how to make the `on_*` methods more discoverable. At the moment a docs page for a motor class has them listed 2/3rds down the class section, intermingled with a bunch of properties we hope no one will ever have to use. The easiest way is probably a usage example at the top of the section.

docs/conf.py
```diff
@@ -311,3 +311,5 @@
 
 # If true, do not generate a @detailmenu in the "Top" node's menu.
 #texinfo_no_detailmenu = False
+
+autodoc_member_order = 'bysource'
```
WasabiFan, line 316: I'd like some feedback on this change. I'm hoping it might make things a bit more logical to sort through... not sure.

WasabiFan:
The current implementations of the `Move*` classes are a bit weird right now. `MoveTank` extends from `MotorSet` and implements the `on_for_*` and related methods appropriately. `MoveSteering` then extends from `MoveTank` and overrides those methods to take the steering parameter rather than individual values. But while it implements `on`, it _doesn't_ implement `off`. So in that case we're expecting people to realize that they should look at the super class for that particular method. Then we have the `MoveJoystick` class which also extends `MoveTank`; but because running for a particular distance doesn't make much sense for joystick control (it is almost solely for human operation), we don't override the other `on_for_*` methods with new parameters. That means that a `MoveJoystick` has an `on_for_rotations` method, for example, but it's actually pointless because it only takes tank parameters. This isn't great for either documentation purposes or IDE suggestions.

The two solutions I see are:
- Explicitly implement an `off` method on the `MoveSteering` class (quick fix). This at least removes the primary mixed messagingby eliminating the need to look at the superclass, although there are other methods on the superclasses which you might want to use anyway, e.g. state checking.
- Refactor `MoveJoystick` so it isn't derived from the other classes. This means that you lose the potentially helpful properties and methods on that class from its parents. This might work well if we had a publically-accessible instance of the inner `MoveTank` which could be operated on directly when needed.

WasabiFan:
@dwalton76 @ddemidov Could I get a review of this? I'd like to see if we can merge this and get a Stretch release out.

Question: where should we put info on usage of the new features? A dedicated page just for this version transition? An announcement (ev3dev.org maybe)? A new changelog? Just with each individual class?

WasabiFan:
Item for discussion: Where do we link to the new upgrading guide?

ddemidov:
> Item for discussion: Where do we link to the new upgrading guide?

Would a "danger note" somewhere around Getting Started / Usage work? Something like

> This documents the new "stretch" version of the library. If you are using older, "jessie" ev3dev installation, go to the older documentation [here]. If you are upgrading from "jessie", look at the [upgrade guide].

docs/conf.py
ddemidov: Both `bysource` (more logical order) and `alphabetical` (easier to find things) make sense to me. Too bad users can not change this dynamically.

ddemidov (approved):
Wow, that is a lot of work! I went though this commit by commit, and it looks good to me. I did not check for completeness (or that the compiled docs actually look good), but I think its better if we merge this and then keep updating the docs as necessary.

WasabiFan:
Thanks for the review! I've merged this and will go update #480 to rebase it on `stretch`.

ndward:
I'd like to make some more videos for beginners about the VS Code workflow but I don't want to make videos that will quickly be made out of date by the (exciting) introduction of the movetank and movesteering classes. It would be very helpful if you could estimate a date for the initial Stretch release of the Python library.

WasabiFan:
I intend to release within a day after I merge #480 and #483, which contain important port name updates and some usability improvements for the `Move*` classes. Those are waiting on review (other maintainers have been more busy than I have recently). I just posted in #480 again as a reminder and hopefully we can get things finished and published soon. I do legitimately plan to get this released ASAP.


How about this: If they aren't already merged by mid-day Sunday (my time, PST) I'll merge them anyway :laughing: Then I'll test things out to make sure that all of that code works and publish to PyPi. It might take a day or two past that to either figure out how to publish the Debian package myself or get someone else to do it.

ddemidov:
I think I still have VM image I used to publish the debian packages, so I could try to make the release when you say it's ready. Although, it may be better if you tried to do it yourself :)

WasabiFan:
Yeah, I can definitely look into it. I know there's a docs page on it at ev3dev.org (which I might need to adapt for Stretch); the issue is that, once I have built the package, I'll need permissions to actually deploy it anywhere 😆 I'll first just need to get the thing building.

PR #412, opened 2017-10-14
Rename namespace: ev3dev -> ev3dev2

WasabiFan:
This is the batch rename operation we've been discussing for Stretch.

There are some test-related files from two years ago which don't seem to be referenced anywhere, so I decided to hold off on updating them until I knew what they were or if they were necessary. Any input on that would be appreciated.


WasabiFan:
I also didn't update the gyro balancer as it appears to use the old `RemoteControl` class and I didn't want to start making changes to things I'm not familiar with (I haven't used the remote control/IR sensor nor the gyro balancer).

dwalton76:
I can update GyroBalancer, filed #414 for that one

dwalton76 (approved):

ev3dev2/button.py
```diff
@@ -31,7 +31,7 @@
 import array
 import time
 import evdev
-from ev3dev import get_current_platform
+from ev3dev2 import get_current_platform
```
ddemidov: Will this work as `from . import get_current_platform`? Could be future-name-change-proof like this.

ev3dev2/button.py
```diff
@@ -46,22 +46,22 @@
 platform = get_current_platform()
 
 if platform == 'ev3':
-    from ev3dev._platform.ev3 import BUTTONS_FILENAME, EVDEV_DEVICE_NAME
+    from ev3dev2._platform.ev3 import BUTTONS_FILENAME, EVDEV_DEVICE_NAME
```
ddemidov: Same here

tests/api_tests.py
```diff
@@ -9,36 +9,36 @@
 from populate_arena import populate_arena
 from clean_arena    import clean_arena
 
-import ev3dev as ev3
```
ddemidov, line 12: Why not just change this line to `import ev3dev2 as ev3`?

WasabiFan: Because it's no longer generic "ev3" interface... It's just the couple of utility methods we dumped into the init file. Do you have a different name suggestion?

ddemidov: Ok, that is a good point. But `ev3` is still shorter than `ev3dev2`, and `ev3dev2` still contains `ev3` in the name. So it is kind of not generic in the first place.

WasabiFan: Agreed :laughing: I'll have to think about it a bit and see what I think in the morning (this 12-hour offset makes communication really weird!)

ddemidov (approved):
This looks good, but needs manual rebasing.

dwalton76:
@WasabiFan I think we are good to go once the conflicts are resolved

WasabiFan:
Just merged; it looks like tests passed, so I'm going to merge this. Thanks for the reviews, apologies I've been so slow!

Issue #413, opened 2017-10-14
ColorSensor: add lab() method


dwalton76:
Returning the color in Lab color space would be handy

WasabiFan:
For a typical user, HSL or HSV would probably be most useful; I very commonly use HSL in my vision algorithms because it's quite easy to use for color segmentation.

PR #417, opened 2017-10-17
Added ColorSensor.lab(), hsv() and hls()

dwalton76:
Resolves issue #413


WasabiFan (approved):
Lots of magic numbers, but I assume it works 😆 If you were super over-achieving you could write tests for this, but that's probably overkill.

PR #421, opened 2017-10-21
Update GyroBalancer module

gregcowell:
This module needs to be updated so that the demo script for the BALANC3R robot will work with the latest changes to the Python bindings. I have ported Laurens Valk's latest segway script.  @dwalton76 did the previous port and I have followed the same style.

### DIFF
--- a/ev3dev/control/GyroBalancer.py
+++ b/ev3dev/control/GyroBalancer.py
@@ -1,4 +1,10 @@
-#!/usr/bin/env python3
+"""Module for a robot that stands on two wheels and uses a gyro sensor.
+
+The robot (eg. BALANC3R) will to keep its balance and move in response to
+the remote control. This code was adapted from Laurens Valk's script at
+https://github.com/laurensvalk/segway.
+
+"""
 # The MIT License (MIT)
 #
 # Copyright (c) 2016 Laurens Valk (laurensvalk@gmail.com)
@@ -10,8 +16,8 @@
 # copies of the Software, and to permit persons to whom the Software is
 # furnished to do so, subject to the following conditions:
 #
-# The above copyright notice and this permission notice shall be included in all
-# copies or substantial portions of the Software.
+# The above copyright notice and this permission notice shall be included in
+# all copies or substantial portions of the Software.
 #
 # THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
 # IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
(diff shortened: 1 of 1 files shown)


WasabiFan:
> ev3dev/control/GyroBalancer.py 100644 → 100755 

Is this actually a standalone executable?

gregcowell:
No, it’s still just a module like before.

gregcowell:
Thanks @WasabiFan, I've fixed the permissions on the file now.

ev3dev/control/GyroBalancer.py
```diff
@@ -1,398 +1,477 @@
...
-# The above copyright notice and this permission notice shall be included in all
-# copies or substantial portions of the Software.
-#
-# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
-# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
-# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
-# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
-# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
-# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
-# SOFTWARE.
```
dwalton76: If we are still using Laurens PID code (which it looks like we are) we should leave his copyright in place unless he has given us permission to remove it.

ev3dev/control/GyroBalancer.py
```diff
@@ -1,398 +1,477 @@
...
     outfile.flush()
 
 
-# Function to set the duty cycle of the motors
-def SetDuty(motorDutyFileHandle, duty):
-    # Clamp the value between -100 and 100
-    duty = min(max(duty, -100), 100)
+def eprint(*args, **kwargs):
+    """Debug print (https://stackoverflow.com/a/14981125)."""
+    print(*args, file=sys.stderr, **kwargs)
```
dwalton76: `log.info(MESSAGE)` or `log.debug(MESSAGE)` would seem better

ev3dev/control/GyroBalancer.py
```diff
@@ -1,398 +1,477 @@
...
+    elif dutyInt < 0:
+        dutyInt = max(-100, dutyInt - frictionOffset)
 
     # Apply the signal to the motor
-    FastWrite(motorDutyFileHandle, duty)
+    FastWrite(motorDutyFileHandle, dutyInt)
 
 
-class GyroBalancer(LargeMotorPair):
+class GyroBalancer():
```
dwalton76: Would have it in inherit from `object`

ev3dev/control/GyroBalancer.py
```diff
@@ -1,398 +1,477 @@
...
+                # On the EV3, "1% speed" corresponds to 1.7 RPM (if speed
+                # control were enabled).
+                RPMperPerPercentSpeed = 1.7
+
+                # Convert this number to the speed in deg/s per "percent speed"
+                degPerSecPerPercentSpeed = RPMperPerPercentSpeed * 360 / 60
+
+                # Convert this number to the speed in rad/s per "percent speed"
+                radPerSecPerPercentSpeed = degPerSecPerPercentSpeed *\
+                    radiansPerDegree
```
dwalton76, line 206: It seems like lines 140-191 could be moved up before the `while True` loop as most of these are either opening filehandles or defining constants.  Especially the part where we reset() the two motors and set the mode to run_direct()...does that really need to happen everytime through the loop?

ev3dev/control/GyroBalancer.py
```diff
@@ -1,398 +1,477 @@
...
+
+                eprint("Hold robot upright. Press Touch Sensor to start.")
+
+                # Wait for Touch Sensor or any Button Press
+                while not self.touchSensor.is_pressed:
+                    time.sleep(0.01)
+
+                # Wait for release and proceed to calibration and balancing
+                while self.touchSensor.is_pressed:
+                    time.sleep(0.01)
```
dwalton76: You can use `self.touchSensor.wait_for_bump()` here (in place of lines 258->264)

ev3dev/control/GyroBalancer.py
dwalton76: Ignore this, I see now (this isn't the balancing loop)

dwalton76 (changes requested):

ev3dev/control/GyroBalancer.py
gregcowell: Even though Python 3 does not require it ? I thought that was only important for code that may be run under Python 2.

gregcowell:
Thanks @dwalton76. I've made all your requested changes.

ev3dev/control/GyroBalancer.py
gregcowell: Yep, there was a misleading comment that confused things. I've removed it.

gregcowell:
@dwalton76 , any more changes needed ? Happy to re-do from scratch if you had something else in mind.

dwalton76 (approved):

dwalton76:
@gregcowell looks good, thanks!

Issue #403, opened 2017-10-08
EV3-G API: Sound.play_note()


dwalton76:
![30606561-16a38a0e-9d40-11e7-9803-830261c4714b](https://user-images.githubusercontent.com/468309/31317060-06cb2ffc-ac08-11e7-9419-cdaf779be144.png)

@EricPobot we chatted about this for a bit in #358 ...can you take a look at implementing a
```
   def play_note(note, duration_ms, volume=100, play_type=PLAY_WAIT_FOR_COMPLETE)
```
in the Sound class?

dwalton76:
@EricPobot do you have the cycles to take a stab at this one?  I don't know enough music theory to implement this :(

ericpascual:
@dwalton76 I'll give it a look ASAP (maybe tonight) and keep you updated. 
I guess that "volume=100" means "volume at 100%". Right ?

EricPobot:
What is the specification of the duration value ? Does 1.0 represent a delay in seconds or something else like a quarter note for instance (i.e. the 'q' notation as used by play_song) ?

dlech:
duration is how long the note plays in seconds

EricPobot:
@dlech I've noted after asking that the example of the signature in the opening post shows "duration_ms". Since play_tone uses ms too, maybe we should keep the same unit (i.e. millis, not seconds) to be homogeneous.

EricPobot:
First implementation done (was not a big deal, since the play_tone method being already there, so it was just a matter of getting the frequency for the passed note symbol). 
I still need to do the appropriate work (fork + pull request) for submitting the changes. This will have to wait until tomorrow since time to go to bed now ;)
I've made some minor changes in a couple of place to remove PEP compliance warnings.

dlech:
> maybe we should keep the same unit (i.e. millis, not seconds) to be homogeneous.

It seems to me that in general, python uses seconds (e.g. `time.sleep()`), so that would be my preference. (and tone playback could be changed to match).

ericpascual:
Agree. Let's align all involved methods to seconds. I'll integrate this in my changes.

EricPobot:
After some thinking, and because the underlying system commands use integer counts of milliseconds, using milliseconds everywhere seemed more appropriate.

Implementation is finished. Am I supposed to push my changes to the repo () ?

WasabiFan:
@EricPobot open a pull request and we can take a look.

EricPobot:
[...]

PR #431, opened 2017-11-16
Proposed implementation for issue #403

EricPobot:
Adds the play_note method, and changes duration parameters of methods exposed as EV3-G blocks so that values are provided in seconds.

Documentation has been added in some places, minor changes done for better PEP compliance and additional method parameter checking added.

### DIFF
--- a/ev3dev/sound.py
+++ b/ev3dev/sound.py
@@ -90,7 +90,7 @@ class Sound(object):
 
     def _validate_play_type(self, play_type):
         assert play_type in self.PLAY_TYPES, \
-            "Invalid play_type %s, must be one of %s" % (play_type, ','.join(self.PLAY_TYPES))
+            "Invalid play_type %s, must be one of %s" % (play_type, ','.join(str(t) for t in self.PLAY_TYPES))
 
     def beep(self, args=''):
         """
--- a/ev3dev/sound.py
+++ b/ev3dev/sound.py
@@ -25,7 +25,7 @@
 
 import sys
 
-if sys.version_info < (3,4):
+if sys.version_info < (3, 4):
     raise SystemError('Must be using Python 3.4 or higher')
 
 import os
@@ -69,6 +69,10 @@ class Sound(object):
             ('G4', 'h'),
             ('D5', 'h')
         ))
+
+    In order to mimic EV3-G API parameters, durations used in methods
+    exposed as EV3-G blocks for sound related operations are expressed
+    as a float number of seconds.
     """
 
     channel = None
@@ -85,7 +89,8 @@ class Sound(object):
     )
 
     def _validate_play_type(self, play_type):
-        assert play_type in self.PLAY_TYPES, "Invalid play_type %s, must be one of %s" % (play_type, ','.join(self.PLAY_TYPES))
+        assert play_type in self.PLAY_TYPES, \
+            "Invalid play_type %s, must be one of %s" % (play_type, ','.join(self.PLAY_TYPES))
 
     def beep(self, args=''):
         """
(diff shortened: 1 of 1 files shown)


ev3dev/sound.py
```diff
@@ -153,10 +161,37 @@ def beep_args(frequency=None, duration=None, delay=None):
...
 
-    def play_tone(self, frequency, duration_ms, delay_ms=100, volume=100, play_type=PLAY_WAIT_FOR_COMPLETE):
+    def play_tone(self, frequency, duration, delay=0.1, volume=100,
+                  play_type=PLAY_WAIT_FOR_COMPLETE):
+        """ Play a tone, specified by its frequency, duration, volume and final delay.
+
+        Args:
+            frequency (int): the tone frequency, in Hertz
+            duration (float): tone duration, in seconds
+            delay (float): delay after tone, in seconds
```
WasabiFan: What's the use for the delay? Why doesn't the caller do that instead?

ev3dev/sound.py
```diff
@@ -153,10 +161,37 @@ def beep_args(frequency=None, duration=None, delay=None):
...
+            the sound playing subprocess PID when no wait play type is selected, None otherwise
+
+        Raises:
+            ValueError: if invalid value for parameter(s)
+        """
         self._validate_play_type(play_type)
+
+        if duration <= 0:
+            raise ValueError('invalid duration (%s)' % duration)
+        if delay <= 0:
```
WasabiFan: Why can't delay be `0`?

ev3dev/sound.py
```diff
@@ -169,9 +204,43 @@ def play_tone(self, frequency, duration_ms, delay_ms=100, volume=100, play_type=
                 play = self.tone([(frequency, duration_ms, delay_ms)])
                 play.wait()
 
-    def play(self, wav_file, play_type=PLAY_WAIT_FOR_COMPLETE):
+    def play_note(self, note, duration, volume=100, play_type=PLAY_WAIT_FOR_COMPLETE):
+        """ Plays a note, given by its name as defined in ``_NOTE_FREQUENCIES``.
```
WasabiFan, line 211: For the musically-inept (e.g. me) could you include explanations to differentiate tone from note? I think tone is arbitrary frequency and notes are named frequencies, but I'm not the expert here.

WasabiFan (approved):

ev3dev/sound.py
ericpascual: You're 100% right. play_note() is only a convenience call for tone() with the corresponding frequency. The tone() method was already there when I started to collaborate to this module, and it can be useful for arbitrary frequencies. However, I'd suggest to remove the example show how to play a song with tone() since IMHO it can lead to some confusion (you question is a perfect example of such a side effect). I didn't take the decision on my own since not the author of this part of the code.

ev3dev/sound.py
ericpascual: It can be. Sorry for the copy/paste of this part of the sanity check code. I'll fix this ASAP.

ev3dev/sound.py
EricPobot: The caller can take care of the delay between successive tones, but using the parameter makes its code simpler by avoiding additional time.delay() calls. 

In addition, chances are that this will give more accurate timings, by removing the overhead added by the time.delay() additional call, the delay being managed directly by the beep system command.

ev3dev/sound.py
EricPobot: I've just pushed a fix for this.

dwalton76:
The only musical instrument I can play is the radio so my knowledge here is about zero :) Looks good to me. Thank you for taking care of this one.

Issue #429, opened 2017-11-02
GyroBalancer: separate drive control loop and tuning from balance loop


gregcowell:
@laurensvalk has suggested [here](https://github.com/ev3dev/ev3dev-lang-python-demo/issues/8#issuecomment-340253122) that it would be nice to separate the drive control loop from the balance loop in the GyroBalancer class. I agree. 

The balance loop is the most complicated part of the code and requires knowledge of control theory to fully understand. We could leave the balance loop inside the GyroBalancer class and separate out the drive control loop into an example user script. The GyroBalancer class could then be used for many different programs rather than for a single remote control application. For example, someone may want to program BALANC3R to find its way through a maze or follow a line. Being able to utilise the balancing loop code without the drive control loop / remote control code would be very useful.

A related issue is the requirement to be able to easily tune the balancing loop parameters from a user script. 

If it is agreed that these changes would be worthwhile, I'd like to work on them.

dwalton76:
Sounds good to me, go for it @gregcowell

dwalton76:
Thank you @gregcowell for working on this!

gregcowell:
> Thank you @gregcowell for working on this!

You're welcome! I really like working on this project.

PR #432, opened 2017-11-20
Separate GyroBalancer drive control loop from balance loop

gregcowell:
Please see issue #429. This facilitates the writing of simple scripts for controlling the movement of a GyroBalancer robot (eg.BALANC3R).

I recommend reviewing the new example scripts [here](https://github.com/ev3dev/ev3dev-lang-python-demo/pull/26) first. 

This pull request also adds the ability to enable debug mode. In debug mode, the GyroBalancer robot logs gyro reading and loop time statistics.

### DIFF
--- a/ev3dev/control/GyroBalancer.py
+++ b/ev3dev/control/GyroBalancer.py
@@ -29,44 +29,48 @@ https://github.com/laurensvalk/segway.
 
 import logging
 import time
+import json
+import queue
+import threading
 from collections import deque
 from ev3dev.power import PowerSupply
 from ev3dev.motor import LargeMotor, OUTPUT_A, OUTPUT_D
-from ev3dev.sensor.lego import GyroSensor, InfraredSensor, TouchSensor
+from ev3dev.sensor.lego import GyroSensor, TouchSensor
+from ev3dev.sound import Sound
+from collections import OrderedDict
 
 log = logging.getLogger(__name__)
 
-########################################################################
-# File I/O functions
-########################################################################
+# Constants
+RAD_PER_DEG = 3.14159265/180
 
+# EV3 Platform specific constants
 
-def FastRead(infile):
-    """Function for fast reading from sensor files."""
-    infile.seek(0)
-    return(int(infile.read().decode().strip()))
+# For the LEGO EV3 Gyro in Rate mode, 1 unit = 1 deg/s
+DEG_PER_SEC_PER_RAW_GYRO_UNIT = 1
 
+# Express the above as the rate in rad/s per gyro unit
+RAD_PER_SEC_PER_RAW_GYRO_UNIT = DEG_PER_SEC_PER_RAW_GYRO_UNIT * RAD_PER_DEG
 
-def FastWrite(outfile, value):
-    """Function for fast writing to motor files."""
-    outfile.truncate(0)
-    outfile.write(str(int(value)))
-    outfile.flush()
+# For the LEGO EV3 Large Motor 1 unit = 1 deg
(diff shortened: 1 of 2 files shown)


ev3dev/control/GyroBalancer.py
```diff
@@ -29,44 +29,48 @@
...
+from ev3dev.sound import Sound
+from collections import OrderedDict
 
 log = logging.getLogger(__name__)
 
-########################################################################
-# File I/O functions
-########################################################################
+# Constants
+RAD_PER_DEG = 3.14159265/180
```
dwalton76: I’m not sure how many decimal places it goes but you could do “from math import pi” and use “pi” here

ev3dev/control/GyroBalancer.py
```diff
@@ -29,44 +29,48 @@
...
+# Express the above as the angle in rad per motor unit
+RAD_PER_RAW_MOTOR_UNIT = DEG_PER_RAW_MOTOR_UNIT * RAD_PER_DEG
 
-def SetDuty(motorDutyFileHandle, duty, frictionOffset, voltageCompensation):
-    """Function to set the duty cycle of the motors."""
-    # Compensate for nominal voltage and round the input
-    dutyInt = int(round(duty*voltageCompensation))
+# On the EV3, "1% speed" corresponds to 1.7 RPM (if speed
+# control were enabled).
+RPM_PER_PERCENT_SPEED = 1.7
```
dwalton76, line 65: Doesn’t this assume LargeMotor is in use? Maybe make this dynamic based on the motor class that is being used.

ev3dev/control/GyroBalancer.py
```diff
@@ -100,383 +105,379 @@ def __init__(self,
...
-                log.info("-----------------------------------")
-                log.info("STOP")
-                log.info("-----------------------------------")
-
-        # Exit cleanly so that all motors are stopped
-        except (KeyboardInterrupt, Exception) as e:
-            log.exception(e)
-            shutdown()
+        self.motor_left = LargeMotor(left_motor_port)
+        self.motor_right = LargeMotor(right_motor_port)
```
dwalton76, line 133: Allow the use to pass in the motor class to use but have it default to LargeMotor? I can’t think of any balancers that use medium motors but hey maybe we could build one :)

dwalton76 (changes requested):
Would be handy to catch SIGTERM and SIGINT and have those set stop_balance. I have an example of this (it uses ‘signal’) in some of the demo robots....MINDCUBER and probably GRIPPER will do it.

dwalton76:
PS
Supporting motors other than LargeMotor isn’t super important to me, feel free to ignore those comments if you want.

gregcowell:
>Would be handy to catch SIGTERM and SIGINT and have those set stop_balance.

Good idea. I've added the relevant signal handlers. On a related note, I suspect that Brickman sends a SIGKILL when the back button is pressed and we can't have a handler for that. 

>you could do “from math import pi” and use “pi” here

Yep, done.

>Supporting motors other than LargeMotor isn’t super important to me, feel free to ignore those comments if you want.

Thanks @dwalton76. I think I'll leave it with just LargeMotor support at least for now.

WasabiFan:
@dwalton76 Can you take a quick look at this and confirm it's good to go?

dwalton76 (approved):

ev3dev/control/GyroBalancer.py
laurensvalk: Yes. the factor is 2.67 for medium motor.

PR #436, opened 2017-12-08
lego.py color sensor documentation fix

vagoston:
https://github.com/ev3dev/ev3dev/issues/317
Indeed, ambient mode turns the led to blue.


WasabiFan:
Thanks! This does indeed match my recollection of how this sensor works.

PR #443, opened 2018-01-12
ev3dev#1026: Sound.speak does not work with double quotes

amandaoneal:


WasabiFan:
@ddemidov In https://github.com/ev3dev/ev3dev/issues/1026#issuecomment-357100218 I suggested using https://docs.python.org/3.4/library/shlex.html#shlex.quote; see the conversation there. Thoughts?

ddemidov:
@WasabiFan , yes, `shlex` should work, same as in several other places in `sound.py`, e.g. here: https://github.com/ev3dev/ev3dev-lang-python/blob/ee75c5c0489e98257eb7b5bfb07487cdbae6844b/ev3dev/sound.py#L104

ddemidov:
Corection: not in the same way, since `shlex.quote` is also needed here. I think something like
```
cmd_line = '/usr/bin/espeak --stdout {0} "{1}" | /usr/bin/aplay -q'.format(espeak_opts, text)
cmd_args = [shlex.quote(a) for a in shlex.split(cmd_line)]
...
play = Popen(args, stdout=n)
```
(note the absence of `shell=True`) should work and would be consistent with the rest of the code.

ddemidov (changes requested):
Please see if using `shlex` would work here

ddemidov:
Final edit:
```py
cmd_line = ['/usr/bin/espeak', '--stdout'] + shlex.split(espeak_opts) + [shlex.quote(text)]
...
play = Popen(args, stdout=n)
```
seems to be more reliable and actually does work:
```py
>>> espeak_opts='-a -b -c'
>>> text = 'Hello " world'
>>> cmd_line = ['/usr/bin/espeak', '--stdout'] + shlex.split(espeak_opts) + [shlex.quote(text)]
>>> cmd_line
['/usr/bin/espeak', '--stdout', '-a', '-b', '-c', '\'Hello " world\'']
```

amandaoneal:
OK, switched over to shlex and tested on my local EV3 setup. Any verdict on if we wanted shell=True in the Popen calls?

ddemidov (approved):
I think `shell=True` is not needed if we use `shlex.split()`. Otherwise, looks good to me.

dwalton76:
@amandaoneal can you drop the `shell=True` part? Once that is out it looks good to go.  Thanks!

dwalton76:
@amandaoneal there are some conflicts that have to be resolved before we can merge.

amandaoneal:
Looked like the only merge conflict was that the folder 'ev3dev' got renamed 'ev3dev2'... Sigh. Should be back in good shape now.

Issue #422, opened 2017-10-23
"address" argument on LEDs


WasabiFan:
I was just looking at the LED code and realized that we define an "address" argument, but it's never actually used in that class — the only potential consumer I can think of is the device class's generic string conversion method. Can we remove it? I don't think there is any meaning for an "address" in the context of LEDs.

ddemidov:
I wonder if non-system LEDs attached to output ports (as in ev3dev/ev3dev#990), have address attribute. If not, then this is a good idea.

dwalton76:
If we remove it from the list of args we just need to set it to something so the `__str__()` doesn't barf

ddemidov:
Would not overriding `__str__` for the `Led` class work as well?

dwalton76:
yep that would work too

dwalton76:
Actually, `Device.__str__()` is ok if no address is defined
```

    def __str__(self):
        if 'address' in self.kwargs:
            return "%s(%s)" % (self.__class__.__name__, self.kwargs.get('address'))
        else:
            return self.__class__.__name__
    
```

PR #446, opened 2018-01-31
"address" argument on LEDs #422

dwalton76:


WasabiFan (approved):
should  the `__str__` implementation be overridden to print the device name in this case?

dwalton76:
It will print the class name if there is no “address”...not sure what the device name would be for the LEDs?

WasabiFan:
I believe the the led structure is `/sys/class/leds/<device_name>/...` such that the name describes which LED it is. That would be the useful information to provide.

dwalton76:
ok so `self.__class__.__name__` will print something like `<ev3dev.ev3.Leds object at 0xb6710dd0>`...I was expecting just `Leds` here I am not sure where this is coming from.

dwalton76:
ah, is it because `class Leds` does not inherit from `class Device`

dwalton76:
quick summary:
- the Led and Leds classes are off on their own right now in terms of `__str__`.  Will fix these up.
- for any Motor object there will be an address so `Device.__str__()` will print something like `LargeMotor(outA)`
- Sensor objects typically don't have an `address` specified so `Device.__str__()` will print just the class name.  I kinda like this because it is nice and simple and not too long...if we print all of `self._path` for that object that is going to be really long.

dwalton76:
This
```
leds = Leds()
log.info("%s: is the Leds object" % leds)
log.info("%s: is a single Led object" % leds.leds['red_left'])
```

results in
```
2018-01-31 03:45:22,179  INFO: Leds: is the Leds object
2018-01-31 03:45:22,217  INFO: red_left: is a single Led object
```

For a single Led object the desc is almost always going to be set so we get the handy `red_left` output

WasabiFan:
Looks good to me! It might be nice to have the class name in there too but I don't think it makes much of a difference; we should avoid over-complicating things.

Issue #420, opened 2017-10-19
Enhancement: one line import


Joshua-Jurgensmeier:
In #419 I confirmed that the current build no longer supports a single import statement for all classes.

> Personally, I found a single import statement (import ev3dev.ev3 as ev3) to be very convenient in teaching. This way students don't need to understand what an import statement is, they just need to copy-paste that line at beginning of every source file. I understand that, in general, such blanket import statements are bad practice, but it is simpler to teach

Agree, or disagree?

PR #447, opened 2018-01-31
one line import #420

dwalton76:
Example:

    import ev3dev2.auto as ev3

    ts = ev3.TouchSensor(ev3.INPUT_1)

### DIFF
--- a/ev3dev2/auto.py
+++ b/ev3dev2/auto.py
@@ -1,0 +1,51 @@
+from ev3dev2 import *
+
+platform = get_current_platform()
+
+if platform == 'ev3':
+    from ev3dev2._platform.ev3 import INPUT_1, INPUT_2, INPUT_3, INPUT_4
+    from ev3dev2._platform.ev3 import OUTPUT_A, OUTPUT_B, OUTPUT_C, OUTPUT_D
+    from ev3dev2._platform.ev3 import LEDS, LED_GROUPS, LED_COLORS
+
+elif platform == 'evb':
+    from ev3dev2._platform.evb import INPUT_1, INPUT_2, INPUT_3, INPUT_4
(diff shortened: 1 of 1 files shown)


dwalton76:
I am not married to this approach, this is just the first way to solve this that popped in my head.

dwalton76:
@WasabiFan do you know of a way to make the CI tests run again? It failed for some non-obvious reason.

ddemidov:
It looks like the problem with tests was on Travis-CI side. I've restarted the build and it passed this time.

ddemidov (approved):
I like this. Also, it is optional, so if someone does not like the approach (for example, because it loads modules that are unnecessary for the current robot), there is always an option to do it manually instead.

dwalton76:
Thanks Denis

yeah seems like this makes it about as easy as it can get but with the penalty of slower startup times due to import every module.  Will leave this open for a day so folks have a chance to look at it and then will merge.

WasabiFan:
This seems reasonable to me. We need to figure out whether we want to recommend the "auto" module or individual imports in e.g. the README (in my WIP PR it uses the individual ones). Also, one of the first things on my to-do list is properly supporting Micropython; I'll need to modify this to exclude some modules in that environment, I think.

dwalton76:
My preference would be to steer folks towards explicitly importing only what they need, that is more of standard python best practice and it makes it more obvious as to where to look to find the code you are using.  

Something I need to test is how long importing everything takes with and without `from ev3dev2.control.webserver import *`.  My memory is that the `from http.server import BaseHTTPRequestHandler, HTTPServer` that webserver.py does is very slow, we may need to exclude `from ev3dev2.control.webserver import *` from `auto.py`.  The majority of folks won't use this anyway.

WasabiFan:
Agreed. In fact, I'd say the same about GyroBalancer and rc_tank.

dwalton76:
k I am going to drop the `from ev3dev2.control` imports and then merge

PR #453, opened 2018-04-10
Brickpi3 support

dwalton76:
This only gets us part of the way to really supporting brickpi3 in ev3dev-lang-python.  Some driver work needs to be done in the kernel for brickpi3 to be event-driven with the motor `state` file.


dwalton76:
ugh, that is a mess, will close this and open another PR

Issue #456, opened 2018-04-16
Importing ev3dev.fonts takes a lot of time


ddemidov:
See https://github.com/ev3dev/ev3dev/issues/1022#issuecomment-381445981

ddemidov:
We could probably get rid of `pkg_resources` by using `__file__` variable inside `ev3dev.fonts.__init__` to get the path to the fonts directory. Then we could just use `glob.glob()` as a replacement for `package_listdir`, and `os.path.join()` for `package_filename`.

ndward:
40s just to load some file names. Wow. I tried 
```
#!/usr/bin/env python3
from pkg_resources import package_listdir, package_filename
```
and after 48s (39s adjusted for download etc time) I got
`ImportError: cannot import name 'package_listdir'`

I also tried
```
#!/usr/bin/env python3
from pkg_resources import resource_listdir, resource_filename
```
and that completed in about 50s, so no time saved.

I believe the font files we are talking about are at usr\lib\python3\dist-packages\ev3dev\fonts. That folder contains 402 items (1.34MB), of which 200 .pbm files and 200 .pil files. It also contains a file `_int_.py` which begins:
```
import pkg_resources
import os.path
from PIL import ImageFont

def available():
    """
    Returns list of available font names.
    """
    names = []
    for f in pkg_resources.resource_listdir('ev3dev.fonts', ''):
        name, ext = os.path.splitext(os.path.basename(f))
        if ext == '.pil':
            names.append(name)
    return sorted(names)
```
Is that available() function relevant or useful?

If there is no other way of speeding up the loading of file names then maybe one solution would be to split the fonts into multiple folders so that only one small folder then has to have its file names loaded? Or maybe we don't need 200 different fonts? Or maybe we don't need both the .pil and the .pbm files?

ndward:
Using Stretch, it seems to take about 45s for this command to run
`import ev3dev.fonts as fonts`

@ddemidov says that this command does not import any fonts but does this instead:
```
import pkg_resources # 40s
import os.path # 1s
from PIL import ImageFont # 3s
```
I have added comments above to indicate how long each import seems to take. The only import that takes a long time is
`import pkg_resources #40s`

@ddemidov says
> pkg_resources is used to get file names for the font files packaged with ev3dev.

[...]

PR #458, opened 2018-04-22
Replace pkg_resources with __file__/glob combination

ddemidov:
This should reduce time to import `ev3dev.fonts`. See #456.
Would be nice if anyone could confirm this.

### DIFF
--- a/ev3dev2/fonts/__init__.py
+++ b/ev3dev2/fonts/__init__.py
@@ -1,16 +1,14 @@
-import pkg_resources
 import os.path
+from glob import glob
 from PIL import ImageFont
 
 def available():
     """
     Returns list of available font names.
     """
-    names = []
-    for f in pkg_resources.resource_listdir('ev3dev.fonts', ''):
-        name, ext = os.path.splitext(os.path.basename(f))
-        if ext == '.pil':
-            names.append(name)
+    font_dir = os.path.dirname(__file__)
+    names = [os.path.basename(os.path.splitext(f)[0])
+            for f in glob(os.path.join(font_dir, '*.pil'))]
     return sorted(names)
 
 def load(name):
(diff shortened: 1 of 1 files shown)


ddemidov:
This works on my machine, but I did not test the actual performance impact on the ev3.

ndward:
Is this issue referring to the latest Stretch image (22 April 2018)? I'm not able to get a connection from my PC to the EV3 with that image. The EV3 says it is connected but neither VS Code nor SFTP Net Drive can connect to it. Just me?

ddemidov:
This implementation should work both for the jessie and stretch images. I hope you would see an improvement on whatever system you used here: https://github.com/ev3dev/ev3dev/issues/1022#issuecomment-381672410.

ddemidov:
Just checked, and this works on the latest stretch image. The following script was used for testing:
```py
from ev3dev2.fonts import *
```
Before the patch:
```sh
time python3 tf.py

real	0m32.798s
user	0m24.747s
sys	0m0.603s
```
After the patch:
```sh
time python3 tf.py

real	0m5.543s
user	0m4.168s
sys	0m0.266s
```

ensonic:
In jessie ev3dev-lang-.python has the same `/usr/lib/python3/dist-packages/ev3dev/fonts/__init__.py`. I've applied the same changes, but did not see the speedup. Do we have less fonts?
```
$ cat tf.py
from ev3dev.fonts import *
$ time python3 tf.py 

real	0m14.355s
user	0m12.070s
sys	0m0.500s
$ sudo mv new.py /usr/lib/python3/dist-packages/ev3dev/fonts/__init__.py
$ time python3 tf.py 

real	0m13.907s
user	0m11.960s
sys	0m0.500s
$ ls -1  /usr/lib/python3/dist-packages/ev3dev/fonts/ | wc -l
402
```

FInally why do we bother to build the 'pbm_file' in load() if we don't use it ever?

ddemidov:
> Do we have less fonts?

We package the fonts ourselves, so it should be the same both for stretch and jessie.

> I've applied the same changes, but did not see the speedup

Can you add some prints to the new `__init__.py` to make sure you are actually using it?

> FInally why do we bother to build the 'pbm_file' in load() if we don't use it ever?

I don't remember why exactly, but I think it was used implicitly. Can you try to comment that out and see if the fonts still work?

PR #462, opened 2018-04-30
Update display for 4.14 kernel

### LINKED ISSUE #455: new display driver in kernel 4.14.x
I just released a new ev3dev kernel on ev3dev-stretch that updates from 4.9.x to 4.14.x. This comes with a new display driver for the EV3 from the mainline kernel.

The new driver still provides a fbdev device (`/dev/fb0`) however, the memory layout is different. Because of limitations in the mainline kernel, there is no monochome. Instead it uses a 32-bit XRGB pixel format.

This causes an error with ev3dev-lang-python. See https://github.com/ev3dev/ev3dev/issues/1022#issuecomment-381444425

ddemidov:
Tested on the latest ([2018-04-22](https://oss.jfrog.org/list/oss-snapshot-local/org/ev3dev/brickstrap/2018-04-22)) stretch image. It works, and is indeed able to produce grayscale images:
```py
from time import sleep
from ev3dev2.auto import Display, Sound
from ev3dev2 import fonts

Sound().beep()

d = Display()
d.clear()
d.draw.rectangle((10,10,60,20,), fill='black')
d.draw.rectangle((10,20,60,30,), fill='gray')
d.draw.text((10,40), 'Hello1', font=fonts.load('luBS14'), fill='black')
d.draw.text((10,50), 'Hello2', font=fonts.load('luBS14'), fill='gray')
d.draw.text((10,60), 'Hello3', font=fonts.load('luBS14'))
d.update()
sleep(5)
```
test

Fixes #455

### DIFF
--- a/ev3dev2/display.py
+++ b/ev3dev2/display.py
@@ -198,14 +198,12 @@ class Display(FbMem):
 
         self.platform = get_current_platform()
 
-        if self.platform == "ev3" and self.var_info.bits_per_pixel == 1:
-            # Pre 4.14 kernel
+        if self.var_info.bits_per_pixel == 1:
             im_type = "1"
-        elif self.platform == "ev3" and self.var_info.bits_per_pixel == 32:
-            # Post 4.14 kernel
-            im_type = "L"
-        elif self.platform == "pistorms" and self.var_info.bits_per_pixel == 16:
+        elif self.var_info.bits_per_pixel == 16:
             im_type = "RGB"
+        elif self.platform == "ev3" and self.var_info.bits_per_pixel == 32:
+            im_type = "L"
         else:
             raise Exception("Not supported")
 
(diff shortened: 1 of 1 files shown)


ev3dev2/display.py
```diff
@@ -195,8 +195,17 @@ class Display(FbMem):
     def __init__(self, desc='Display'):
         FbMem.__init__(self)
 
+        if self.var_info.bits_per_pixel == 1:
+            im_type = "1"
+        elif self.var_info.bits_per_pixel == 16:
+            im_type = "RGB"
+        elif self.var_info.bits_per_pixel == 32:
```
dlech: Hmm... we can probably get away with this since there aren't currently any other XRGB displays on ev3dev devices, but if there ever are, then they won't be usable in color.

ddemidov: Do you suggest to use platform here instead of `bits_per_pixel`?

dlech: I suggest using both.

```
elif self.var_info.bits_per_pixel == 32:
    if platform == EV3:
        grayscale
    else:
        really 32-bit
```

ddemidov: 69accdf does something similar. Tested on the EV3.

ev3dev2/display.py
```diff
@@ -195,8 +196,21 @@ class Display(FbMem):
...
 
+        self.platform = get_current_platform()
+
+        if self.platform == "ev3" and self.var_info.bits_per_pixel == 1:
+            # Pre 4.14 kernel
+            im_type = "1"
+        elif self.platform == "ev3" and self.var_info.bits_per_pixel == 32:
+            # Post 4.14 kernel
+            im_type = "L"
+        elif self.platform == "pistorms" and self.var_info.bits_per_pixel == 16:
```
dlech: Please only check for the special case of platform on EV3. We use 16bpp on at least 3 different devices.

dlech:
Why is the background yellow?

ddemidov:
Probably just a side effect of `fbcat` followed by `convert` to png. The ev3 screen looks normal, but it also could be because 'x' in 'xrgb' is in the wrong position. Since we are now using same byte for all three components, we could  also put it into 'x', since it's not used anyway.

dlech:
I have a hunch that it is because of this line:

    return ((v << 24) | (v << 16) | (v << 8))

this fills XRG and leaves B as 0, which would make yellow.

ddemidov:
b18428b results in 
test

WasabiFan:
I saw a comment elsewhere that this or a related fix was released already; what's the status here?

ddemidov:
This is waiting for an approval (I think it can be merged).

dlech (approved):
LGTM

Issue #466, opened 2018-05-05
branch confusion


dlech:
1. There appears to be commits for ev3dev-stretch in the `jessie` branch
2. What is the purpose of the `ev3dev2-namespace` branch? It seems that the `stretch` branch contains the `ev3dev2` namespace already.
3. There is no branch for the 1.x releases (tags).

dlech:
Re: item 3, the reason I am asking about this is because I am going to go ahead do a v1.2.0 release that fixes the kernel breakage in ev3dev-stretch since it doesn't appear that 2.x is ready for release yet.

dwalton76:
Item 1: can you point me at some of them?

Item 2: I am pretty sure ev3dev2-namespace can be deleted but I was waiting for confirmation from @wasabifan, see #449.

dlech:
> Item 1: can you point me at some of them?

* https://github.com/ev3dev/ev3dev-lang-python/commit/57e1991ace2d9011d073cb28f042c57473de03ff
* https://github.com/ev3dev/ev3dev-lang-python/commit/078b98a24b9c8fdc2b49dba86279df14df8e8769

dwalton76:
I don’t have a good answer...looks like we made a mistake :(

dlech:
So how should we fix it?

WasabiFan:
I've been out of the loop; it looks like the only problematic changes in those two commits are the updates to the LED and button strings, is that correct? Assuming this wasn't already done, does anyone see a reason not to revert those two particular parts of those changes in a new commit? Also, why hasn't anyone encountered an issue with that yet (or have they, and I'm just oblivious)?

ddemidov:
> Assuming this wasn't already done, does anyone see a reason not to revert those two particular parts of those changes in a new commit?

:+1:

> Also, why hasn't anyone encountered an issue with that yet (or have they, and I'm just oblivious)?

Probably because the changes to jessie were made after the package was released?

WasabiFan:
I believe this has been fixed; please reopen if it's still an issue.

PR #470, opened 2018-05-21
Remove Stretch-only changes accidentally introduced to Jessie branch

WasabiFan:
#466

### DIFF
--- a/ev3dev/ev3.py
+++ b/ev3dev/ev3.py
@@ -47,10 +47,10 @@ class Leds(object):
 
 # ~autogen led-colors platforms.ev3.led>currentClass
 
-    red_left = Led(name_pattern='led0:red:brick-status')
-    red_right = Led(name_pattern='led1:red:brick-status')
-    green_left = Led(name_pattern='led0:green:brick-status')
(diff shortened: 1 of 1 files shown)


ddemidov:
There is some problem with `ev3dev-lang-fake-sys` submodule. The `jessie` branch points to @rhempel's version of it, and the `stretch` branch points to mine. I think both should point to https://github.com/ev3dev/ev3dev-lang-fake-sys. The error in travis seems to be caused by this.

ddemidov:
Correction, the jessie branch correctly points to https://github.com/ev3dev/ev3dev-lang-fake-sys; this PR tries to update it?

And while stretch branch does point to mine version of ev3dev-lang-fake-sys, it is correctly redirected to ev3dev one (but we still need to update the address I think).

WasabiFan:
It looks like the current Jessie branch points to rhempel/ev3dev-lang-fake-sys. I don't understand why exactly it failed, but I'll see if I can switch it to point to the ev3dev instance and go from there.

WasabiFan:
This needs to be rebased after #477 is merged.

ddemidov (approved):

PR #480, opened 2018-06-29
Accept unit classes in motor groups

### LINKED ISSUE #474: Accept SpeedInteger classes in motor groups
`MotoveSteering` and friends, as well as (probably) `MotorSet`, don't accept the new `SpeedInteger` classes right now. Adding that should be as simple as calling the percentage conversion method before using the arguments.

WasabiFan:
I did some things... primarily enabling unit classes in motor groups, plus some other fixes. Also created tests for these things because I don't trust my programming abilities (nor Python).

Targeting this PR on the docs branch because it has other changes to these same files which I don't want to deal with as merge conflicts.

Fixes #474


WasabiFan:
I don't know why that CI error is happening... clearly the files aren't opened as r/w but I don't see why that is different there.

ddemidov:
[It looks like](https://travis-ci.org/ev3dev/ev3dev-lang-python/builds/398078593#L522) the first failure is at [api_tests.py:144](https://github.com/ev3dev/ev3dev-lang-python/blob/83e70660aad3cdb3190d9cb38ed16342df5303d2/tests/api_tests.py#L144), and [the specific error](https://travis-ci.org/ev3dev/ev3dev-lang-python/builds/398078593#L538) is `io.UnsupportedOperation: File not open for writing`.

ddemidov:
Could it be the file permissions are read-only in fake-sys repo?

EDIT: nope, those seem to be ok. I am out of ideas for now.

WasabiFan:
The CI issue should be fixed by #481.

ddemidov:
> The CI issue should be fixed by #481.

Can you rebase this on stretch then?

WasabiFan:
I've rebased both PRs. I think it probably makes the most sense to merge #411 before this one, just to keep the commits somewhat logical. That PR is complete aside from the fact that I'd like to add some info about how to use the new features; I'd appreciate a review of the content over there as well.

WasabiFan:
@ddemidov Could you check over this and #483? Once I merge these two PRs I plan to test all the basic functions on my EV3 and then publish a release -- my intent is to do that within the next few days if at all possible.

ddemidov (commented):
I was out of the loop on this set of changes, and I did not test if it works on the ev3, so I think I'll just let you decide if this is ready to merge.

Some specific things that caught my eye is old-style formatting (`'%s' % (v)` instead of `'{}'.format(v)`), but I am not sure how strict are we about these kinds of things.

Also, `print` [is used](https://github.com/ev3dev/ev3dev-lang-python/pull/480/files#diff-13751cf3fd528f6a548100cbf55db214L42) to spew warnings as opposed to `logging` module [elswhere](https://github.com/ev3dev/ev3dev-lang-python/blob/ev3dev-stretch/ev3dev2/motor.py#L878). We should probably decide on one of those.

ddemidov (approved):
This looks good to me (I assume @dlech has already checked debian-related changes).

PR #481, opened 2018-07-12
Fix CI tests

WasabiFan:
First commit adds a failing test (we didn't write to properties before); second commit fixes it 💪😎. I don't entirely understand why the files had different permissions on the CI server from what I see when I clone in Ubuntu locally, but whatever the issue is, checking for user permissions seems to work. I'll validate this on an EV3 soon to confirm I didn't cause any regressions.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,7 @@ install:
 - pip install Pillow
 - pip install evdev
 script:
+- chmod -R g+rw ./tests/fake-sys/devices/**/*
 - ./tests/api_tests.py
 deploy:
   provider: pypi
--- a/tests/api_tests.py
+++ b/tests/api_tests.py
@@ -86,5 +86,15 @@ class TestAPI(unittest.TestCase):
         self.assertEqual(s.address,         'in1')
         self.assertEqual(s.value(0),        16)
 
+    def test_medium_motor_write(self):
+        clean_arena()
+        populate_arena({'medium_motor' : [0, 'outA']})
+
+        m = MediumMotor()
+
+        self.assertEqual(m.speed_sp, 0)
+        m.speed_sp = 500
+        self.assertEqual(m.speed_sp, 500)
+
 if __name__ == "__main__":
     unittest.main()


ev3dev2/__init__.py
```diff
@@ -199,8 +199,8 @@ def __repr__(self):
     def _attribute_file_open(self, name):
         path = os.path.join(self._path, name)
         mode = stat.S_IMODE(os.stat(path)[stat.ST_MODE])
-        r_ok = mode & stat.S_IRGRP
-        w_ok = mode & stat.S_IWGRP
+        r_ok = mode & stat.S_IRUSR
+        w_ok = mode & stat.S_IWUSR
```
ddemidov: I am not sure now, but I think group permission are checked for a reason. Attributes are root-owned, and to read some of them, we add users to a specific groups. See f49158b93aef7eb312de7da5f45a7606384e1fbe and #84

ddemidov: Can we alter group permissions on the CI server before running the tests?

WasabiFan: Hmph, I was wondering about that. We probably can add such logic to the `fake-sys` script as long as we know what to add... I assume it's a `chmod` call with these flags added but I don't fully understand the Unix permissions model so I'd have to test it.

ddemidov: May be something like `chown -R $USER:$USER` to change group to the current user and/or `chmod -R g+rw` to allow group member to read/write?

ddemidov: We use copy the devices from [devices](https://github.com/ev3dev/ev3dev-lang-fake-sys/tree/master/devices), and python's [`shutil.copytree`](https://github.com/ev3dev/ev3dev-lang-fake-sys/blob/master/populate_arena.py#L33) should copy permissions, so may be adding `chmod/chown` calls to the travis script would be enough:

https://github.com/ev3dev/ev3dev-lang-python/blob/jessie/.travis.yml#L8

WasabiFan:
I can't reproduce this error locally, despite using the official docker image and replicating the build commands used in the real environment. Nonetheless, there certainly is a difference.


Locally:
```
$ stat tests/fake-sys/arena/tacho-motor/motor0/speed_sp
  File: ‘tests/fake-sys/arena/tacho-motor/motor0/speed_sp’
  Size: 2               Blocks: 8          IO Block: 4096   regular file
Device: 801h/2049d      Inode: 2766152     Links: 1
Access: (0664/-rw-rw-r--)  Uid: ( 2000/  travis)   Gid: ( 2000/  travis)
Access: 2018-07-12 05:43:02.784530600 +0000
Modify: 2018-07-12 05:42:46.762545800 +0000
Change: 2018-07-12 05:46:25.805545600 +0000
 Birth: -
```

On their agent:
```
$ stat ./tests/fake-sys/arena/tacho-motor/motor0/speed_sp
  File: './tests/fake-sys/arena/tacho-motor/motor0/speed_sp'
  Size: 2         	Blocks: 8          IO Block: 4096   regular file
Device: fd07h/64775d	Inode: 21730405    Links: 1
Access: (0644/-rw-r--r--)  Uid: ( 2000/  travis)   Gid: ( 2000/  travis)
Access: 2018-07-12 05:53:49.629314075 +0000
Modify: 2018-07-12 05:53:44.725738838 +0000
Change: 2018-07-12 05:53:49.749303731 +0000
 Birth: -
```

Note the bit of interest -- group write -- isn't set in the real env. Going to test out some things now.

Build in question: https://travis-ci.org/ev3dev/ev3dev-lang-python/builds/402969142

WasabiFan:
Local:
```
$ stat tests/fake-sys/devices/medium_motor/speed_sp
  File: ‘tests/fake-sys/devices/medium_motor/speed_sp’
  Size: 2               Blocks: 8          IO Block: 4096   regular file
Device: 801h/2049d      Inode: 2766092     Links: 1
Access: (0664/-rw-rw-r--)  Uid: ( 2000/  travis)   Gid: ( 2000/  travis)
Access: 2018-07-12 05:43:02.764530600 +0000
Modify: 2018-07-12 05:42:46.762545800 +0000
Change: 2018-07-12 05:42:46.762545800 +0000
 Birth: -
```

Remote:
```
$ stat tests/fake-sys/devices/medium_motor/speed_sp
  File: 'tests/fake-sys/devices/medium_motor/speed_sp'
  Size: 2         	Blocks: 8          IO Block: 4096   regular file
Device: fd09h/64777d	Inode: 34657809    Links: 1
Access: (0644/-rw-r--r--)  Uid: ( 2000/  travis)   Gid: ( 2000/  travis)
Access: 2018-07-12 06:01:22.772310467 +0000
Modify: 2018-07-12 06:01:17.936466389 +0000
Change: 2018-07-12 06:01:17.936466389 +0000
 Birth: -
```

I'm inclined to believe that setting that bit in the source directory both will work and is the correct approach. Will test.

WasabiFan:
> The command "chmod -R g+rw ./tests/fake-sys/devices/**/*" exited with 0.
> The command "./tests/api_tests.py" exited with 0.

🎉 I'll update the PR tomorrow (my time; I need some sleep).

```yaml
script:
- chmod -R g+rw ./tests/fake-sys/devices/**/*
- ./tests/api_tests.py
```

WasabiFan:
Updated with the new fix

ddemidov (approved):

PR #482, opened 2018-07-17
debian: update control file for ev3dev-lang-python 2.x

dlech:
This changes the name of the debian package so that the 2.x version can be installed at the same time as the 1.x version

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -1,13 +1,13 @@
-Source: python-ev3dev
+Source: python-ev3dev2
 Maintainer: Ralph Hempel <rhempel@hempeldesigngroup.com>
 Section: python
 Priority: optional
-Standards-Version: 3.9.5
+Standards-Version: 3.9.8
 Build-Depends: python3-setuptools (>= 0.6b3), python3-all (>= 3.4), debhelper (>= 9), dh-python, python3-pillow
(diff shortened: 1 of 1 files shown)


dlech:
Maybe update the Maintainer: field too? :smile:

ddemidov (approved):

WasabiFan:
How is it that the package name is `python-ev3dev` yet IIRC it's installed as `python3-ev3dev`?

dlech:
the first package listed is the source package. The second package is the binary package, which is `python3-ev3dev2`. The binary package is the one that gets installed.

WasabiFan:
Got it. What's the use case for source packages for Python libraries?

WasabiFan (approved):

WasabiFan:
Thanks for the PR!

Re: maintainer field, can it be a list of multiple people?

ddemidov:
> can it be a list of multiple people?

or a team name?

dlech:
> What's the use case for source packages for Python libraries?

All debian packages have a source package, because, you know, open source :wink:.

You can, for example, download this with `apt source python-ev3dev`.

> Re: maintainer field, can it be a list of multiple people?

Nope.

> or a team name?

Sure, as long as there is a valid email address to go with it.

PR #483, opened 2018-07-18
Port fixes from 1.2.0 tag to stretch release branch; update EV3 port names

### LINKED ISSUE #476: Update to new Stretch port names
As mentioned in https://github.com/ev3dev/ev3dev-lang-python/issues/339, motor port names have changed in Stretch. We're going to need to update the library and test infrastructure.

WasabiFan:
These changes originally got lost; I tried to port the platform changes properly but I don't have the devices to test it. @dlech could you look over this and verify the platform config?

Result of https://github.com/ev3dev/ev3dev/issues/1130

Fixes #476


ev3dev2/_platform/pistorms.py
```diff
@@ -4,28 +4,28 @@
...
 INPUT_2 = 'pistorms:BBS2'
-INPUT_3 = 'pistorms:BAS2'
-INPUT_4 = 'pistorms:BAS1'
+INPUT_3 = 'pistorms:BBS1'
+INPUT_4 = 'pistorms:BBS2'
 
 
-BUTTONS_FILENAME = None
+BUTTONS_FILENAME = '/dev/input/by-path/platform-3f804000.i2c-event'
 EVDEV_DEVICE_NAME = None
```
WasabiFan: I'm not sure what this name should be; suggestions welcome.

ev3dev2/_platform/pistorms.py
```diff
@@ -4,28 +4,28 @@
...
-OUTPUT_C = 'pistorms:BAM2'
-OUTPUT_D = 'pistorms:BAM1'
+OUTPUT_A = 'pistorms:BAM1'
+OUTPUT_B = 'pistorms:BAM2'
+OUTPUT_C = 'pistorms:BBM1'
+OUTPUT_D = 'pistorms:BBM2'
 
-INPUT_1 = 'pistorms:BBS1'
+INPUT_1 = 'pistorms:BAS1'
 INPUT_2 = 'pistorms:BBS2'
```
dlech:
```python
INPUT_2 = 'pistorms:BAS2'
```

ev3dev2/_platform/pistorms.py
dlech: I assume this is meant to match the event device name, which would be `PiStorms` in this case.

https://github.com/ev3dev/lego-linux-drivers/blob/ev3dev-stretch/pistorms/pistorms_input.c#L205

debian/changelog
```diff
@@ -1,3 +1,30 @@
+python-ev3dev (1.2.0) stable; urgency=medium
```
dlech: Since we are changing the package name, it would probably be easier to just start the changelog over rather than try to list all of the changes from 1.x to 2.x.

debian/gbp.conf
```diff
@@ -0,0 +1,3 @@
+[DEFAULT]
+debian-branch=ev3dev-stretch
+debian-tag=ev3dev-stretch/%(version)s
```
dlech, line 3: this will require you to change the branch name from `stretch` to `ev3dev-stretch` in order to build the package using `gbp`.

Since we will have both a 1.x and 2.x in ev3dev-strech, I was thinking of suggesting that this repository have `ev3dev-stretch-1.x` and `ev3dev-stretch-2.x` branches, in which case this would need to be changed to `debian-branch=ev3dev-stretch-2.x` (tag line doesn't need to be changed - and probably best if it isn't changed).

WasabiFan: I changed the branch to `stretch` to match the repo. It might make more sense to switch the branch names to `ev3dev-{stretch, jessie}` to match other repos, but I'm don't particularly care.

dlech: Making them all the same (`ev3dev-stretch`) would be my inclination, but you have the parallel 1.x and 2.x branches here, so calling this branch `stretch-2.x` might make more sense, perhaps?

WasabiFan: Would you be opposed to `ev3dev-stretch-1.x` and `ev3dev-stretch`? Or perhaps `ev3dev-stretch-legacy` for the 1.x series? My hope is that I can avoid typing `-2.x` if it isn't necessary.

dlech: First option sounds fine. (If you make `ev3dev-stretch` the default branch on GitHub, you should never have to type it in anyways 😉 .)

WasabiFan: > First option sounds fine. 

I just pushed the new branches and deleted the old `stretch[-1.x]` branches. I don't have admin access on this repo though; could you switch the default branch to `ev3dev-jessie`?

> If you make ev3dev-stretch the default branch on GitHub, you should never have to type it in anyways 😉

We'll do that once we release the first stretch version, I think. If everyone's running code from the jessie branch I'd rather make that the default -- once there's Stretch code out in the wild that changes.

WasabiFan:
I think I fixed everything that was pointed out in review so I'm going to merge this. If there are any issues we can fix it in a future PR (either one of us notices it or people report it).

PR #487, opened 2018-07-19
Update setup.py

dlech:
This updates setup.py with a new package name (`python-ev3dev2`), adds 2.x to the description, change the author/email to a shared team name and fixes the GitHub URL.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -3,13 +3,13 @@ from git_version import git_version
 
 
 setup(
-    name='python-ev3dev',
+    name='python-ev3dev2',
     version=git_version(),
-    description='Python language bindings for ev3dev',
-    author='Ralph Hempel',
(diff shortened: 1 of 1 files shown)


WasabiFan:
This looks good to me :+1:; do you also want to update `packages` now or would you prefer that I do it?

dlech:
`packages` has already been updated to `ev3dev2`. Or am I missing something?

WasabiFan:
I suppose that depends on what the list is supposed to represent; if it's only modules in their own folder, I think it's up-to-date, but if it's all modules from the user's perspective it should also include `display`, `button`, `motor`, etc.

dlech:
Yeah, I think "packages" translates more-or-less to folders with an `__init__.py`.

WasabiFan (approved):

PR #490, opened 2018-07-29
Sync README, docs and code behavior

WasabiFan:
Also, minor improvement to LED parameter options.

The docs in the `Sound` class were quite badly formatted after going through Sphinx so I updated all of those methods to use the standard Sphinx/reST style which will be rendered properly. It would be nice to update other areas to do the same.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -22,10 +22,9 @@ Getting Started
 
 This library runs on ev3dev_. Before continuing, make sure that you have set up
 your EV3 or other ev3dev device as explained in the `ev3dev Getting Started guide`_.
-Make sure that you have a kernel version that includes ``-10-ev3dev`` or higher (a
-larger number). You can check the kernel version by selecting "About" in Brickman
-and scrolling down to the "kernel version". If you don't have a compatible version,
-`upgrade the kernel before continuing`_.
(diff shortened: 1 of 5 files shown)


README.rst
```diff
@@ -144,7 +158,7 @@ If you want to make your robot speak, you can use the ``Sound.speak`` method:
   from ev3dev2.sound import Sound
 
   sound = Sound()
-  sound.speak('Welcome to the E V 3 dev project!').wait()
+  sound.speak('Welcome to the E V 3 dev project!')
```
ddemidov, line 161: I started to comment asking why did we remove this, when we still are returning `Popen` objects from `Sound` methods, then noticed we mostly don't anymore. The exception is `beep`, which still needs `.wait()` to make it synchronous. Should it be changed as well?

ddemidov (approved):

README.rst
WasabiFan: Good point. I'll modify `beep` to match the behavior of the other methods.

WasabiFan: I've updated `beep` and its derivative `tone`, and in the process fixed the docs example because these methods aren't static anymore.

PR #500, opened 2018-08-30
Improve speed integer and rename "speed_pct" to "speed"

### LINKED ISSUE #479: Should SpeedInteger actually be floating-point?
We have these new `SpeedInteger` classes. But... as it turns out, it might make more sense to have them be `SpeedFloat`s (or, more likely, a better name like `SpeedMeasure` or simply `Speed`). My thinking is: given that we're converting from whatever input units are provided to native units (eventually), we should minimize intermediate truncations to enable the granularity that ticks-per-second provides. This should just be a matter of changing the name, inheritance and documentation mentions.

WasabiFan:
- Rename `SpeedInteger` to `SpeedMeasure`
- Use an instance field rather than inheriting from `int`
- Drop `_pct` name suffix (was confusing users and doesn't have meaning much anymore)
- Use native motor units internally rather than percentages
- Update tests to be tougher on rounding (now that we only round once at the end)

Note that there is a slight breaking change here: anyone referencing the "speed_pct" argument by name will get an error. Our last release was a beta, so we aren't breaking semver by doing this (although I want to minimize breaking changes as much as possible).

Fixes #479

### DIFF
--- a/docs/motors.rst
+++ b/docs/motors.rst
@@ -10,7 +10,7 @@ Motor classes
 Units
 -----
 
-Most methods which run motors with accept a ``speed`` or ``speed_pct`` argument. While this can be provided as an integer which will be interpreted as a percentage of max speed, you can also specify an instance of any of the following classes, each of which represents a different unit system:
+Most methods which run motors with accept a ``speed`` argument. While this can be provided as an integer which will be interpreted as a percentage of max speed, you can also specify an instance of any of the following classes, each of which represents a different unit system:
 
 .. autoclass:: SpeedInteger
 .. autoclass:: SpeedPercent
--- a/docs/motors.rst
+++ b/docs/motors.rst
@@ -10,9 +10,9 @@ Motor classes
 Units
 -----
 
-Most methods which run motors with accept a ``speed`` argument. While this can be provided as an integer which will be interpreted as a percentage of max speed, you can also specify an instance of any of the following classes, each of which represents a different unit system:
+Most methods which run motors will accept a ``speed`` argument. While this can be provided as an integer which will be interpreted as a percentage of max speed, you can also specify an instance of any of the following classes, each of which represents a different unit system:
 
-.. autoclass:: SpeedInteger
+.. autoclass:: SpeedValue
 .. autoclass:: SpeedPercent
 .. autoclass:: SpeedNativeUnits
 .. autoclass:: SpeedRPS
--- a/ev3dev2/motor.py
+++ b/ev3dev2/motor.py
@@ -878,9 +878,9 @@ class Motor(Device):
         assert rotations >= 0, "rotations is {}, must be >= 0".format(rotations)
 
         if speed > 0:
-            self.position_sp = self.position + int(rotations * self.count_per_rot)
+            self.position_sp = self.position + int(round(rotations * self.count_per_rot))
         else:
-            self.position_sp = self.position - int(rotations * self.count_per_rot)
+            self.position_sp = self.position - int(round(rotations * self.count_per_rot))
 
     def _set_position_degrees(self, speed, degrees):
 
@@ -1758,8 +1758,8 @@ class MoveTank(MotorSet):
             "Either left_speed or right_speed must be non-zero"
         
         return (
-            int(left_speed),
-            int(right_speed)
+            left_speed,
+            right_speed
         )
 
     def on_for_rotations(self, left_speed, right_speed, rotations, brake=True, block=True):
@@ -1785,10 +1785,10 @@ class MoveTank(MotorSet):
             right_rotations = rotations
 
         # Set all parameters
-        self.left_motor.speed_sp = left_speed_native_units
+        self.left_motor.speed_sp = int(round(left_speed_native_units))
         self.left_motor._set_position_rotations(left_speed_native_units, left_rotations)
         self.left_motor._set_brake(brake)
-        self.right_motor.speed_sp = right_speed_native_units
+        self.right_motor.speed_sp = int(round(right_speed_native_units))
         self.right_motor._set_position_rotations(right_speed_native_units, right_rotations)
         self.right_motor._set_brake(brake)
 
@@ -1819,10 +1819,10 @@ class MoveTank(MotorSet):
             right_degrees = degrees
 
         # Set all parameters
-        self.left_motor.speed_sp = left_speed_native_units
+        self.left_motor.speed_sp = int(round(left_speed_native_units))
         self.left_motor._set_position_degrees(left_speed_native_units, left_degrees)
         self.left_motor._set_brake(brake)
-        self.right_motor.speed_sp = right_speed_native_units
+        self.right_motor.speed_sp = int(round(right_speed_native_units))
         self.right_motor._set_position_degrees(right_speed_native_units, right_degrees)
         self.right_motor._set_brake(brake)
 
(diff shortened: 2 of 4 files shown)


docs/motors.rst
```diff
@@ -10,7 +10,7 @@ Motor classes
 Units
 -----
 
-Most methods which run motors with accept a ``speed`` or ``speed_pct`` argument. While this can be provided as an integer which will be interpreted as a percentage of max speed, you can also specify an instance of any of the following classes, each of which represents a different unit system:
+Most methods which run motors with accept a ``speed`` argument. While this can be provided as an integer which will be interpreted as a percentage of max speed, you can also specify an instance of any of the following classes, each of which represents a different unit system:
```
ddemidov: > Most methods which run motors **with** accept

will?

ev3dev2/motor.py
```diff
@@ -67,105 +67,126 @@
     raise Exception("Unsupported platform '%s'" % platform)
 
 
-class SpeedInteger(int):
+class SpeedMeasure():
```
ddemidov: Not sure about 'SpeedMeasure' name. How about 'SpeedValue' or 'SpeedUnit'?

ev3dev2/motor.py
```diff
@@ -67,105 +67,126 @@
...
         """
-        Return the motor speed percentage represented by this SpeedPercent
+        Return this SpeedPercent in native motor units
         """
-        return self
+        return self.percent / 100 * motor.max_speed
 
 
-class SpeedNativeUnits(SpeedInteger):
+class SpeedNativeUnits(SpeedMeasure):
```
ddemidov: Just a thought: there are `SpeedRPS`, `SpeedDPS`, etc. Why not call this `SpeedCPS` for the sake of uniformity? I understand that this class is special in the sense it *is* native, so I am not insisting.

ev3dev2/motor.py
```diff
@@ -67,105 +67,126 @@
...
     """
     Speed in rotations-per-second.
     """
 
+    def __init__(self, rotations_per_second):
+        self.rotations_per_second = rotations_per_second
+
     def __str__(self):
-        return int.__str__(self) + " rps"
+        return str(self.rotations_per_second) + " rps"
```
ddemidov: Here it is 'rps', and above it is 'counts/sec'. Should we stick to one form?

ev3dev2/motor.py
```diff
@@ -879,67 +898,67 @@ def _set_brake(self, brake):
...
-        if not speed_pct or not rotations:
-            log.warning("({}) Either speed_pct ({}) or rotations ({}) is invalid, motor will not move" .format(self, speed_pct, rotations))
+        if not speed or not rotations:
+            log.warning("({}) Either speed ({}) or rotations ({}) is invalid, motor will not move" .format(self, speed, rotations))
             self._set_brake(brake)
             return
 
-        self.speed_sp = int((speed_pct * self.max_speed) / 100)
-        self._set_position_rotations(speed_pct, rotations)
+        self.speed_sp = int(speed)
```
ddemidov: How does `int(speed)` work? I don't see `__int__(self)` in definition of `SpeedMeasure`.

ddemidov: Ah, I see that `speed` is already in native counts.

ddemidov (commented):
I've left some comments, but overall this looks good to me.

ev3dev2/motor.py
WasabiFan: The `_speed_native_units` call will convert the parameter to a number in counts/sec, assuming percentage if it's a straight integer. So here `speed` is just a number already.

This isn't exactly right, though; it should be `int(round(speed))` so we round rather than truncating.

ev3dev2/motor.py
WasabiFan: Those sound good; I figure I'll rename to  `SpeedValue`.

ev3dev2/motor.py
WasabiFan: I'm a bit worried that people won't know what `CPS` is, and you can't easily look it up like you would for RPM because it isn't a common abbreviation. I do like the brevity, but I'm having trouble coming up with a name that's recognizable but still short.

ev3dev2/motor.py
ddemidov: Do we ever use `_speed_native_units` without `int(round(.))` around it? Is it worth making these a part of `to_native_units` call?

ev3dev2/motor.py
WasabiFan: "CPS" does show up on sites which list acronyms (I'm a bit surprised), so perhaps it's not as obscure as I thought... not sure.

ev3dev2/motor.py
WasabiFan: I think the string forms should error on the side of clarity over brevity, so how about I change them all to the form `rot/sec`, `deg/min`, etc?

ev3dev2/motor.py
ddemidov: Google says "CPS" is cost-per-sale :). But I did need to look into comments to understand what DPS or DPM was anyway. What do think of `SpeedUnits` instead of `SpeedNativeUnits`?

ev3dev2/motor.py
ddemidov: That sounds good

ev3dev2/motor.py
WasabiFan: Yeah, the `Move*` classes convert to native units then do math on it. Back when we used percentages instead, this led to rounding two or three times, which I'd say is significant. The tests were having to use comparison margins of something like +/-7. That would be less bad now that we're using native units so rounding has roughly 1/10 of the effect, but still I'd like to avoid intermediate rounding as much as possible.

It might make sense to add a utility method somewhere to do the rounding… not sure if it's worth it.

ddemidov: Ok, that is a good enough explanation for me.

ev3dev2/motor.py
WasabiFan: I feel like `SpeedUnits` would indicate that it's the base class without any particular units. Perhaps we could simplify this problem and call it `SpeedNative` -- I feel like that communicates that it's "whatever the motor measures in" while still being short.

ddemidov: `MotorSpeedUnits`? Anyway, if people will mostly use other convenience classes and won't have to type the name of this class all the time, then the point is probably moot.

WasabiFan:
Updated. I didn't change the `SpeedNativeUnits` name because we didn't come up with anything that was very clear and I figure people won't be using it much anyway.

All our CI builds are failing now, both on this branch and `ev3dev-stretch`. I'm not sure what changed, but it isn't our code.

ddemidov (approved):

ddemidov:
Looks like this line is the source of the CI troubles:

https://github.com/ev3dev/ev3dev-lang-python/blob/e2dd9a1180eed72e20a22f5f53744b06754b9048/.travis.yml#L9

ddemidov:
git blame says the commit that is responsible for that is https://github.com/ev3dev/ev3dev-lang-python/commit/6af0b14145bc5ebc6ec516bc2127efc5ba191971, but the commit was tested succesfully here: https://travis-ci.org/ev3dev/ev3dev-lang-python/builds/404695800?utm_source=github_status&utm_medium=notification

ddemidov:
Looks like pip install fails on building evdev:

https://travis-ci.org/ev3dev/ev3dev-lang-python/builds/423161128#L545

ddemidov:
Installing evdev 1.0.0 fixed Travis. (1.1.0 was released [about 4 days ago](https://pypi.org/project/evdev/#history))

ddemidov:
https://github.com/gvalkov/python-evdev/issues/98

WasabiFan:
Thanks for the investigation @ddemidov; everything looks good for us now, at least we have a workaround.

PR #502, opened 2018-09-06
Optimistically check the mode of the sensor before setting it on read

### LINKED ISSUE #498: IR sensor doesn't respond to toggling off the "beacon" button
- **ev3dev version:** 4.14.58-ev3dev-2.2.1-ev3
- **ev3dev-lang-python version:** python3-ev3dev2

Code: https://gist.github.com/Uldiniad/2608aa8e4a152fe443b151f0d6c5e08b
Output:
```
// Clicked beacon (runs medium_motor forever)
WARNING:root:State: True
WARNING:root:Running hand_motor forever
// Pressed touch sensor (stops medium_motor)
WARNING:root:Touch sensor pressed. Stopping hand_motor
WARNING:root:Touch sensor pressed. Stopping hand_motor
WARNING:root:Touch sensor pressed. Stopping hand_motor
WARNING:root:Touch sensor pressed. Stopping hand_motor
WARNING:root:Touch sensor pressed. Stopping hand_motor
WARNING:root:Touch sensor pressed. Stopping hand_motor
WARNING:root:Touch sensor pressed. Stopping hand_motor
WARNING:root:Touch sensor pressed. Stopping hand_motor
WARNING:root:Touch sensor pressed. Stopping hand_motor
WARNING:root:Touch sensor pressed. Stopping hand_motor
WARNING:root:Touch sensor pressed. Stopping hand_motor
// Press any other button than beacon (runs large motor until released)
WARNING:root:State: False **//anomaly: large motors should not be in auto_home nor its internals**
// Clicked beacon again (runs medium_motor forever)
WARNING:root:State: True
WARNING:root:Running hand_motor forever
// Pressed touch sensor again (stops medium_motor)
WARNING:root:Touch sensor pressed. Stopping hand_motor
WARNING:root:Touch sensor pressed. Stopping hand_motor
WARNING:root:Touch sensor pressed. Stopping hand_motor
WARNING:root:Touch sensor pressed. Stopping hand_motor
// Pressed beacon again, nothing happens **anomaly: every time beacon is pressed it should execute move(state)*. The only way to have beacon run the medium_motor forever again is to execute the sequence to trigger the first anomaly*
```

WasabiFan:
As we've [recently discovered](https://github.com/ev3dev/ev3dev-lang-python/issues/498), some sensors don't respond well to having their mode set repeatedly<sup>1</sup>. Here's my proposed fix, which I think simultaneously improves sensor operation all-around.

<sup>1: _My guess is that this was intentional on LEGO's part; the beacon button in particular has a delay on it, presumably added on the sensor side to be resilient to momentary losses of line-of-sight. That timer is probably reset every time we write the mode._</sup>

---

I did some benchmarking with an IR sensor (UART):

```
robot@ev3dev:~/ev3dev-lang-python$ python3 -m timeit -s "import ev3dev2.sensor; s = ev3dev2.sensor.Sensor(ev3dev2.sensor.INPUT_3)" "s.mode"
1000 loops, best of 3: 975 usec per loop

robot@ev3dev:~/ev3dev-lang-python$ python3 -m timeit -s "import ev3dev2.sensor; s = ev3dev2.sensor.Sensor(ev3dev2.sensor.INPUT_3)" "s.mode = 'IR-REMOTE'"
10 loops, best of 3: 97.8 msec per loop
```

In other words, reading the mode takes around 1ms, and writing takes 100x that. This is generally what I would expect, given that setting the mode on a UART sensor always requires writing to the serial line and might even wait for a response... I haven't looked into it past a cursory glance at the top-level "set" method. I believe reading is just a memory read (+ sysfs overhead).

For comparison, here's an analog sensor (the only one IIRC), which only has one mode and presumably no-ops when setting the mode anyway (I didn't bother looking up the driver code for this one):

```
robot@ev3dev:~/ev3dev-lang-python$ python3 -m timeit -s "import ev3dev2.sensor; s = ev3dev2.sensor.Sensor(ev3dev2.sensor.INPUT_3)" "s.mode"
1000 loops, best of 3: 834 usec per loop

robot@ev3dev:~/ev3dev-lang-python$ python3 -m timeit -s "import ev3dev2.sensor; s = ev3dev2.sensor.Sensor(ev3dev2.sensor.INPUT_3)" "s.mode = 'TOUCH'"
1000 loops, best of 3: 827 usec per loop
````

So, this one is around 1ms for both operations. I'm going to guess that this is just sysfs file I/O overhead.

I didn't see much use in finding an I2C sensor, given that pretty much all of the EV3 ones are UART. It's probably somewhere in between.

---

What this tells me is that we wait 100ms every time you read a sensor, for no good reason. Given that reading the mode is only 1ms, it seems reasonable to me that we should do the read and only write if we need it to change. In most cases, that's probably the right assumption. Technically, this does introduce a potential race condition, but I think the risk is minor.

 I've included that in this PR, and added some related tests.

Fixes #498

### DIFF
--- a/ev3dev2/sensor/lego.py
+++ b/ev3dev2/sensor/lego.py
@@ -54,7 +54,7 @@ class TouchSensor(Sensor):
         A boolean indicating whether the current touch sensor is being
         pressed.
         """
-        self.mode = self.MODE_TOUCH
+        self._ensure_mode(self.MODE_TOUCH)
         return self.value(0)
 
     @property
@@ -182,7 +182,7 @@ class ColorSensor(Sensor):
         """
         Reflected light intensity as a percentage. Light on sensor is red.
         """
-        self.mode = self.MODE_COL_REFLECT
+        self._ensure_mode(self.MODE_COL_REFLECT)
         return self.value(0)
 
     @property
@@ -190,7 +190,7 @@ class ColorSensor(Sensor):
         """
         Ambient light intensity. Light on sensor is dimly lit blue.
         """
-        self.mode = self.MODE_COL_AMBIENT
+        self._ensure_mode(self.MODE_COL_AMBIENT)
         return self.value(0)
 
     @property
@@ -206,7 +206,7 @@ class ColorSensor(Sensor):
           - 6: White
           - 7: Brown
         """
-        self.mode = self.MODE_COL_COLOR
+        self._ensure_mode(self.MODE_COL_COLOR)
         return self.value(0)
 
     @property
@@ -226,7 +226,7 @@ class ColorSensor(Sensor):
 
         If this is an issue, check out the rgb() and calibrate_white() methods.
         """
-        self.mode = self.MODE_RGB_RAW
+        self._ensure_mode(self.MODE_RGB_RAW)
         return self.value(0), self.value(1), self.value(2)
 
     def calibrate_white(self):
@@ -384,7 +384,7 @@ class ColorSensor(Sensor):
         """
         Red component of the detected color, in the range 0-1020.
         """
-        self.mode = self.MODE_RGB_RAW
+        self._ensure_mode(self.MODE_RGB_RAW)
         return self.value(0)
 
     @property
@@ -392,7 +392,7 @@ class ColorSensor(Sensor):
         """
         Green component of the detected color, in the range 0-1020.
         """
-        self.mode = self.MODE_RGB_RAW
+        self._ensure_mode(self.MODE_RGB_RAW)
         return self.value(1)
 
     @property
@@ -400,7 +400,7 @@ class ColorSensor(Sensor):
         """
         Blue component of the detected color, in the range 0-1020.
         """
-        self.mode = self.MODE_RGB_RAW
+        self._ensure_mode(self.MODE_RGB_RAW)
         return self.value(2)
 
 
@@ -440,12 +440,12 @@ class UltrasonicSensor(Sensor):
 
     @property
     def distance_centimeters_continuous(self):
-        self.mode = self.MODE_US_DIST_CM
+        self._ensure_mode(self.MODE_US_DIST_CM)
         return self.value(0) * self._scale('US_DIST_CM')
 
     @property
     def distance_centimeters_ping(self):
-        self.mode = self.MODE_US_SI_CM
+        self._ensure_mode(self.MODE_US_SI_CM)
         return self.value(0) * self._scale('US_DIST_CM')
 
     @property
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: tests/api_tests.py [hunk 1]
```diff
@@ -93,6 +93,19 @@ def test_infrared_sensor(self):
         self.assertEqual(s.num_values,      1)
         self.assertEqual(s.address,         'in1')
         self.assertEqual(s.value(0),        16)
+        self.assertEqual(s.mode,            "IR-PROX")
+
+        s.mode = "IR-REMOTE"
+        self.assertEqual(s.mode,            "IR-REMOTE")
+
+        val = s.proximity
+        # Our test environment writes to actual files on disk, so while "seek(0) write(...)" works on the real device, it leaves trailing characters from previous writes in tests. "s.mode" returns "IR-PROXTE" here.
```
WasabiFan, line 113: There isn't an obvious easy fix here so I'm inclined to just leave this as-is; open to suggestions.

ddemidov: We could monkey patch the sensor class when on Travis.

WasabiFan: 👍  Done

ddemidov (approved):
This makes a lot of sense.

ev3dev2/sensor/lego.py
```diff
@@ -440,12 +440,12 @@ def __init__(self, address=None, name_pattern=SYSTEM_DEVICE_NAME_CONVENTION, nam
...
     @property
     def distance_centimeters_continuous(self):
-        self.mode = self.MODE_US_DIST_CM
+        self._ensure_mode(self.MODE_US_DIST_CM)
         return self.value(0) * self._scale('US_DIST_CM')
 
     @property
     def distance_centimeters_ping(self):
-        self.mode = self.MODE_US_SI_CM
+        self._ensure_mode(self.MODE_US_SI_CM)
```
dlech: This mode is special. The value does not update unless you set the mode. So this change will cause the sensor the become unresponsive after the first call.

ev3dev2/sensor/lego.py
```diff
@@ -458,12 +458,12 @@ def distance_centimeters(self):
...
     @property
     def distance_inches_continuous(self):
-        self.mode = self.MODE_US_DIST_IN
+        self._ensure_mode(self.MODE_US_DIST_IN)
         return self.value(0) * self._scale('US_DIST_IN')
 
     @property
     def distance_inches_ping(self):
-        self.mode = self.MODE_US_SI_IN
+        self._ensure_mode(self.MODE_US_SI_IN)
```
dlech: Ditto on this mode.

ev3dev2/sensor/lego.py
dlech: interesting, it also appears that this mode has a problem with locking up the sensor if you poll it too often. http://docs.ev3dev.org/projects/lego-linux-drivers/en/ev3dev-stretch/sensor_data.html#lego-ev3-us-mode3-value0

It would be worth copying that note to the python docs if it isn't there already

WasabiFan: Thanks; that's a very good point. I'll revert these two back to just setting the mode directly and copy over the note.

PR #514, opened 2018-09-10
Micropython support

### LINKED ISSUE #492: Micropython support
It looks like we don't have an issue for this yet.

This is my highest priority after releasing the initial v2.0; I've already done a bit of work to get this up-and-running but not all the core modules are ready and tested yet. I'm putting it in the preemptive v2.1 milestone to track it; the main thing I still don't know is how packaging has to work.

WasabiFan:
- [x] Make necessary changes to support motors and sensors on micropython (cursory on-EV3 testing)
- [x] Get tests up-and-running on Travis with Micropython entry in build matrix
- [x] Thorough testing pass on-device
- [ ] Better handling of unsupported features (currently buttons and sound)
- [ ] ~~Document usage under Micropython~~ (#528)
  - Test out `micropython.native` decorator to see if it works
- [ ] APT packaging
- Stretch goal: support buttons and sound

Example build: https://travis-ci.org/ev3dev/ev3dev-lang-python/builds/426662475

Depends on https://github.com/ev3dev/ev3dev-lang-fake-sys/pull/13
Fixes #492


ev3dev2/__init__.py
```diff
@@ -28,6 +28,15 @@
 if sys.version_info < (3,4):
     raise SystemError('Must be using Python 3.4 or higher')
 
+def is_micropython():
+    return sys.implementation.name == "micropython"
+
+def chain_exception(exception, cause):
```
WasabiFan, line 34: I'm not sure if I like how I did this... Micropython doesn't support exception chaining, so we have to suppress the `from` when running on that interpreter. But now `chain_exception` shows up in the call stack.

.travis/install-micropython.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/sh
+set -ex
+
+cd ~/
+
+# Build micropython from source
+# TODO: cache micropython build output
```
ddemidov, line 12: Here is how I did it:

Adding a path to travis cache:

https://github.com/ddemidov/amgcl/blob/master/.travis.yml#L39-L42

Checking if path exists (that is, if cache was unpacked successfully):

https://github.com/ddemidov/amgcl/blob/master/.travis/install_boost.sh#L5

ev3dev2/__init__.py
ddemidov: I guess we could implement our own Exception subclass, that would carry its parent exception around, but I am not sure if that would be any better than what you did.

ddemidov (approved):
That is surprisingly small amount of changes to support micropython. Looks good to me!

ev3dev2/__init__.py
```diff
@@ -242,20 +252,22 @@ def _raise_friendly_access_error(self, driver_error, attribute):
         if not isinstance(driver_error, OSError):
             raise driver_error
 
-        if driver_error.errno == errno.EINVAL:
+        driver_errorno = driver_error.args[0] if is_micropython() else driver_error.errno
```
ddemidov, line 258: It would feel nicer if this would be implemented as a function (`driver_errno()`) but I guess it is OK as is, since this is the only place we need it.

.travis.yml
```diff
@@ -1,23 +1,28 @@
...
 deploy:
   provider: pypi
   user: Denis.Demidov
   password:
     secure: cTDN4WMJEjxZ0zwFwkLPpOIOSG/1JlHbQsHzd/tV9LfjBMR0nJR5HrmyiIO1TE5Wo6AFSFd7S3JmKdEnFO1ByvN/EFtsKGRUR9L6Yst4KfFi4nHwdZ7vgPTV0nNdvgC1YM/3bx8W8PcjFSm/k8awx7aicwXj09yDA8ENTf5XedaX22Z+9OKhb1mKola1cqEoc0GwaYzd8UX0Ruwh9/6RRbvTt7zn8BCZc9vIVqNd6mZgBWY9zAU40ZZsjYORbiZmDNCygEI+RViZ51M58WYkPPhoivzcG9em8DMRS7SC3CjWGapiOaXUHa3Fhnn+IQ+Q8Xv9YU5+qmj65MWQy3SSnMnvuxmrLf4aLoOlSJUMhDarjni4tdBOTX5PdOkdmhskyQt1DqDrw0+WhPItYfGe5zQfQwqW+YOpGbOipAeU+844arPo5jvZG/IOBX2qVUwdSxo8Y/98ocjqoZOq8b5xkWtJef0Kh1RCkp1bR2XELQVe76qeWqQxWz3OPqq+wK3xeNvj5kMQmytl3dCEB//D6UcES7Qr8YxD+LWoaIf32JIj/4LaCXXuxMVH+PJ68Oc72/ox0qLmXK0qhbea2QvaqXyGDrO+a2X6VbMdH32D2xHzH4Mg75xLnXnSaFvGovhYl1zEVcYUDioxrXZEuDmymGf9nH2mivJ24Fon6u+C3QQ=
   on:
     tags: true
-    condition: $TRAVIS_TAG != ev3dev-stretch/*
+    condition:
+      - $TRAVIS_TAG != ev3dev-stretch/*
```
WasabiFan, line 30: I suspect that this (the list for `condition`)  isn't right. I need to take another look and figure out how to do it correctly.

WasabiFan: Never mind... although the documentation says

> condition: deploy when a _single_ bash condition evaluates to true. This must be a string value, and is equivalent to `if [[ <condition> ]]; then <deploy>; fi`.

It emits the correct script:

```bash
  if [[ ($TRAVIS_REPO_SLUG = "ev3dev/ev3dev-lang-python") && ($TRAVIS_TAG != ev3dev-stretch/*) && ($USE_MICROPYTHON = false) && ("$TRAVIS_TAG" != "") ]]; then
```

.travis/install-micropython.sh
WasabiFan: Done. Thanks for the reference!

PR #516, opened 2018-09-11
Fixed MoveSteering error when steering value is >50 or <-50, degrees …

stubs12:
Fix a problem where MoveSteering.on_for_degrees() with steering values larger than 50 or smaller than -50 caused negative degrees to be calculated, leading to an assert.  Just used abs() on the calculated degrees, the negative speed takes care of the direction.

### DIFF
--- a/ev3dev2/motor.py
+++ b/ev3dev2/motor.py
@@ -1808,9 +1808,9 @@ class MoveTank(MotorSet):
 
         if left_speed_native_units > right_speed_native_units:
             left_degrees = degrees
-            right_degrees = float(right_speed / left_speed_native_units) * degrees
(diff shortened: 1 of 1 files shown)


WasabiFan:
This looks good to me; there definitely was an error on my part when changing units here. Could you [add a test](https://github.com/ev3dev/ev3dev-lang-python/blob/ev3dev-stretch/tests/api_tests.py#L137) for this case as well?

WasabiFan:
I have included the fix from here for the bad property name in #523. We are now removing the requirement for positive powers, so the code around this area has been changed by that PR to solve the problem. I won't be merging the PR as the relevant part is in #523. Thanks for the contribution!

Issue #515, opened 2018-09-11
Blocking motor calls return when motor state becomes "stalled"


WasabiFan:
Currently, we use `wait_until_not_moving` to determine when to return from a blocking motor call. From its docstring:

> Blocks until `running` is not in `self.state` or `stalled` is in `self.state`. The condition is checked when there is an I/O event related to the `state` attribute.  Exits early when `timeout` (in milliseconds) is reached.
> Returns `True` if the condition is met, and `False` if the timeout is reached.

In other words, if I request to move the motor for ten rotations and then part way through its motion the motor stalls, the call returns then while the motor is still trying to complete the movement. Additionally, there doesn't seem to be any way that the caller can tell whether the return was triggered by a `stalled` state or not.

Does this seem like the right behavior? It makes sense to me that a user might not want their code to freeze if the motor stalls, but at the same time I feel like this blurs the contract of what it means for one of these calls to return.

ddemidov:
You could in principle check `motor.state` immediately upon return from the function. We can make it clear in the documentation that the function may spuriously return (when the motor stalls temporarily), similar to how it's done in [C++ documentation for `std::condition_variable::wait`](https://en.cppreference.com/w/cpp/thread/condition_variable/wait):

> wait causes the current thread to block until the condition variable is notified or a spurious wakeup occurs

WasabiFan:
I think, if we want to support the "block and then check state" pattern, we should look into returning a flag from these blocking functions indicating whether the motor completed its movement or not (based on why we returned). Checking the `state` afterward seems like a recipe for race conditions. I believe this would require some additional data from `wait_until_not_moving`.

Some other options:

2. Add a parameter (or add a new valid value to the existing one) so the user can specify whether they want to return on stall or not. Perhaps the default could be _no_ (i.e., "if the motor isn't at its goal I don't want to continue"), and then you could switch the behavior to signal that you are now ready to handle the early return. 
3. Throw an exception when the motor stalls. I think this would make a lot of sense in the case where the user knows to add a `try` where needed, but makes it seem fatal to a user who doesn't.

I kind of like (2), perhaps in combination with either (1) or (3). I think the current behavior might make programs behave inconsistently when running at low speeds, so making "don't return on stall" the default option appeals to me.

dwalton76:
The current behavior is pretty critical for Mindcuber, that bot often has to do motor turns that result in stalls. 

Option 2 above is doable today via a wait_while(“running”) call.

[...]

PR #523, opened 2018-09-15
Flexible relative movement and improved parameter handling

### LINKED ISSUE #504: python-ev3dev v2 won't allow negative angles for motors
ev3dev version: 4.14.61-ev3dev-2.2.2-ev3
ev3dev-lang-python version: 
ii  python3-ev3dev                 1.2.0                    all                     Python language bindings for ev3dev
ii  python3-ev3dev2               2.0.0~beta1        all                      Python language bindings for ev3dev

I have been exploring EV3 Python v2 in depth over two weeks and have sent dozens of comments to @wasabifan. He has kindly given lots of feedback, suggesting which comments should be raised as issues here. Based on that feedback I am publishing several issues, of which this is no doubt the most serious.

The rule that angles must be positive for **on_for_rotations** and **on_for_degrees** for single motors and motor pairs is catastrophic, it seems to me! Even these lines generate errors:
```
#!/usr/bin/env python3
from ev3dev2.motor import MoveSteering, OUTPUT_B, OUTPUT_C
steer_pair = MoveSteering(OUTPUT_B, OUTPUT_C)
steer_pair.on_for_degrees(steering=100, speed=40, degrees=685)
```
Error: 
  File "/home/robot/MINE/Basics_3.py", line 9, in <module>
    steer_pair.on_for_degrees(steering=100, speed=40, degrees=685)
  File "/usr/lib/python3/dist-packages/ev3dev2/motor.py", line 1899, in on_for_degrees
    MoveTank.on_for_degrees(self, SpeedNativeUnits(left_speed), SpeedNativeUnits(right_speed), degrees, brake, block)
  File "/usr/lib/python3/dist-packages/ev3dev2/motor.py", line 1807, in on_for_degrees
    self.right_motor._set_position_degrees(right_speed_native_units, right_degrees)
  File "/usr/lib/python3/dist-packages/ev3dev2/motor.py", line 869, in _set_position_degrees
    assert degrees >= 0, "degrees is %s, must be >= 0" % degrees
AssertionError: **degrees is -685.0, must be >= 0**

It seems that any steering command using steering < -50 or steering > 50 will fail since this attempts to turn a wheel backwards by setting the angle negative, which is forbidden.
This also fails, with the same error:
`tank_pair.on_for_degrees(left_speed=30, right_speed=-40, degrees=60)`
AssertionError: **degrees is -7.619047619047619, must be >= 0**

And of course this fails
`mB.on_for_degrees(speed_pct=50, degrees=-60)`
AssertionError: **degrees is -60, must be >= 0**

PLEASE allow negative angles! It will make everything so much simpler and anyone smart enough to be able to program in EV3 Python must surely be able to understand that if both the speed and angle are negative then the result will be a turn in the positive angle direction with a positive speed since that is mathematically equivalent to a turn at a negative speed through a negative angle. There is no contradiction. It's a mathematical fact that (-6) * (-3) = (+6) * (+3). That doesn’t mean that -6 is not negative.

You want Python v2 to behave like EV3-G where possible but EV3-G allows negative angles. A negative power and negative angle cause the motor to turn in the positive direction, no problem.

### LINKED ISSUE #505: python-ev3dev v2 program freezes if go to current position
ev3dev version: 4.14.61-ev3dev-2.2.2-ev3
ev3dev-lang-python version:
ii python3-ev3dev 1.2.0 all Python language bindings for ev3dev
ii python3-ev3dev2 2.0.0~beta1 all Python language bindings for ev3dev

If the script instructs the motor to go to a position that it is already in then this freezes the program! For example, this will freeze the program (i.e. the program will not terminate):
```
!/usr/bin/env python3
from ev3dev2.motor import LargeMotor, OUTPUT_B
from time import sleep
large_motor = LargeMotor(OUTPUT_B)
large_motor.on_to_position(speed_pct=10, position=-120)
sleep(2)
large_motor.on_to_position(speed_pct=10, position=-120) 
```

### LINKED ISSUE #507: python3-ev3dev v2 angles of zero work badly with on_for_*()
ev3dev version: 4.14.61-ev3dev-2.2.2-ev3
ev3dev-lang-python version:
ii python3-ev3dev 1.2.0 all Python language bindings for ev3dev
ii python3-ev3dev2 2.0.0~beta1 all Python language bindings for ev3dev

This is a separate issue to the problem of **negative angles** not being handled correctly by the new python3-ev3dev v2 motor functions that I reported in issue #504.

It seems to me that if angles of zero are used with the new v2 commands **on_for_degrees()** and **on_for_rotations()** for motor **pairs** then the program will freeze. This is surely a bug.

I recently tried to run this script which simply sets the speed of the robot to equal the ambient light intensity:

```
 #!/usr/bin/env python3
from ev3dev2.motor import MoveSteering, OUTPUT_B, OUTPUT_C
from ev3dev2.sensor.lego import ColorSensor
from ev3dev2.button import Button
from time import sleep

cl = ColorSensor()
btn = Button()
steer_pair = MoveSteering(OUTPUT_B, OUTPUT_C)

while not btn.any():    # Stop program by pressing any button
    steer_pair.on(steering=0,speed=cl.ambient_light_intensity)
    sleep(0.2)
```

The script should work but quickly fails because as soon as the light intensity falls to zero there is an error:
  File "/home/robot/MINE/example_6.py", line 12, in <module>
    steer_pair.on(steering=0,speed=cl.ambient_light_intensity)
  File "/usr/lib/python3/dist-packages/ev3dev2/motor.py", line 1913, in on
    MoveTank.on(self, SpeedNativeUnits(left_speed), SpeedNativeUnits(right_speed))
  File "/usr/lib/python3/dist-packages/ev3dev2/motor.py", line 1844, in on
    (left_speed_native_units, right_speed_native_units) = self._unpack_speeds_to_native_units(left_speed, right_speed)
  File "/usr/lib/python3/dist-packages/ev3dev2/motor.py", line 1739, in _unpack_speeds_to_native_units
    "Either left_speed or right_speed must be non-zero"
AssertionError: **Either left_speed or right_speed must be non-zero**

There also seem to be a problem with **single** motor commands that use angles, though in this case a warning is triggered rather than an error. Here is a script to demonstrate how an angle of zero makes **single** motor commands give a warning (as opposed to an error) while an angle of zero for a **motor pair** makes the program freeze:

```
#!/usr/bin/env python3
from ev3dev2.motor import MoveSteering, LargeMotor, OUTPUT_B, OUTPUT_C
from time import sleep

sound = Sound()
lm = LargeMotor(OUTPUT_B)
steer_pair = MoveSteering(OUTPUT_B, OUTPUT_C)

sound.beep()
'''
For single motors, 
degrees=0 triggers a WARNING that zero is not valid but it is a warning
not an error, and the program continues correctly
'''
lm.on_for_degrees(speed_pct=30, degrees = 0)

sleep(5)
sound.beep()
'''
For motor pairs, 
degrees=0 causes the script to FREEZE, so it never completes
'''
steer_pair.on_for_degrees(steering=0, speed=30, degrees = 0)
```

As shown in the first script, it is perfectly normal that a motor or motor pair may sometimes receive the order to rotate through zero degrees, so the motor or motor pair must be able to handle this without giving a warning or terminating the program. I imagine EV3-G is perfectly happy with power values of zero and that they do cause the program to terminate.

WasabiFan:
The primary goals here were to:
- Remove unnecessary cases where we warn or throw an error
- Make handling of negative speeds and distances intuitive/logical, like multiplication
- Fix cases of zero distance and/or speed

In the process, I:
- Refactored `_set_position_rotations` and `_set_position_degrees`
  - To minimize duplication there is now only one, which operates on degrees, and rotations-based methods call the degree versions. This is one more function call (which is slow) but I figure it's worth it.
  - The method is called `_set_rel_position_degrees_and_speed_sp` and handles setting both `position_sp` _and_ `speed_sp` so it can normalize negatives
- Replaced our custom `position_sp = position + delta; command = 'run-to-abs-pos'` with using `run-to-rel-pos`. I read through #360 where this was originally implemented, and don't see why it was done this way. (let me know if I missed something, @dwalton76)
- Updated the logic to choose distances for each motor so that it correctly handles negative speeds
- Removed a warning that really wasn't helpful
 
One question: On an individual motor, if you ask it to move a nonzero distance at zero speed, it will return immediately because the driver calls that a "stalling" state. However, with a motor pair, I adopted a different behavior: in that particular case it throws a `ValueError`. Do others agree with this set of behaviors or should one of them be changed?

Closes #516 (supersedes changes)
Fixes #504 
Fixes #505
Fixes #507

### DIFF
--- a/ev3dev2/motor.py
+++ b/ev3dev2/motor.py
@@ -1808,9 +1808,9 @@ class MoveTank(MotorSet):
 
         if left_speed_native_units > right_speed_native_units:
             left_degrees = degrees
-            right_degrees = float(right_speed / left_speed_native_units) * degrees
+            right_degrees = abs(float(right_speed_native_units / left_speed_native_units) * degrees)
         else:
-            left_degrees = float(left_speed_native_units / right_speed_native_units) * degrees
+            left_degrees = abs(float(left_speed_native_units / right_speed_native_units) * degrees)
             right_degrees = degrees
 
         # Set all parameters
--- a/ev3dev2/motor.py
+++ b/ev3dev2/motor.py
@@ -1752,12 +1752,12 @@ class MoveTank(MotorSet):
         # v_l = d_l/t, v_r = d_r/t
         # therefore, t = d_l/v_l = d_r/v_r
 
-        # TODO: this is wrong. consider negatives.
-        if left_speed_native_units > right_speed_native_units:
+        # larger speed by magnitude is the "outer" wheel, and rotates the full "rotations"
+        if abs(left_speed_native_units) > abs(right_speed_native_units):
             left_rotations = rotations
-            right_rotations = abs(float(right_speed_native_units / left_speed_native_units)) * rotations
+            right_rotations = abs(right_speed_native_units / left_speed_native_units) * rotations
         else:
-            left_rotations = abs(float(left_speed_native_units / right_speed_native_units)) * rotations
+            left_rotations = abs(left_speed_native_units / right_speed_native_units) * rotations
             right_rotations = rotations
 
         # Set all parameters
@@ -1786,11 +1786,12 @@ class MoveTank(MotorSet):
         """
         (left_speed_native_units, right_speed_native_units) = self._unpack_speeds_to_native_units(left_speed, right_speed)
 
-        if left_speed_native_units > right_speed_native_units:
+        # larger speed by magnitude is the "outer" wheel, and rotates the full "degrees"
+        if abs(left_speed_native_units) > abs(right_speed_native_units):
             left_degrees = degrees
-            right_degrees = abs(float(right_speed_native_units / left_speed_native_units)) * degrees
+            right_degrees = abs(right_speed_native_units / left_speed_native_units) * degrees
         else:
-            left_degrees = abs(float(left_speed_native_units / right_speed_native_units)) * degrees
+            left_degrees = abs(left_speed_native_units / right_speed_native_units) * degrees
             right_degrees = degrees
 
         # Set all parameters
--- a/ev3dev2/motor.py
+++ b/ev3dev2/motor.py
@@ -1746,33 +1746,7 @@ class MoveTank(MotorSet):
         ``rotations`` while the motor on the inside will have its requested
         distance calculated according to the expected turn.
         """
-        (left_speed_native_units, right_speed_native_units) = self._unpack_speeds_to_native_units(left_speed, right_speed)
-
-        # proof of the following distance calculation: consider the circle formed by each wheel's path
-        # v_l = d_l/t, v_r = d_r/t
-        # therefore, t = d_l/v_l = d_r/v_r
-
-        # larger speed by magnitude is the "outer" wheel, and rotates the full "rotations"
-        if abs(left_speed_native_units) > abs(right_speed_native_units):
-            left_rotations = rotations
-            right_rotations = abs(right_speed_native_units / left_speed_native_units) * rotations
-        else:
-            left_rotations = abs(left_speed_native_units / right_speed_native_units) * rotations
-            right_rotations = rotations
-
-        # Set all parameters
-        self.left_motor._set_rel_position_degrees_and_speed_sp(left_rotations * 360, left_speed_native_units)
-        self.left_motor._set_brake(brake)
-        self.right_motor.speed_sp = int(round(right_speed_native_units))
-        self.right_motor._set_rel_position_degrees_and_speed_sp(right_rotations * 360, right_speed_native_units)
-        self.right_motor._set_brake(brake)
-
-        # Start the motors
-        self.left_motor.run_to_rel_pos()
-        self.right_motor.run_to_rel_pos()
-
-        if block:
-            self._block()
+        MoveTank.on_for_degrees(self, left_speed, right_speed, rotations * 360, brake, block)
 
     def on_for_degrees(self, left_speed, right_speed, degrees, brake=True, block=True):
         """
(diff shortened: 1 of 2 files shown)


ev3dev2/motor.py
```diff
@@ -1804,26 +1758,36 @@ def on_for_degrees(self, left_speed, right_speed, degrees, brake=True, block=Tru
         ``degrees`` while the motor on the inside will have its requested
         distance calculated according to the expected turn.
         """
+
+        if left_speed == 0 and right_speed == 0 and degrees != 0:
+            raise ValueError("Can't travel a nonzero distance at zero speed")
```
ddemidov: Two points: 
* There is probably some larger than zero speed threshold where motor will stall, so zero is not that special
* I can imagine a situation where speed is set by some function (e.g. based on user input), where it may legitimately cross zero. Not sure if raising an exception would be a good idea in this case.

WasabiFan: I went back-and-forth on this so I'm not attached to this behavior. I agree with your points; what would you suggest the behavior be?

If we ask the motor driver to go a nonzero distance with zero speed, it will report being `running stalled` indefinetely, and in turn will immediately cause our `wait_until_not_moving` to return. So would your suggestion be to adopt that behavior? That feels weird to me, but it's consistent with our single-motor methods (#515) so probably makes sense.

ddemidov: > If we ask the motor driver to go a nonzero distance with zero speed, it will report being running stalled indefinetely, and in turn will immediately cause our wait_until_not_moving to return.

I think a return from `wait_until_not_moving()` here is an expected outcome. The speed the user asked was too small, so the motor is stalled, and is not moving.

WasabiFan: OK, thanks for the second opinion; I've updated the PR.

WasabiFan:
@ddemidov When you have the chance, can you take another look at this? Once this PR and #524 are good-to-go I plan to release a new version of the package.

WasabiFan:
I'm going to merge this now and do some end-to-end testing so I can release the package today. If there are any issues here that I haven't caught, we can fix them in a separate PR.

PR #524, opened 2018-09-15
Update changelog for v2.0.0-beta2

WasabiFan:
When I get approval here I plan to -- pending approval on the latter -- merge #516 and #523 and then do a package release. There are a couple other issues I intend to work on over the coming days, but there's enough good stuff in here that I feel like we should make it available so delays on my side don't drag things out longer.


WasabiFan:
49b48e8ec468952099e5d0ebf172350af370cae3

Issue #392, opened 2017-10-04
Differential drive class


fisherds:
This is just a potential feature request not a bug.  (probably a duplicate of some other feature request but I didn't find it)

There are many EV3 Lego creations that are mobile robots (most EV3 designs have wheels or tracks actually).  When doing the programming for a mobile robot, users need to convert motor degrees into distances and into whole robot turn amounts.  Obviously we can do these conversions, but for new users it would be nice to provide a convenience class for EV3 mobile robots.  My suggestion for a name is Driver, but it could be any name we like, DifferentialDrive, although more accurate, just felt too long.

Example usage:
`import ev3dev.ev3 as ev3
robot_base = ev3.Driver(ev3.OUTPUT_B, ev3.OUTPUT_C, 3.3, 17)`

Where the arguments are:
- Left motor port
- Right motor port
- Wheel diameter (in cm in this case)
- Distance between wheels (in cm in this case)
- Plus an optional parameter not show to reverse motor direction

Once constructed users could use commands like:
```
- robot_base.drive(100)  # to drive forward 100 cm
- robot_base.turn(90) # to turn right 90 degrees
- robot_base.arc_right(15) # arc right with about a circle with a radius of 15
- robot_base.units = ev3.Driver.UNITS_INCHES  # cm by default
- robot_base.driving_speed = 4  # inches per second now, since units changed to degrees
- robot_base.turning_speed = 300 # degrees per second
```

These are not my ideas.  I've used Clinton Blackmore's Enchanting program (https://www.youtube.com/watch?v=N9HZwysMqWY) for NXT with kids for years.  It's an old Scratch based program from 2013 for NXT that he has not updated or maintained.    His differential drive mechanism is extremely flexible, robust and easy to use.  We use his driving commands with the old NXT robot fleet very heavily.  It'd be useful if we provided a Driver class of some sort.

Obviously there are a lot of design decisions about the structure of such a class and there are issues with turning inaccuracies with tank treads instead of two wheels.  Despite the concerns it'd be a highly used convenience class.

Here is Clinton's Scratch based configuration of differential drive
![configure_drive](https://user-images.githubusercontent.com/295078/31155750-76d9d6b4-a87e-11e7-9b5c-a47eae9bff04.png)

And some of this Scratch commands:
![commands](https://user-images.githubusercontent.com/295078/31155761-88ff95c2-a87e-11e7-90b1-6e4878db6ed1.png)

[...]

PR #536, opened 2018-10-26
MoveDifferential

dwalton76:
For issue #392


ev3dev2/__init__.py
```diff
@@ -209,6 +209,9 @@ def __str__(self):
     def __repr__(self):
         return self.__str__()
 
+    def __lt__(self, other):
+        return str(self) < str(other)
```
dwalton76, line 214: This allows us to sort lists of Device objects

ev3dev2/motor.py
```diff
@@ -67,13 +67,19 @@
...
     :class:`SpeedPercent`, :class:`SpeedRPS`, :class:`SpeedRPM`,
     :class:`SpeedDPS`, and :class:`SpeedDPM`.
     """
-    pass
+
+    def __lt__(self, other):
+        return self.to_native_units() < other.to_native_units()
+
+    def __rmul__(self, other):
+        return self.__mul__(other)
```
dwalton76, line 86: So we can multiple SpeedValue objects by an int, float or another SpeedValue object of the same type.  I'm not sure how much use multiplying by another object of the same type will be but being able to scale one by multiple by a float or reverse the direction by multiplying by -1 is handy.

ev3dev2/motor.py
```diff
@@ -89,11 +95,20 @@ def __init__(self, percent):
...
+            return SpeedPercent(self.percent * other)
+        else:
+            raise Exception("{} can only be multiplied by a SpeedPercent, int, or float".format(self))
+
     def to_native_units(self, motor):
         """
         Return this SpeedPercent in native motor units
         """
-        return self.percent / 100 * motor.max_speed
+        return round((self.percent * motor.max_speed) / 100)
```
dwalton76: speed_sp must be an integer so I changed all of the to_native_unit() calls to round() what they are returning

ev3dev2/motor.py
```diff
@@ -697,22 +762,25 @@ def time_sp(self, value):
         self._time_sp = self.set_attr_int(self._time_sp, 'time_sp', value)
 
     def run_forever(self, **kwargs):
-        """Run the motor until another command is sent.
+        """
```
dwalton76, line 747: Making the docstring format consistent throughout the file

ev3dev2/motor.py
```diff
@@ -1569,6 +1652,8 @@ def __init__(self, motor_specs, desc=None):
         for motor_port in sorted(motor_specs.keys()):
             motor_class = motor_specs[motor_port]
             self.motors[motor_port] = motor_class(motor_port)
+            self.motors[motor_port].ramp_up_sp = 0
+            self.motors[motor_port].ramp_down_sp = 0
```
dwalton76: When we create a new Motor object should we be initializing most variables?  It hit me that if the previous robot that ran set ramp_up_sp or ramp_down_sp that could cause problems with MoveDifferential trying to drive in a perfect circle, etc.

ev3dev2/motor.py
```diff
@@ -1725,10 +1822,8 @@ def __init__(self, left_motor_port, right_motor_port, desc=None, motor_class=Lar
         self.max_speed = self.left_motor.max_speed
 
     def _block(self):
```
dwalton76: Thoughts on moving this up to the MotorSet class?

ev3dev2/motor.py
```diff
@@ -1739,18 +1834,6 @@ def _unpack_speeds_to_native_units(self, left_speed, right_speed):
             right_speed
         )
 
-    def on_for_rotations(self, left_speed, right_speed, rotations, brake=True, block=True):
```
dwalton76, line 1724: I moved this to after on_for_degrees() because on_for_rotations() call on_for_degrees()

ev3dev2/motor.py
```diff
@@ -1927,6 +2059,183 @@ def get_speed_steering(self, steering, speed):
...
+        if abort:
+            self._set_brake(brake)
+            return
+
+        rotations = distance_mm / self.wheel.circumference_mm
+        log.debug("%s: on_for_rotations distance_mm %s, rotations %s, speed %s" % (self, distance_mm, rotations, speed))
+
+        MoveTank.on_for_rotations(self, speed, speed, rotations, brake, block)
+
+    # Distance here could be
```
dwalton76: I need to chop this comment...right now distance has to be in mm.

ev3dev2/wheel.py
```diff
@@ -0,0 +1,27 @@
...
+    """
+
+    def __init__(self, diameter_mm, width_mm):
+        self.diameter_mm = float(diameter_mm)
+        self.width_mm = float(width_mm)
+        self.radius_mm = float(diameter_mm / 2)
+        self.circumference_mm = diameter_mm * pi
+
+
+class EV3RubberWheel(Wheel):
```
dwalton76: I think we should add some other commonly used wheels but will do that via another pull request

utils/move_differential.py
```diff
@@ -0,0 +1,56 @@
...
+from ev3dev2.motor import OUTPUT_A, OUTPUT_B, MoveDifferential, SpeedRPM
+from ev3dev2.wheel import EV3RubberWheel
+from math import pi
+import argparse
+import logging
+import sys
+
+# command line args
+'''
+parser = argparse.ArgumentParser(description="Used to adjust the position of a motor in an already assembled robot")
```
dwalton76: will chop this commented out code

ev3dev2/motor.py
ddemidov: Just nitpicking here: a product of two speed values is no longer a speed value, it has units of [m^2]/[s^2].

ev3dev2/motor.py
ddemidov: See related discussion here: https://github.com/ev3dev/ev3dev-lang-python/pull/500#discussion_r214096109

ev3dev2/motor.py
```diff
@@ -102,17 +117,27 @@ class SpeedNativeUnits(SpeedValue):
...
 
+    def __mul__(self, other):
+
+        if isinstance(other, SpeedNativeUnits):
+            return SpeedNativeUnits(self.native_counts * other.native_counts)
+        elif isinstance(other, (float, int)):
+            return SpeedNativeUnits(self.native_counts * other)
+        else:
+            raise Exception("{} can only be multiplied by a SpeedNativeUnits, int, or float".format(self))
+
```
ddemidov, line 128: This block is duplicated basically unchanged below. Can we move it to the base class by changing the first `if` to
```
if isinstance(other, self.__class__):
```
?

ev3dev2/motor.py
ddemidov: Why not use `Motor.reset()` instead?

ev3dev2/motor.py
ddemidov: Makes sense. Same applies to the `off()` method below.

ev3dev2/wheel.py
ddemidov: Are rubber wheels from home edition and from education edition have the same diameter and width?

ev3dev2/wheel.py
```diff
@@ -0,0 +1,27 @@
...
+class Wheel(object):
+    """
+    A base class for various types of wheels, tires, etc
+    All units are in mm
+    """
+
+    def __init__(self, diameter_mm, width_mm):
+        self.diameter_mm = float(diameter_mm)
+        self.width_mm = float(width_mm)
+        self.radius_mm = float(diameter_mm / 2)
```
ddemidov: Why do we need both radius and diameter?

ev3dev2/motor.py
ddemidov: Also, it looks like failing tests on Travis may be caused by this change.

ev3dev2/motor.py
dwalton76: I will simplify this and only support multiplying by int and float I can’t think of a good use case for anything more complicated than that.

ev3dev2/wheel.py
WasabiFan: I recall there being different rubber tires and associated rims which looked similar but had different diameters. I think there was one inner rim with three peg holes for a special axle mount piece and a different one which was a bit smaller and lacked them. So maybe we need particular part numbers here.

ev3dev2/motor.py
dwalton76: ah I missed those round() calls...mine are not needed will chop them.

ev3dev2/__init__.py
WasabiFan: Could you add that as a comment?

ev3dev2/motor.py
dwalton76: Will do that

ev3dev2/wheel.py
dwalton76: Not sure what it hurts?

ev3dev2/motor.py
dwalton76: The problem is each child class has its own unique variable that must be multiplied.

ev3dev2/__init__.py
dwalton76: will do

ev3dev2/wheel.py
dwalton76: I will dig up the part number and put it in the docstring for this class

ev3dev2/wheel.py
ddemidov: One of those could be a property, so we would not need to duplicate information and keep the fields in sync.

ev3dev2/motor.py
```diff
@@ -1638,11 +1699,26 @@ def reset(self, motors=None):
             motor.reset()
 
     def stop(self, motors=None):
+        """
+        stop the motors without setting the stop_action (brake, hold, coast)
```
WasabiFan: Could you explain what this actually means to someone who doesn't know how `stop_action`s work?

ev3dev2/motor.py
```diff
@@ -1638,11 +1699,26 @@ def reset(self, motors=None):
...
+        """
         motors = motors if motors is not None else self.motors.values()
 
         for motor in motors:
             motor.stop()
 
+    def off(self, brake=True, motors=None):
+        """
+        set the stop_action prior to stopping the motors
```
WasabiFan: Same as previous comment

ev3dev2/motor.py
```diff
@@ -1765,12 +1839,20 @@ def on_for_degrees(self, left_speed, right_speed, degrees, brake=True, block=Tru
         # therefore, t = d_l/v_l = d_r/v_r
         
         if degrees == 0 or (left_speed_native_units == 0 and right_speed_native_units == 0):
+            log.warning("{}: degrees {}, left_speed_native_units {}, right_speed_native_units {}, robot will not move"
+                .format(self, degrees, left_speed_native_units, right_speed_native_units))
+            left_degrees = 0
+            right_degrees = 0
+
+        elif left_speed_native_units == right_speed_native_units:
```
WasabiFan: What's happening here? It looks like the logic is different now.

ev3dev2/motor.py
```diff
@@ -1781,13 +1863,47 @@ def on_for_degrees(self, left_speed, right_speed, degrees, brake=True, block=Tru
         self.right_motor._set_rel_position_degrees_and_speed_sp(right_degrees, right_speed_native_units)
         self.right_motor._set_brake(brake)
 
+        left_motor_runtime_eta = float(left_degrees / left_speed_native_units)
```
WasabiFan: `(degrees) / (native units per second) != time`; what if native units aren't degrees?

ev3dev2/motor.py
```diff
@@ -1781,13 +1863,47 @@ def on_for_degrees(self, left_speed, right_speed, degrees, brake=True, block=Tru
...
+        right_motor_runtime_eta = float(right_degrees / right_speed_native_units)
+        left_motor_position_init = self.left_motor.position
+        right_motor_position_init = self.right_motor.position
+
+        log.debug("{}: on_for_degrees {}".format(self, degrees))
+        log.debug("{}: left_speed {}, left_speed_native_units {}, left_degrees {}, left-position {}->{}, ETA {}s".format(
```
WasabiFan: Do we have any idea if debug logging has a perf impact?

ev3dev2/motor.py
```diff
@@ -1803,6 +1919,9 @@ def on_for_seconds(self, left_speed, right_speed, seconds, brake=True, block=Tru
         self.right_motor.time_sp = int(seconds * 1000)
         self.right_motor._set_brake(brake)
 
+        log.debug("%s: on_for_seconds %ss at left-speed %s, right-speed %s" %
```
WasabiFan, line 1872: If we're using debug logging more widely can you add some info somewhere (README? Docs?) saying how to enable the output?

ev3dev2/motor.py
```diff
@@ -1924,6 +2037,181 @@ def get_speed_steering(self, steering, speed):
         return (left_speed, right_speed)
 
 
+class MoveDifferential(MoveTank):
+    """
+    For all methods speed can be an integer or a SpeedInteger object
```
WasabiFan: Can you add a full documentation comment here explaining what the class is?

ev3dev2/motor.py
```diff
@@ -1924,6 +2037,181 @@ def get_speed_steering(self, steering, speed):
...
+    """
+
+    def __init__(self, left_motor_port, right_motor_port, wheel_class, wheel_distance_mm, desc=None, motor_class=LargeMotor):
```
WasabiFan: These params will need some doc comments

ev3dev2/motor.py
```diff
@@ -1924,6 +2037,181 @@ def get_speed_steering(self, steering, speed):
...
+        MoveTank.__init__(self, left_motor_port, right_motor_port, desc, motor_class)
+        self.wheel = wheel_class()
+
+        # The distance between the mid point of the wheels.  You may need to
+        # do some test drives to find the correct value for your robot.  It
+        # is not as simple as measuring the distance between the midpoints of
+        # the two wheels. The weight of the robot, center of gravity, etc come
+        # into play.  You can use utils/move_differential.py to call on_arc_left()
+        # to do some test drives in 1-foot circles.  Tweak wheel_distam_mm until
```
WasabiFan: Using a 1-foot circle to find dimensions in millimeters? 🤔

ev3dev2/motor.py
```diff
@@ -1924,6 +2037,181 @@ def get_speed_steering(self, steering, speed):
...
+        self.wheel_distance_mm = wheel_distance_mm
+
+        # The circumference of the circle made if this robot were to rotate in place
+        self.circumference_mm = self.wheel_distance_mm * pi
+
+        self.min_circle_radius_mm = self.wheel_distance_mm / 2
+
+    def on_for_distance(self, speed, distance_mm, brake=True, block=True):
+        """
+        Drive forward distance_mm
```
WasabiFan: Forward? How do negative speeds/distances interact?

ev3dev2/motor.py
```diff
@@ -1924,6 +2037,181 @@ def get_speed_steering(self, steering, speed):
...
+
+    def _on_arc(self, speed, radius_mm, distance_mm, brake, block, arc_right):
+        """
+        Drive in a circle with 'radius' for 'distance'
+        """
+        abort = False
+
+        if not speed:
+            log.warning("{}: speed is invalid ({}), tank will not move".format(self, speed))
+            abort = True
```
WasabiFan: Do we really need the additional machinery for checking validity then aborting? I think it's fine to get one error then to fix that and get another if needed.

ev3dev2/motor.py
```diff
@@ -1924,6 +2037,181 @@ def get_speed_steering(self, steering, speed):
...
+            left_speed = float(circle_inner_mm/circle_outer_mm) * right_speed
+
+        log.debug("%s: arc %s, radius %s, distance %s, left-speed %s, right-speed %s, circle_outer_mm %s, circle_middle_mm %s, circle_inner_mm %s" %
+            (self, "right" if arc_right else "left",
+             radius_mm, distance_mm, left_speed, right_speed,
+             circle_outer_mm, circle_middle_mm, circle_inner_mm
+            )
+        )
+
+        if distance_mm is None:
```
WasabiFan: This feels spooky... at a minimum we need to heavily document it. Is there another reasonable way to structure a "run infinitely" request?

ev3dev2/motor.py
```diff
@@ -1924,6 +2037,181 @@ def get_speed_steering(self, steering, speed):
...
+            log.debug("%s: arc %s, circle_middle_percentage %s, circle_outer_final_mm %s, outer_wheel_rotations %s, outer_wheel_degrees %s" %
+                (self, "right" if arc_right else "left",
+                 circle_middle_percentage, circle_outer_final_mm,
+                 outer_wheel_rotations, outer_wheel_degrees
+                )
+            )
+
+            MoveTank.on_for_degrees(self, left_speed, right_speed, outer_wheel_degrees, brake, block)
+
+    def on_arc_right(self, speed, radius_mm, distance_mm, brake=True, block=True):
```
WasabiFan, line 2135: I'm having a bit of trouble understanding the use case for the API being structured like this (having separate "left" and "right" versions with radii), could you give some examples?

ev3dev2/motor.py
```diff
@@ -1924,6 +2037,181 @@ def get_speed_steering(self, steering, speed):
...
+
+        # If degrees is negative rotate counter-clockwise
+        else:
+            rotations = distance_mm / self.wheel.circumference_mm
+            MoveTank.on_for_rotations(self, speed * -1, speed, rotations, brake, block)
+
+    def turn_right(self, speed, degrees, brake=True, block=True):
+        self._turn(speed, abs(degrees), brake, block)
+
+    def turn_left(self, speed, degrees, brake=True, block=True):
```
WasabiFan, line 2176: doc comments?

ev3dev2/wheel.py
```diff
@@ -0,0 +1,60 @@
...
+
+A great reference when adding new wheels is http://wheels.sariel.pl/
+"""
+from math import pi
+
+
+class Wheel(object):
+    """
+    A base class for various types of wheels, tires, etc
+    All units are in mm
```
WasabiFan: Docs on what to do with this?

ev3dev2/motor.py
dwalton76: How about:

Stop the motors without specifying the “stop action”.  “stop actions” are one of the following 
- coast to a stop
- apply the brake to stop but do not hold the motor in position once stopped
- apply the brake to stop and hold the motor in position once stopped

ev3dev2/motor.py
dwalton76: I am going to chop the ETA debugs it feels like overkill.

ev3dev2/motor.py
dwalton76: I was just trying to simplify the code for the scenario where the left speed and right speed are equal. The end behavior should be the same as before

ev3dev2/motor.py
dwalton76: I have tried to measure its impact in the past but the delta was so small I couldn’t really measure it. Unless you do something crazy like drop into a loop and call log.debug 1000x I don’t think you will ever notice an impact.

ev3dev2/motor.py
dwalton76: Yeah will write something up.

ev3dev2/motor.py
dwalton76: A 1-meter circle felt like overkill :) They can drive in whatever size circle they want, they just need to tweak the wheel_distance until it makes an exact circle.

ev3dev2/motor.py
dwalton76: I will drop the “forward”. Negative speeds and distances should be fine but I will test to be sure.

ev3dev2/motor.py
dwalton76: Honestly I would be fine with just a bunch of asserts here but that means we don’t call set_brake when something goes wrong. I could live with that though. Thoughts?

dwalton76: There are other places where we only have asserts and others where we log a warning and call set_brake.  We should make these consistent. My vote would be for asserts everywhere.

ev3dev2/motor.py
dwalton76: We could chop it? The only thing it buys you is the ability to drive in a circle for forever...not sure how useful that would be. They could always set the distance to some huge number to get the same end result.

ev3dev2/motor.py
dwalton76: Maybe rename them to
- arc_clockwise
- arc_counter_clockwise

would be better?

ev3dev2/motor.py
```diff
@@ -1638,11 +1699,26 @@ def reset(self, motors=None):
...
     def stop(self, motors=None):
+        """
+        stop the motors without setting the stop_action (brake, hold, coast)
+        """
         motors = motors if motors is not None else self.motors.values()
 
         for motor in motors:
             motor.stop()
 
+    def off(self, brake=True, motors=None):
```
WasabiFan: Now that I look at this... I think having "stop" and "off" is confusing. Can we just have one with a keyword arg for a stop action?

ev3dev2/motor.py
WasabiFan: The behavior here is that it will use the last-used `stop_action`, right? I think that's the key here. And honestly, I'm slightly worried about the fact that this is public API surface area if its purpose isn't necessarily clear.

ev3dev2/motor.py
WasabiFan: OK, if the behavior should be the same I'll take your word for it

ev3dev2/motor.py
WasabiFan: On the EV3? It sounds believable either way, so I'm fine.

ev3dev2/motor.py
WasabiFan: Ah, OK. That sounds good to me.

ev3dev2/motor.py
dwalton76: yep I tested on EV3 back when I looked into it

ev3dev2/motor.py
WasabiFan: You could create a utility method which does an assertion and stops the motors if the assert fails.

I like the behavior where it stops the motor if an error occurs, but maybe it isn't necessary and we can go for the simple option instead.

ev3dev2/motor.py
WasabiFan: Yeah, I vote we chop it. There isn't really a reason to drive in a circle forever 😆

ev3dev2/motor.py
dwalton76: I tend to agree. We had them before though, I was just moving them up to the MotorSet class. How about we keep stop and chop off?

ev3dev2/motor.py
WasabiFan: What I'm trying to figure out is... what is the use case for being able to _arc with a radius_ for the purposes of driving? What cases do you know the radius and want to drive that circle?

Also, this brings up a related point: We have separate classes for "tank", "steering", and other drive patterns, but there isn't explicitly a reason that one needs to use _just tank_ or _just steering_. Certainly in some scenarios it makes sense, but if I'm trying to drive a particular series of lines and curves, it would be nice to be able to mix the various forms. I am pretty sure that works with our existing structure but it doesn't feel like it was designed to do so.

ev3dev2/motor.py
dwalton76: Right...which could lead to some unexpected behavior. What do you say we
- delete off()
- stop() needs to accept an arg to specify the stop_action. I think we just have to document the heck out of the three options.

ev3dev2/motor.py
dwalton76: There are probably a ton of places where we could raise an exception without first stopping the motors :( Let’s just do asserts and keep it nice and simple. The user will pretty always need their program to catch exceptions and stop the motors as a result.

ev3dev2/motor.py
dwalton76: I could see it being applicable in FLL where you want to drive in a partial circle to get around some object. Also the ability to drive in a circle of a specific radius was one of the main asks in the issue that was opened. 

I think MoveTank and MoveSteering are separate because EV3-G has these as separate blocks? Maybe we can combine these into MoveTank and have MoveSteering be a child class that doesn’t change anything? I need to look through exactly what MoveSteering does though. 

Heck could we also put all of this MoveDifferential code into MoveTank? Seems like we could

ev3dev2/motor.py
dwalton76: Filed #540 for the off() part

ev3dev2/motor.py
dwalton76: Filed #540 for this

ev3dev2/motor.py
dwalton76: Will tackle this via #540

ev3dev2/motor.py
```diff
@@ -890,8 +941,8 @@ def on_for_rotations(self, speed, rotations, brake=True, block=True):
         ``speed`` can be a percentage or a :class:`ev3dev2.motor.SpeedValue`
         object, enabling use of other units.
         """
-        if speed is None or rotations is None:
-            raise ValueError("Either speed ({}) or rotations ({}) is None".format(self, speed, rotations))
+        assert speed, "{}: speed is invalid ({}), motor will not move".format(self, speed)
```
WasabiFan:
```
>>> assert 0
AssertionError
```

ev3dev2/motor.py
```diff
@@ -1055,6 +1095,9 @@ class MediumMotor(Motor):
 
     def __init__(self, address=None, name_pattern=SYSTEM_DEVICE_NAME_CONVENTION, name_exact=False, **kwargs):
 
+        if platform in ('brickpi', 'brickpi3'):
+            raise Exception("%s is unaware of different motor types, use LargeMotor instead" % platform)
```
WasabiFan: Why can't someone set the port driver to the medium motor class explicitly and use it like a medium motor?

ev3dev2/motor.py
```diff
@@ -1681,21 +1731,37 @@ def wait(self, cond, timeout=None, motors=None):
     def wait_until_not_moving(self, timeout=None, motors=None):
         motors = motors if motors is not None else self.motors.values()
 
+        # sort the motors so the robot always waits on the motors in the
+        # same order everytime it runs. This makes troubleshooting easier.
+        motors = sorted(list(motors))
```
WasabiFan: You can't change the list of motors after construction, right? So how about we sort it in the constructor instead?

ev3dev2/motor.py
```diff
@@ -1781,13 +1837,33 @@ def on_for_degrees(self, left_speed, right_speed, degrees, brake=True, block=Tru
         self.right_motor._set_rel_position_degrees_and_speed_sp(right_degrees, right_speed_native_units)
         self.right_motor._set_brake(brake)
 
+        log.debug("{}: on_for_degrees {}".format(self, degrees))
+        log.debug("{}: left_speed {}, left_speed_native_units {}, left_degrees {}, left-position {}->{}".format(
```
WasabiFan: OK... another perf-related comment. Even if the debug log output is disabled, this will still be evaluated, and therefore I believe we are introducing file I/O for the `position` and `position__sp` properties just to throw out their values. I'm fine with the logging calls in general, but unnecessary file I/O is an issue for me.

ev3dev2/motor.py
```diff
@@ -1920,10 +2003,174 @@ def get_speed_steering(self, steering, speed):
             right_speed *= speed_factor
         else:
             left_speed *= speed_factor
-        
+
         return (left_speed, right_speed)
 
 
+class MoveDifferential(MoveTank):
```
WasabiFan, line 2002: The other classes inherit from MoveTank, but honestly I think I would prefer composition over inheritence. This is a peer to MoveTank so while it seems reasonable to use that other class's behavior I think any common stuff should be either re-exposed manually or moved up one level in the tree. The inheritance makes documentation pretty tough.

ev3dev2/motor.py
```diff
@@ -1920,10 +2003,174 @@ def get_speed_steering(self, steering, speed):
...
+
+    - rotate in place in a circle (clockwise or counter clockwise) for a
+      specified number of degrees
+
+    - drive in an arc (clockwise or counter clockwise) of a specified radius
+      for a specified distance
+
+    New arguements:
+
+    wheel_class - A class typically from ev3dev2/wheel.py. This is used to
```
WasabiFan: I think we need a code example including imports at the bottom of this comment (see other classes for code formatting) to provide a copy-pastable example.

ev3dev2/wheel.py
```diff
@@ -0,0 +1,67 @@
...
+    of the robot.
+    """
+
+    def __init__(self, diameter_mm, width_mm):
+        self.diameter_mm = float(diameter_mm)
+        self.width_mm = float(width_mm)
+        self.circumference_mm = diameter_mm * pi
+
+    @property
+    def radius_mm(self):
```
WasabiFan, line 38: Can you add a docs page so that these classes show up? I think we can follow the model from the unit section of the motor page (https://python-ev3dev.readthedocs.io/en/ev3dev-stretch/motors.html#units).

tests/api_tests.py
```diff
@@ -162,29 +162,6 @@ def test_motor_on_for_degrees(self):
         self.assertEqual(m.speed_sp, int(round(0.75 * 1050)))
         self.assertEqual(m.position_sp, 100)
 
-        # zero speed (on-device, this will return immediately due to reported stall)
-        m.on_for_degrees(0, 100)
```
WasabiFan: Why are these tests being deleted?

utils/move_differential.py
```diff
@@ -0,0 +1,46 @@
+#!/usr/bin/env python3
+
+"""
+Used to adjust the position of a motor in an already assembled robot
+where you can"t move the motor by hand.
```
WasabiFan: copy-paste error?

ev3dev2/motor.py
```diff
@@ -1920,10 +2003,174 @@ def get_speed_steering(self, steering, speed):
...
+
+    - drive in a straight line for a specified distance
+
+    - rotate in place in a circle (clockwise or counter clockwise) for a
+      specified number of degrees
+
+    - drive in an arc (clockwise or counter clockwise) of a specified radius
+      for a specified distance
+
+    New arguements:
```
WasabiFan: spelling

ev3dev2/motor.py
dwalton76: Not sure what you are asking here?

WasabiFan: If speed is zero it looks like this assertion will fail; previously if speed was zero the error _would not_ be thrown. Same with rotations.

ev3dev2/motor.py
dwalton76: Is that doable? I have never tried it. I added the snippet above after trying to use a medium motor on brickpi3 but hit issues and dlech mentioned that it cannot detect different motor types

WasabiFan: I don't know for sure. My assumption was that you could manually set the driver and it would work, it was just that by default the ports were all initialized as large motors for simplicity.

ev3dev2/motor.py
dwalton76: We could do it in the constructor if we make self.motors an OrderedDict. I like the thought of that, then the order will be consistent everywhere.

ev3dev2/motor.py
dwalton76: Fair point on the I/O involved with getting the position. I’ll comment these out that way we can bring them back easily enough if we need them for troubleshooting

ev3dev2/motor.py
dwalton76: What do you say we tackle how to organize these as part of the issue regarding a potential drive.py?

ev3dev2/wheel.py
dwalton76: Is there a howto on writing docs? It has been ages since I have tried to edit them. What repo are they in?

tests/api_tests.py
dwalton76: They are doing things like passing in speeds and degrees with values of 0 which now are caught by asserts.

utils/move_differential.py
dwalton76: LOL yeah will fix

ev3dev2/motor.py
dwalton76: This is basically MoveTank but with known wheel circumference and wheel distance which makes more things possible. It is adding to MoveTank but can still use all of the functionality provided by MoveTank. To me this is very much a child class not a peer class.

ev3dev2/motor.py
dwalton76: Right but this is part of what we discussed yesterday, we were all over the place how to handle 0 values for speed, etc. We did a mix of
- used an assert
- logged a warning  and stopped the motors
- not check at all for 0 values

I thought we agreed to use asserts so I made them all consistent to use asserts?

ev3dev2/motor.py
dwalton76: I think if someone wants to test/document that then we can drop the exception at that point.

ev3dev2/motor.py
WasabiFan: Zero speed/distance are a supported scenario; there was a fair amount of discussion on the topic in https://github.com/ev3dev/ev3dev-lang-python/pull/523.

ev3dev2/motor.py
WasabiFan: I would prefer to not introduce the exception given that it isn't an error case in existing releases. If someone depends on this class, I don't see a reason to break them.

ev3dev2/motor.py
dwalton76: What is the desired behavior...log a warning and return? I’m fine either way we just need to make it consistent

ev3dev2/wheel.py
WasabiFan: There are files in this repo's `docs/` which declare the top-level structure. Many of the sections just have a tag to auto-gen docs from doc comments of a class. You can preview by either manually running Sphinx locally or enabling the branch as a private version in RTD.

ev3dev2/motor.py
WasabiFan: The existing tests document the expected behavior in all scenarios. In general a `0` causes a no-op — it's fine to ask to go nowhere. No error is logged because it isn't a failing scenario.

ev3dev2/motor.py
dwalton76: My memory was rusty but after going back and re-reading https://github.com/ev3dev/ev3dev/issues/1059 I do not think "set the port driver to the medium motor class explicitly" will work (see the first comment in that issue). So it is an error case today if the user attempts to use the MediumMotor class on a brickpi3.  

This change isn't needed as part of MoveDifferential though so I opened #543 to track it....will drop this part of the diff from the patch here.

dwalton76:
@WasabiFan I think I have most things addressed besides the needed docstrings updates...will tackle those tonight.

Using collections.OrderedDict causes the micropython tests to fail, it does not have "collections".  How have we been handling this scenario?  I could catch it and use a regular dict instead but then we do not have the sorting we want.  Another way we could handle this would be to go back to a regular dict but also create sorted list of motors in __init__.  We would then use this sorted list anytime we need to loop over the list of motors.

ddemidov:
> Using collections.OrderedDict causes the micropython tests to fail, it does not have "collections".

http://docs.micropython.org/en/v1.9.3/pyboard/library/ucollections.html

ev3dev2/wheel.py
dwalton76: I checked something in but I am not familiar with Sphinx or RTD (Real Time Documentation?) so I do not know how to check the end result.  Unfortunately googling for "ev3dev sphinx" gets a hit on every page of our docs due to the "Built with Sphinx using..." text at the bottom.

dwalton76: ok I initially created a wheels.rst but then the build failed because that was never referenced from anywhere (doh!).  I started looking at where it should be referenced and then realized that all of the these wheel docs are probably a better fit in others.rst instead of being a standalone rst file.

docs/other.rst
```diff
@@ -141,3 +141,38 @@ Lego Port
 
 .. autoclass:: ev3dev2.port.LegoPort
     :members:
+
+Wheels
+------
+All Wheel class units are in millimeters.  The ``diameter_mm`` and ``width_mm`` of the wheel must be specified whenever a new child Wheel class is added to ``ev3dev2/wheel.py``. The diameter and width for various lego wheels can be found at http://wheels.sariel.pl/
```
WasabiFan: Isn't the sentence about `wheel.py` an implementation detail that users don't need to know about?

ev3dev2/motor.py
```diff
@@ -30,8 +30,16 @@
 
 import select
 import time
+
+# pytyon3 uses collections
```
WasabiFan: typo

ev3dev2/motor.py
```diff
@@ -1920,10 +2002,191 @@ def get_speed_steering(self, steering, speed):
...
+
+    - rotate in place in a circle (clockwise or counter clockwise) for a
+      specified number of degrees
+
+    - drive in an arc (clockwise or counter clockwise) of a specified radius
+      for a specified distance
+
+    New arguments:
+
+    wheel_class - A class typically from ev3dev2/wheel.py. This is used to
```
WasabiFan: Link to wheel section in docs? There should be examples for motor units. IIRC you might need to manually add a tag to the docs page.

ev3dev2/motor.py
```diff
@@ -1696,6 +1758,10 @@ def wait_while(self, s, timeout=None, motors=None):
         for motor in motors:
             motor.wait_while(s, timeout)
 
+    def _block(self):
+        self.wait_until('running', timeout=WAIT_RUNNING_TIMEOUT)
```
WasabiFan, line 1751: What's the change here? There used to be a `_block` in `MoveTank` but now that seems to have been removed and this was added. But I don't see changes to the usage.

If this is intended to replace the batch version in `MoveTank`, I have some additional questions about potential impacts.

ev3dev2/motor.py
WasabiFan: OK, I see what you're saying; the `MoveTank` methods aren't being replaced here so you can still use them. I think it makes sense now. I was thinking this was more like the others but that's a different case.

docs/other.rst
```diff
@@ -141,3 +141,38 @@ Lego Port
 
 .. autoclass:: ev3dev2.port.LegoPort
     :members:
+
+Wheels
```
WasabiFan, line 145: In all honesty, I think the `other` docs page is pretty unweildy. We have so much stuff here that it just becomes difficult to browse. What do you think about breaking from tradition here and giving this its own file, and then we can track an issue for splitting out some of the other classes (like buttons) which are getting pretty big?

docs/other.rst
dwalton76: will chop that sentence

ev3dev2/motor.py
dwalton76: No change to the functionality this was just moving it up to the ``MotorSet`` class

docs/other.rst
dwalton76: Let's commit the wheels docs here for now and then we can figure out how we want to split this file up afterwards?

ev3dev2/motor.py
dwalton76: Given that none of the other docstrings (none of the ones that I looked at at least...Motor, ColorSensor, etc) are linking to their respective docs sections I would say lets leave it as is.  If we want to have docstrings include links to their docs sections lets do that via a separate issue.

dwalton76:
I think this one is ready to go.  Do you guys want me to squash the commits down into one big "MoveDifferential" commit or just merge this as is?

ev3dev2/motor.py
WasabiFan: The only other case where usage of one class depends on another is with units -- and we have links there. For example:

https://github.com/ev3dev/ev3dev-lang-python/blob/efe9a7398d666f46fda4a6c887bd8e03dfde130f/ev3dev2/motor.py#L890-L895

In these cases the user wants to have quick access to the relevant documentation. Remember, this is used both in docstrings _and_ online rendered docs pages.

ev3dev2/motor.py
WasabiFan: Can you expand on that?

It looks to me like the logic was changed from...

```
wait for motor A to start
wait for motor B to start
wait for motor A to stop
wait for motor B to stop
```

...to...

```
wait for motor A to start
wait for motor A to stop
wait for motor B to start
wait for motor B to stop
```

Which we've had issues with before.

dwalton76: The logic is still
```
wait for motor A to start
wait for motor B to start
wait for motor A to stop
wait for motor B to stop
```

``MotorSet.wait_until()`` loops over all of the motors so it will wait for A to start then wait for B to start...same basic logic for ``MotorSet.wait_until_not_moving``

ev3dev2/motor.py
dwalton76: will fix

docs/other.rst
dwalton76: Filed #559 to track this part

dwalton76:
Thank you for all of the feedback/conversation on this one.  This ended up being a fairly large patch in order to implement the MoveDifferential functionality.  At this point there haven't been any major changes to the patch in several days so I am going to go ahead and merge it....we can address any more tweaks via another commit.

Issue #543, opened 2018-10-30
brickpi3 and medium motors


dwalton76:
> The firmware in BrickPi3 (and earlier BrickPi models) doesn't know about different kinds of motors, so we just treat everything like an NXT motor.
-- dlech in https://github.com/ev3dev/ev3dev/issues/1059

The work-around is to use the LargeMotor class for the medium motor.  One potential problem is if the user's program relies on max_speed since that value is different for medium vs large motors. 

If the user does try to use the MediumMotor class we will fail to set the driver (see the intro in https://github.com/ev3dev/ev3dev/issues/1059).  One option for presenting the user with a more informed error message is to raise an exception if the MediumMotor class is used:

```
        if platform in ('brickpi', 'brickpi3'):
            raise Exception("%s is unaware of different motor types, use LargeMotor instead" % platform)
```

dwalton76:
Running
```
Image file:         ev3dev-stretch-rpi2-generic-2017-12-13
Kernel version:     4.14.71-ev3dev-2.3.0-rpi2
Brickman:           0.10.0
BogoMIPS:           38.40
Board:              board0
BOARD_INFO_HW_REV=A22082
BOARD_INFO_MODEL=Raspberry Pi 3 Model B Rev 1.2
BOARD_INFO_SERIAL_NUM=000000002f94f0a9
BOARD_INFO_TYPE=main
Board:              board1
BOARD_INFO_FW_VER=1.4.3
BOARD_INFO_HW_REV=3.2.1
BOARD_INFO_MODEL=Dexter Industries BrickPi3
BOARD_INFO_SERIAL_NUM=E1DA1CC95150324336202020FF19312F
BOARD_INFO_TYPE=aux
```

Testing with:
```
#!/usr/bin/env python3

from ev3dev2.motor import OUTPUT_A, LargeMotor, MediumMotor, SpeedRPM

mmotor = MediumMotor(OUTPUT_A)
#mmotor = LargeMotor(OUTPUT_A)
mmotor.on_for_seconds(SpeedRPM(40), 3)
```

If you try to use the MediumMotor class the brickpi thinks that it is not connected
```
robot@pi3[ev3dev-lang-python]# ./utils/foo.py
Traceback (most recent call last):
  File "./utils/foo.py", line 5, in <module>
    mmotor = MediumMotor(OUTPUT_A)
[...]

PR #546, opened 2018-10-31
brickpi always use LargeMotor driver

dwalton76:
For issue #543


dwalton76:
closing this, we are going to raise an Exception instead in this scenario

PR #548, opened 2018-10-31
wait_until_not_moving should consider 'running holding' as 'not moving'

### LINKED ISSUE #547: wait_until_not_moving should consider "running holding" as "not moving"
Details are in https://github.com/ev3dev/ev3dev/issues/1188 but in a nutshell when the motor has finished moving if the brake is being applied the state should be ``running holding``.  This is the case with brickpi3 but not EV3 (the EV3 behavior is the one that is incorrect).

``wait_until_not_moving`` needs to return if the state is ``running holding``

### LINKED ISSUE #558: blocking issue on motor runs on ev3dev2 (maybe ev3dev too)
- **ev3dev version:** 4.14.71-ev3dev-2.3.0-rpi2
- **ev3dev-lang-python version:** 
```
Desired=Unknown/Install/Remove/Purge/Hold
| Status=Not/Inst/Conf-files/Unpacked/halF-conf/Half-inst/trig-aWait/Trig-pend
|/ Err?=(none)/Reinst-required (Status,Err: uppercase=bad)
||/ Name                        Version            Architecture       Description
+++-===========================-==================-==================-===========================================================
ii  python3-ev3dev2             2.0.0~beta2        all                Python language bindings for ev3dev
```

So running m.on_for_degrees(-30, 90, block=True) works on the Lego brick (returning control when position is reached) and the same command on the Dexter brick doesn't return control. The code for repro is : 

```
from ev3dev2.motor import LargeMotor, OUTPUT_A, SpeedPercent
m = LargeMotor(OUTPUT_A)
m.on_for_degrees(SpeedPercent(10), 180, block = True)

```
This will be stuck forever on the dexter (polling issue?) but not on the lego brick. We can repro on 2 different Dexter devices with EV3 and NXT.

dwalton76:
Fixes #558
Fixes #547

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,5 +1,6 @@
 python-ev3dev2 (2.0.0~beta3) stable; urgency=medium
 
+  * wait_until_not_moving should consider "running holding" as "not moving"
   * MoveJoyStick should use SpeedPercent
   * renamed GyroSensor rate_and_angle to angle_and_rate
 
--- a/ev3dev2/motor.py
+++ b/ev3dev2/motor.py
@@ -813,10 +813,13 @@ class Motor(Device):
 
     def wait_until_not_moving(self, timeout=None):
         """
-        Blocks until ``running`` is not in ``self.state`` or ``stalled``
-        or ``holding` are in ``self.state``.  The condition is checked
-        when there is an I/O event related to the ``state`` attribute.
-        Exits early when ``timeout`` (in milliseconds) is reached.
+        Blocks until one of the following conditions are met:
+        - ``running`` is not in ``self.state``
+        - ``stalled`` is in ``self.state``
+        - ``holding`` is in ``self.state``
+        The condition is checked when there is an I/O event related to
+        the ``state`` attribute.  Exits early when ``timeout`` (in
+        milliseconds) is reached.
 
         Returns ``True`` if the condition is met, and ``False`` if the timeout
         is reached.
--- a/ev3dev2/motor.py
+++ b/ev3dev2/motor.py
@@ -813,10 +813,10 @@ class Motor(Device):
 
     def wait_until_not_moving(self, timeout=None):
         """
-        Blocks until ``running`` is not in ``self.state`` or ``stalled`` is in
-        ``self.state``.  The condition is checked when there is an I/O event
-        related to the ``state`` attribute.  Exits early when ``timeout``
-        (in milliseconds) is reached.
+        Blocks until ``running`` is not in ``self.state`` or ``stalled``
+        or ``holding` are in ``self.state``.  The condition is checked
+        when there is an I/O event related to the ``state`` attribute.
+        Exits early when ``timeout`` (in milliseconds) is reached.
 
         Returns ``True`` if the condition is met, and ``False`` if the timeout
         is reached.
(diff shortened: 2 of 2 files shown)


dwalton76:
For issue #547

ev3dev2/motor.py
```diff
@@ -813,10 +813,10 @@ def wait(self, cond, timeout=None):
 
     def wait_until_not_moving(self, timeout=None):
         """
-        Blocks until ``running`` is not in ``self.state`` or ``stalled`` is in
-        ``self.state``.  The condition is checked when there is an I/O event
-        related to the ``state`` attribute.  Exits early when ``timeout``
-        (in milliseconds) is reached.
+        Blocks until ``running`` is not in ``self.state`` or ``stalled``
+        or ``holding` are in ``self.state``.  The condition is checked
```
WasabiFan: Perhaps a bulleted list would make this more readable?

debian/changelog
```diff
@@ -1,5 +1,6 @@
 python-ev3dev2 (2.0.0~beta3) stable; urgency=medium
 
+  * wait_until_not_moving should consider "running holding" as "not moving"
```
WasabiFan, line 3: Given that most of these changes are by you but I'm probably going to be the one doing the package release (and hopefully I'll get some time to add some changes of my own), I think we need section headers with the name of the person who made the changes.

WasabiFan (approved):
Minor suggestions

debian/changelog
dwalton76: So something like
```
python-ev3dev2 (2.0.0~beta3) stable; urgency=medium

  * MoveJoyStick should use SpeedPercent
  * renamed GyroSensor rate_and_angle to angle_and_rate
 -- Daniel Walton <dwalton76@gmail.com>

  * foo
  * bar
 -- Kaelin Laundry <wasabifan@outlook.com>  Sat, 27 Oct 2018 21:18:00 -0700
```

I don't know if that falls into the standard debian changelog format (probably not) or if it matters if we deviate from the format.

Personally I am fine if my name is not associated with my commits in the changelog.  The commits in the git history will still show as being by me.  If I ever needed to reference work here when applying for a job, etc I would point people at the commit history since that is more detailed than the change log.

ev3dev2/motor.py
dwalton76: Will do
```
        Blocks until one of the following conditions are met:
        - ``running`` is not in ``self.state``
        - ``stalled`` is in ``self.state``
        - ``holding`` is in ``self.state``
        The condition is checked when there is an I/O event related to
        the ``state`` attribute.  Exits early when ``timeout`` (in
        milliseconds) is reached.
```

debian/changelog
dlech: Have a look back in history:

https://github.com/ev3dev/ev3dev-lang-python/blob/55f57ada176905a1db073423f428bd83dfd6b6ca/debian/changelog

dlech: And yes, format matters very much in this file.

dwalton76: Ack on format being important.  Lets do this...once this group of pull requests goes in I'll come back and update the changelog to add a ``[Daniel Walton]`` section for my commits.  That way I don't have to update 9 PRs.

WasabiFan: Sounds good.

Issue #534, opened 2018-10-05
wait_until_angle_changed_by(delta) is not direction-sensitive


ndward:
4.14.61-ev3dev-2.2.2-ev3
ii  python3-ev3dev 1.2.0        all          Python language bindings for ev3d
ii  python3-ev3dev 2.0.0~beta2  all          Python language bindings for ev3d

https://python-ev3dev.readthedocs.io/en/ev3dev-stretch/sensors.html#ev3dev2.sensor.lego.GyroSensor.wait_until_angle_changed_by

wait_until_angle_changed_by(delta): Wait until angle has changed by specified amount.

It's not clear from the above description whether this is direction-sensitive. For example, if I have delta = +90° (clockwise rotation) and the sensor is turned 100° counterclockwise, does that mean the angle has changed by delta?

My tests make me think that the direction does NOT matter, but quite often users will be looking for a turn _in a specific direction_, so I'm surprised the function cannot handle that. Maybe the function can be made direction-sensitive?

dwalton76:
The code is using the abs() of the delta so the direction does not matter.

WasabiFan:
@dwalton76 I think Nigel is suggesting that it would be useful to make it direction-sensitive. It seems reasonable to me that we could add an argument which skipped the `abs` call to enable that scenario.

PR #551, opened 2018-10-31
GyroSensor wait_until_angle_changed_by added direction_sensitive option

dwalton76:
For issue #534

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,5 +1,6 @@
 python-ev3dev2 (2.0.0~beta3) stable; urgency=medium
 
+  * GyroSensor wait_until_angle_changed_by added honor_direction option
   * MoveJoyStick should use SpeedPercent
   * renamed GyroSensor rate_and_angle to angle_and_rate
 
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,6 +1,6 @@
 python-ev3dev2 (2.0.0~beta3) stable; urgency=medium
 
-  * GyroSensor wait_until_angle_changed_by added honor_direction option
+  * GyroSensor wait_until_angle_changed_by added direction_sensitive option
   * Added MoveDifferential
   * Sound: removed play() in favor of play_file()
   * suppress unittest ResourceWarning messages
--- a/ev3dev2/sensor/lego.py
+++ b/ev3dev2/sensor/lego.py
@@ -626,16 +626,32 @@ class GyroSensor(Sensor):
         self._ensure_mode(self.MODE_GYRO_ANG)
         self._direct = self.set_attr_raw(self._direct, 'direct', 17)
 
-    def wait_until_angle_changed_by(self, delta):
+    def wait_until_angle_changed_by(self, delta, honor_direction=False):
         """
         Wait until angle has changed by specified amount.
+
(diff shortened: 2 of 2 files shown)


dwalton76:
Tested via
```
#!/usr/bin/env python3

from ev3dev2.sensor.lego import GyroSensor
from ev3dev2.auto import INPUT_1, INPUT_2

gyro = GyroSensor(INPUT_2)
print("{} angle {}".format(gyro, gyro.angle))
gyro.wait_until_angle_changed_by(-90, True)
print("{} angle {}".format(gyro, gyro.angle))
```

ev3dev2/sensor/lego.py
```diff
@@ -626,16 +626,32 @@ def reset(self):
         self._ensure_mode(self.MODE_GYRO_ANG)
         self._direct = self.set_attr_raw(self._direct, 'direct', 17)
 
-    def wait_until_angle_changed_by(self, delta):
+    def wait_until_angle_changed_by(self, delta, honor_direction=False):
```
WasabiFan: I think "honor" might be confusing to non-native English speakers; could we call it something else? Maybe `direction_sensitive`? `specific_direction`?

ev3dev2/sensor/lego.py
```diff
@@ -626,16 +626,32 @@ def reset(self):
...
         """
         assert self.mode in (self.MODE_GYRO_G_A, self.MODE_GYRO_ANG,
                              self.MODE_TILT_ANG),\
             'Gyro mode should be MODE_GYRO_ANG, MODE_GYRO_G_A or MODE_TILT_ANG'
         start_angle = self.value(0)
-        while abs(start_angle - self.value(0)) < delta:
-            time.sleep(0.01)
+
+        if honor_direction:
+            if delta > 0:
```
WasabiFan, line 645: This feels like a lot of duplicated code. Can we combine this to be one loop? I think this condition might work:

```python
while (
        (abs(self.value(0) - start_angle) < delta)  # the sensor still hasn't moved far enough... 
        or (honor_direction and copysign(delta, 1) == copysign(self.value(0) - start_angle, 1))) # … or, it has moved in the wrong direction
```


I'm not entirely sure my line break structure will fly with the Python interpreter, but you get the gist.

dwalton76: Personally I find the expanded/verbose model a little easier to follow :)

ev3dev2/sensor/lego.py
dwalton76: ``direction_sensitive`` sounds good

ev3dev2/sensor/lego.py
WasabiFan: I just want to remove the duplicated loop code... whether that's with a more clever loop condition or something else doesn't matter to me.

debian/changelog
```diff
@@ -1,5 +1,6 @@
 python-ev3dev2 (2.0.0~beta3) stable; urgency=medium
 
+  * GyroSensor wait_until_angle_changed_by added honor_direction option
```
WasabiFan: new parameter name

WasabiFan (approved):

Issue #532, opened 2018-10-05
play() and play_file() are almost identical. Do we need both?


ndward:
4.14.61-ev3dev-2.2.2-ev3
ii  python3-ev3dev 1.2.0        all          Python language bindings for ev3d
ii  python3-ev3dev 2.0.0~beta2  all          Python language bindings for ev3d

play() and play_file() are apparently identical except that one can set the volume with play_file(). Do we need both functions? Can one be suppressed? Is it clear which one should be used (it's not clear to me).

dwalton76:
Seems like we can combine them. #358 has the EV3-G blocks for sound.

PR #552, opened 2018-10-31
Sound: removed play() in favor of play_file()

dwalton76:
For issue #532

### DIFF
--- a/ev3dev2/sound.py
+++ b/ev3dev2/sound.py
@@ -261,7 +261,7 @@ class Sound(object):
         if not 0 < volume <= 100:
             raise ValueError('invalid volume (%s)' % volume)
 
-        if wav_file.endswith(".mp3"):
+        if not wav_file.endswith(".wav"):
             raise ValueError('invalid sound file (%s), only .wav files are supported' % wav_file)
 
         if not os.path.isfile(wav_file):
--- a/ev3dev2/sound.py
+++ b/ev3dev2/sound.py
@@ -98,7 +98,7 @@ class Sound(object):
         See `beep man page`_ and google `linux beep music`_ for inspiration.
 
         :param string args: Any additional arguments to be passed to ``beep`` (see the `beep man page`_ for details)
-        
+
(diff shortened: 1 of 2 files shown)


ev3dev2/sound.py
```diff
@@ -247,16 +247,27 @@ def play_note(self, note, duration, volume=100, play_type=PLAY_WAIT_FOR_COMPLETE
...
-        :param play_type: The behavior of ``play`` once playback has been initiated
+        :param play_type: The behavior of ``play_file`` once playback has been initiated
         :type play_type: ``Sound.PLAY_WAIT_FOR_COMPLETE``, ``Sound.PLAY_NO_WAIT_FOR_COMPLETE`` or ``Sound.PLAY_LOOP``
 
         :returns: When ``Sound.PLAY_NO_WAIT_FOR_COMPLETE`` is specified, returns the spawn subprocess from ``subprocess.Popen``; ``None`` otherwise
         """
+        if not 0 < volume <= 100:
+            raise ValueError('invalid volume (%s)' % volume)
+
+        if wav_file.endswith(".mp3"):
```
WasabiFan: Why not require that it _does_ end in `.wav` rather than _not ending_ in `.mp3`?

WasabiFan (commented):
I think there are code examples in this file and maybe in other docs pages which use the old method name; those also need to be updated. Looks good to me otherwise.

ev3dev2/sound.py
dwalton76: I was worried about breaking things if someone had a wav file that did not end in .wav. That would be pretty silly though...will change it to enforce .wav.

WasabiFan: I come from the Windows world where file extensions are used for a lot of logic, so maybe my POV is different. wav files without extensions might be common, I'm not sure.

WasabiFan:
I think example code needs to be updated too.

dwalton76:
Fixed the example in the docstring

Issue #544, opened 2018-10-30
Motors and zero speed, degrees, rotations, etc


dwalton76:
The Motor, MoveTank, etc classes are not consistent in how they are handling requests to move at speed 0, degrees 0, etc.  Some raise an exception, some use asserts, some log a warning and call set_brake(), etc.  We need to decide how we want to handle these scenarios and make them all consistent.

Motor.on_for_rotations
- raises ValueError is speed or rotations is None

Motor.on_for_degrees
- raises ValueError is speed or degrees is None

Motor.on_to_position
- assert speed

Motor.on_for_seconds
- if not speed or not seconds we log a warning, call set_brake(), then return

Motor.on
- if not speed we log a warning, call set_brake(), then return

MoveTank.on_for_degrees
- if degrees or speeds are 0 we log a warning but do not return, we execute the rest of the method

MoveSteering.get_speed_steering
- assert steering >= -100 and steering <= 100

MoveDifferential.on_for_distance
- if not speed or not distance_mm log a warning, call set_brake() and return

MoveDifferential._on_arc
- if not speed, not distance_mm or not radius_mm log a warning, call set_brake() and return

MoveDifferential._turn
- if not speed or not degrees log a warning, call set_brake() and return

WasabiFan:
I think the points here apply: https://github.com/ev3dev/ev3dev-lang-python/pull/523#discussion_r217879269

My vote is to follow that behavior: if speed is zero, the call no-ops (no error, no warning; it may or may not command the motor, depending on how the driver handles it).

Values like radius and steering are separate and exceptions should be thrown if they are invalid.

dwalton76:
That works for me.  Just to clarify how I am interpreting this:
- speed, degrees, rotations and seconds should not have any sanity checks applied, just set them and let the driver do as instructed the best it can.
- invalid steering or invalid radius should raise an Exception. Any others?

WasabiFan:
I think one thing to add is that our code should step in if the driver behavior is bad. In the case of moving no distance, the motor drivers did not handle it well, which spawned https://github.com/ev3dev/ev3dev/issues/1170 and https://github.com/ev3dev/ev3dev/issues/1173. So, if sending a zero freezes your code, we need to handle it; if it just immediately returns we're OK.

dwalton76:
ok for the ones that fall into this category how do we want to handle them?
- silently return?
- log a warning and return?
- log a warning, call set_brake() and return?
[...]

PR #553, opened 2018-11-01
motor: remove speed 0, degrees 0, etc checks

dwalton76:
For issue #544

### DIFF
--- a/ev3dev2/motor.py
+++ b/ev3dev2/motor.py
@@ -890,9 +890,6 @@ class Motor(Device):
         ``speed`` can be a percentage or a :class:`ev3dev2.motor.SpeedValue`
         object, enabling use of other units.
         """
-        if speed is None or rotations is None:
-            raise ValueError("Either speed ({}) or rotations ({}) is None".format(self, speed, rotations))
-
         speed_sp = self._speed_native_units(speed)
         self._set_rel_position_degrees_and_speed_sp(rotations * 360, speed_sp)
         self._set_brake(brake)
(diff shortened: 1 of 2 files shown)


dwalton76:
These two tests failed
```
======================================================================
ERROR: test_motor_on_for_degrees (__main__.TestAPI)
----------------------------------------------------------------------
  File "tests/api_tests.py", line 182, in test_motor_on_for_degrees
    m.on_for_degrees(None, 100)
  File "tests/../ev3dev2/motor.py", line 909 in on_for_degrees
    speed_sp = self._speed_native_units(speed)
  File "tests/../ev3dev2/motor.py", line 864, in _speed_native_units
    assert -100 <= speed <= 100,\
TypeError: unorderable types: int() <= NoneType()
======================================================================
ERROR: test_motor_on_for_rotations (__main__.TestAPI)
----------------------------------------------------------------------
  File "tests/api_tests.py", line 201, in test_motor_on_for_rotations
    m.on_for_rotations(None, 5)
  File "tests/../ev3dev2/motor.py", line 893, in on_for_rotations
    speed_sp = self._speed_native_units(speed)
  File "tests/../ev3dev2/motor.py", line 864, in _speed_native_units
    assert -100 <= speed <= 100,\
TypeError: unorderable types: int() <= NoneType()
```
Do we officially support accepting None for speed, rotations, degrees, or seconds?  I did not test these when I was running through all the permutations.

WasabiFan:
I don't see a reason to support it. That being said, it certainly is nice when code tells you "None isn't a valid parameter value" when you make a mistake rather than "unorderable types: int() <= NoneType()" 😆

dwalton76:
> I don't see a reason to support it. That being said, it certainly is nice when code tells you "None isn't a valid parameter value" when you make a mistake rather than "unorderable types: int() <= NoneType()" 😆

We do not check/test for other unsupported types though (strings for instance)...any objections to just chopping these two test cases?

WasabiFan:
I think it's fine to remove them.

Issue #533, opened 2018-10-05
Incorrect documentation for rectangle()


ndward:
4.14.61-ev3dev-2.2.2-ev3
ii  python3-ev3dev 1.2.0        all          Python language bindings for ev3d
ii  python3-ev3dev 2.0.0~beta2  all          Python language bindings for ev3d

At this location:
https://python-ev3dev.readthedocs.io/en/ev3dev-stretch/other.html#ev3dev2.display.Display.rectangle
the description 

rectangle(clear_screen=True, **x=10, y=10, width=80, height=40**, fill_color='black', outline_color='black')

does not properly describe the function.  It behaves like 

rectangle(clear_screen=True, **x1=10, y1=10, x2=80, y2=40**, fill_color='black', outline_color='black')

PR #557, opened 2018-11-02
Display: correct xy variable names passed to Display.rectangle()

dwalton76:
For issue #533

### DIFF
--- a/ev3dev2/display.py
+++ b/ev3dev2/display.py
@@ -347,7 +347,7 @@ class Display(FbMem):
         if clear_screen:
             self.clear()
 
-        return self.draw.rectangle((x, y, width, height), fill=fill_color, outline=outline_color)
+        return self.draw.rectangle((x, y, x+width, y+height), fill=fill_color, outline=outline_color)
 
     def point(self, clear_screen=True, x=10, y=10, point_color='black'):
         """
--- a/ev3dev2/display.py
+++ b/ev3dev2/display.py
@@ -339,15 +339,16 @@ class Display(FbMem):
 
         return self.draw.ellipse((x1, y1, x2, y2), fill=fill_color, outline=outline_color)
 
-    def rectangle(self, clear_screen=True, x=10, y=10, width=80, height=40, fill_color='black', outline_color='black'):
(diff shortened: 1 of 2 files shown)


ev3dev2/display.py
```diff
@@ -347,7 +347,7 @@ def rectangle(self, clear_screen=True, x=10, y=10, width=80, height=40, fill_col
         if clear_screen:
             self.clear()
 
-        return self.draw.rectangle((x, y, width, height), fill=fill_color, outline=outline_color)
+        return self.draw.rectangle((x, y, x+width, y+height), fill=fill_color, outline=outline_color)
```
WasabiFan: So we're calling this a code bug rather than a documentation issue? I don't have anything against that, but it's worth keeping in mind that this will silently change code behavior.

dwalton76: yeah I looked at the PIL docs...it is a code bug on our part.

WasabiFan: I was just noting that right now you can do `my_display.rectangle(True, x1, y1, x2, y2)` and it works. So we could say it's a bug with our documentation and variable naming rather than with the passed parameters.

dwalton76: True...it would probably be less confusing to stay consistent with PIL long term though. Our API matches there everywhere else (I think)

WasabiFan: Oh, really? Maybe I am looking in the wrong place. It seems to me that PIL's methods use point pairs: https://pillow.readthedocs.io/en/3.1.x/reference/Image.html

dwalton76: True, I see what you are saying now.  Given that our ``rectangle()`` accepts ``x, y, width, height`` it seems like a less drastic change to fix the behavior (ie the code change in the current pull request)  than it does to change our ``rectange()`` to accept ``x1, y1, x2, y2``.

WasabiFan: Agreed; I think the naming/docs change would make the most sense given that we've released it with that behavior already.

WasabiFan (approved):

Issue #527, opened 2018-09-21
Add port name constants for stacked BrickPi's


WasabiFan:
Mentioned here: https://github.com/ev3dev/ev3dev-lang-python/issues/525

It seems it would be easy enough to add those. I believe the limit is 4 stacked BrickPi's at once.

dwalton76:
Cool I did not know you could stack them :)

WasabiFan:
Fixed by #562 and #572

PR #562, opened 2018-11-13
Add port name constants for stacked brickpi3s

dwalton76:
For issue #527 

I only have one brickpi3 so I do not have a way to test this.

### DIFF
--- a/ev3dev2/_platform/brickpi3.py
+++ b/ev3dev2/_platform/brickpi3.py
@@ -1,9 +1,25 @@
 from collections import OrderedDict
 
+# Up to four brickpi3s can be stacked
 OUTPUT_A = 'spi0.1:MA'
 OUTPUT_B = 'spi0.1:MB'
 OUTPUT_C = 'spi0.1:MC'
 OUTPUT_D = 'spi0.1:MD'
+
+OUTPUT_E = 'spi0.1:ME'
+OUTPUT_F = 'spi0.1:MF'
+OUTPUT_G = 'spi0.1:MG'
+OUTPUT_H = 'spi0.1:MH'
+
+OUTPUT_I = 'spi0.1:MI'
+OUTPUT_J = 'spi0.1:MJ'
+OUTPUT_K = 'spi0.1:MK'
+OUTPUT_L = 'spi0.1:ML'
+
+OUTPUT_M = 'spi0.1:MM'
+OUTPUT_N = 'spi0.1:MN'
+OUTPUT_O = 'spi0.1:MO'
+OUTPUT_P = 'spi0.1:MP'
 
 INPUT_1 = 'spi0.1:S1' 
 INPUT_2 = 'spi0.1:S2' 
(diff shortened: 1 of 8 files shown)


ev3dev2/_platform/brickpi3.py
```diff
@@ -1,10 +1,26 @@
...
+OUTPUT_L = 'spi0.1:ML'
+
+OUTPUT_M = 'spi0.1:MM'
+OUTPUT_N = 'spi0.1:MN'
+OUTPUT_O = 'spi0.1:MO'
+OUTPUT_P = 'spi0.1:MP'
+
 INPUT_1 = 'spi0.1:S1' 
 INPUT_2 = 'spi0.1:S2' 
 INPUT_3 = 'spi0.1:S3'
```
dlech: What about constants for the inputs?

dwalton76: doh! will fix

tests/api_tests.py
```diff
@@ -151,6 +155,11 @@ def test_infrared_sensor(self):
         self.assertEqual(val,               [])
 
     def test_medium_motor_write(self):
+
+        # brickpi/brickpi3 only support LargeMotor
+        if ev3dev2.get_current_platform() in ("brickpi", "brickpi3"):
```
WasabiFan: Is the issue here that running the tests on-device doesn't use the `fake` platform? I'd prefer to fix that rather than working around the detected platform.

WasabiFan:
I can't remember how this appears in docs... I'm guessing we need to manually add a mention of these new port constants though.

tests/api_tests.py
dwalton76: It uses the fake platform, the problem is on a brickpi/brickpi3 we raise now raise an Exception if any Motor class other than LargeMotor, is used.

WasabiFan: But if it's using the fake platform how does it know it's on a BrickPi?

dwalton76: I see what you are saying now...let me think about this for a bit

WasabiFan: My recollection was that the solution you implemented for the test was not to use the `fake` platform, but instead to manually import some things from `fake` in the tests. I think we should override the detected platform in tests.

dwalton76: We could have ``get_current_platform()`` check for a ``USE_FAKE_SYS`` env variable that is set by api_tests.py.

WasabiFan: Sounds like that would work.

WasabiFan:
Does this show up in the docs?

dwalton76:
We appear to have just about zero documentation on the ``OUPUT_*`` variables...maybe I am missing something?
```
robot@ev3dev:~/ev3dev-lang-python/docs$ git grep OUTPUT
upgrading-to-stretch.md:from ev3dev2.motor import Motor, OUTPUT_A
robot@ev3dev:~/ev3dev-lang-python/docs$
```

PR #564, opened 2018-11-13
cache attributes that never change

dwalton76:
For issue #365

### DIFF
--- a/ev3dev2/__init__.py
+++ b/ev3dev2/__init__.py
@@ -57,7 +57,7 @@ def get_current_platform():
     """
     board_info_dir = '/sys/class/board-info/'
 
-    if not os.path.exists(board_info_dir):
+    if not os.path.exists(board_info_dir) or os.environ.get("FAKE_SYS"):
         return 'fake'
 
     for board in os.listdir(board_info_dir):
@@ -291,7 +291,7 @@ class Device(object):
         value = self._attr_cache.get(keyword)
 
         if value is None:
-            (filehandle, value) = self.get_attr_int(self._count_per_rot, keyword)
+            (filehandle, value) = self.get_attr_int(filehandle, keyword)
             self._attr_cache[keyword] = value
 
         return (filehandle, value)
@@ -309,7 +309,7 @@ class Device(object):
         value = self._attr_cache.get(keyword)
 
         if value is None:
-            (filehandle, value) = self.get_attr_string(self._count_per_rot, keyword)
+            (filehandle, value) = self.get_attr_string(filehandle, keyword)
             self._attr_cache[keyword] = value
 
         return (filehandle, value)
@@ -328,7 +328,7 @@ class Device(object):
         value = self._attr_cache.get(keyword)
 
         if value is None:
-            (filehandle, value) = self.get_attr_set(self._count_per_rot, keyword)
+            (filehandle, value) = self.get_attr_set(filehandle, keyword)
             self._attr_cache[keyword] = value
 
         return (filehandle, value)
--- a/ev3dev2/__init__.py
+++ b/ev3dev2/__init__.py
@@ -145,7 +145,12 @@ class DeviceNotFound(Exception):
 class Device(object):
     """The ev3dev device base class"""
 
-    __slots__ = ['_path', '_device_index', 'kwargs']
+    __slots__ = [
+        '_path',
+        '_device_index',
+        '_attr_cache',
+        'kwargs',
+    ]
 
     DEVICE_ROOT_PATH = '/sys/class'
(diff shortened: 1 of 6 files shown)


dwalton76:
Tested with

```
#!/usr/bin/env python3

from time import perf_counter
from ev3dev2.motor import OUTPUT_A, OUTPUT_B, LargeMotor

lMotor = LargeMotor(OUTPUT_A)
rMotor = LargeMotor(OUTPUT_B)

LOOPS = 10000
lMotor.run_forever()
rMotor.run_forever()

# without caching this takes 7290ms
# with caching it takes 1325ms
startTime = perf_counter()
for a in range(0,LOOPS):
    lMotor.commands
    rMotor.commands
    lMotor.driver_name
    rMotor.driver_name
    lMotor.count_per_rot
    rMotor.count_per_rot
    lMotor.count_per_m
    rMotor.count_per_m
    lMotor.max_speed
    rMotor.max_speed
    lMotor.full_travel_count
    rMotor.full_travel_count

endTime = perf_counter()
lMotor.stop()
rMotor.stop()

print("{}s".format(endTime - startTime))
```
- without caching this takes 7290ms
- with caching it takes 1325ms

ev3dev2/__init__.py
```diff
@@ -281,6 +287,15 @@ def get_attr_int(self, attribute, name):
         attribute, value = self._get_attribute(attribute, name)
         return attribute, int(value)
 
+    def get_cached_attr_int(self, filehandle, keyword):
+        value = self._attr_cache.get(keyword)
+
+        if value is None:
+            (filehandle, value) = self.get_attr_int(self._count_per_rot, keyword)
```
WasabiFan: Specifically `_count_per_rot`? 😕

ev3dev2/motor.py
```diff
@@ -1065,7 +1065,7 @@ def list_motors(name_pattern=Motor.SYSTEM_DEVICE_NAME_CONVENTION, **kwargs):
     """
     class_path = abspath(Device.DEVICE_ROOT_PATH + '/' + Motor.SYSTEM_CLASS_NAME)
 
-    return (Motor(name_pattern=name, name_exact=True)
```
WasabiFan: Why? I don't see any reason to believe these will always be `LargeMotor`s.

WasabiFan (dismissed):

ev3dev2/__init__.py
dwalton76: crap this is a huge cut-n-paste error...will fix all of these

ev3dev2/motor.py
dwalton76: This is so this can work on brickpi

WasabiFan: - It should be OK to instantiate a `Motor` on BrickPi -- it's only `MediumMotor` that doesn't work. A `Motor` can control a physical large motor. The only difference between the classes is the driver it filters for.
- The method here supports `MediumMotor`s too and has to work on other platforms. With this change `MediumMotor` is broken.

dwalton76: I think we need to revisit the exception we are raising on brickpi for non-LargeMotor scenarios....will tackle that via a different issue.

ev3dev2/port.py
```diff
@@ -98,15 +98,15 @@ def driver_name(self):
         Returns the name of the driver that loaded this device. You can find the
         complete list of drivers in the [list of port drivers].
         """
-        self._driver_name, value = self.get_attr_string(self._driver_name, 'driver_name')
+        (self._driver_name, value) = self.get_cached_attr_string(self._driver_name, 'driver_name')
```
WasabiFan, line 101: Are you sure this can be cached? Aren't you able to change the driver _using this class_?

ev3dev2/port.py
```diff
@@ -98,15 +98,15 @@ def driver_name(self):
...
         return value
 
     @property
     def modes(self):
         """
         Returns a list of the available modes of the port.
         """
-        self._modes, value = self.get_attr_set(self._modes, 'modes')
+        (self._modes, value) = self.get_cached_attr_set(self._modes, 'modes')
```
WasabiFan, line 109: Ditto here

ev3dev2/port.py
dlech: This is the driver name for the port driver. It does not change. You are probably thinking of `set_device` which sets the sensor that is connected to the port.

ev3dev2/port.py
dlech: this does not change either

WasabiFan (approved):

Issue #565, opened 2018-11-14
brickpi/brickpi3 should be allowed to use Motor class 


dwalton76:
We recently started raising an Exception if someone uses a motor class other than ``LargeMotor`` on brickpi or brickpi3.  This needs a tweak to allow the use of the parent ``Motor`` class on brickpi/brickpi3.

WasabiFan:
I think this was fixed in #566

PR #566, opened 2018-11-14
brickpi(3) support use of the Motor class

dwalton76:
For issue #565

### DIFF
--- a/ev3dev2/motor.py
+++ b/ev3dev2/motor.py
@@ -355,7 +355,7 @@ class Motor(Device):
 
     def __init__(self, address=None, name_pattern=SYSTEM_DEVICE_NAME_CONVENTION, name_exact=False, **kwargs):
 
-        if platform in ('brickpi', 'brickpi3') and not isinstance(self, LargeMotor):
+        if platform in ('brickpi', 'brickpi3') and type(self).__name__ != 'Motor' and not isinstance(self, LargeMotor):
             raise Exception("{} is unaware of different motor types, use LargeMotor instead".format(platform))
 
         if address is not None:
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,6 +1,7 @@
 python-ev3dev2 (2.0.0~beta3) stable; urgency=medium
 
   [Daniel Walton]
+  * brickpi(3) support use of the Motor class
   * Moved MoveTank.off() to MotorSet
   * cache attributes that never change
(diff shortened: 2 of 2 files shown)


dwalton76:
Before
```
robot@pi3[ev3dev-lang-python]# ./utils/stop_all_motors.py 
  File "./utils/stop_all_motors.py", line 8, in <module>
    for motor in list_motors():
  File "/home/robot/ev3dev-lang-python/ev3dev2/motor.py", line 1069, in <genexpr>
    for name in list_device_names(class_path, name_pattern, **kwargs))
  File "/home/robot/ev3dev-lang-python/ev3dev2/motor.py", line 359, in __init__
    raise Exception("{} is unaware of different motor types, use LargeMotor instead".format(platform))
Exception: brickpi3 is unaware of different motor types, use LargeMotor instead
robot@pi3[ev3dev-lang-python]#
```

After
```
robot@pi3[ev3dev-lang-python]# ./utils/stop_all_motors.py 
robot@pi3[ev3dev-lang-python]# 
```

ev3dev2/motor.py
```diff
@@ -355,7 +355,7 @@ class Motor(Device):
 
     def __init__(self, address=None, name_pattern=SYSTEM_DEVICE_NAME_CONVENTION, name_exact=False, **kwargs):
 
-        if platform in ('brickpi', 'brickpi3') and not isinstance(self, LargeMotor):
+        if platform in ('brickpi', 'brickpi3') and type(self).__name__ != 'Motor' and not isinstance(self, LargeMotor):
```
WasabiFan, line 358: Can you just move the check into the `MediumMotor` class?

dwalton76: We could but we would also need to put the check in ``ActuonixL12100Motor``, ``ActuonixL1250Motor`` and any other future ``Motor`` child classes...seems cleaner to have it in Motor

WasabiFan, line 358: Is there a single style of instance check that would work for both of the latter cases (just to make it easier to read)?

WasabiFan (approved):

ev3dev2/motor.py
dwalton76: This was the cleanest solution I could come up with :( You can't use ``isinstance(self, Motor)`` because that will return True for any child class of ``Motor``.

Will go ahead and merge this in...we can always tweak later if we come up with a cleaner solution.

PR #573, opened 2018-12-15
motors.py: added missing parameter 'motors' to MotorSet.is_stalled

v-gar:
I noticed that something like

```
m = MoveTank(...)
m.is_stalled()
```

fails while e.g. `m.is_overloaded()` works. I looked up the appropriate method and found a missing parameter (motors=None) which was the reason for the error.
This PR will fix this issue by adding the missing parameter.

### DIFF
--- a/ev3dev2/motor.py
+++ b/ev3dev2/motor.py
@@ -1737,7 +1737,7 @@ class MotorSet(object):
         return self._is_state(motors, LargeMotor.STATE_OVERLOADED)
 
     @property
-    def is_stalled(self):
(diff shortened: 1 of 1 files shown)


WasabiFan (approved):

WasabiFan:
Thanks!

WasabiFan:
Added to changelog in https://github.com/ev3dev/ev3dev-lang-python/commit/8f58cc97a4e5d49828d7365ab4a86e18f0032cac; let me know if you'd prefer different attribution.

PR #581, opened 2018-12-19
Expose ev3dev2.version.__version__ as ev3dev2.__version__

WasabiFan:
I don't think we advertise this anywhere, but nonetheless it's nice to have it available from the root module.


dwalton76:
There isn't an ``ev3dev2/version.py`` though...not sure I am following where the version is actually defined?

WasabiFan:
`git_version.py` generates it as part of the packaging process. So in builds from the development tree it doesn't exist, but in releases it does.

PR #590, opened 2019-01-20
Missing rename namespace: ev3dev -> ev3dev2 for gyro

remvee:
This file was missed in PR #412.

Without this change the [balanc3r robot demo drive code](https://github.com/ev3dev/ev3dev-lang-python-demo/blob/stretch/robots/BALANC3R/drive.py) will exit in error failing to load `ev3dev.power`.


WasabiFan (approved):

WasabiFan:
Thanks!

PR #601, opened 2019-03-20
Allow Display.text_pixels() to receive a loaded font or a font string.

### LINKED ISSUE #600: Display.text_pixel and Display.text_grid laggy with non-default font.
- **ev3dev version:** 4.14.80-ev3dev-2.3.1-ev3
- **ev3dev-lang-python version:** 1.2.0, 2.0.0~beta2

When drawing text to the display, each Display.text_*() call with a font string specified takes nearly 0.5 seconds to execute.  This is problematic for programs that want to use the screen for status updates, while time sensitive actions are happening, such as driving while watching for a color sensor value to change, or a button to be pressed, etc.

 I was able to trace the lag to the loading of the font file.  I hacked a local change that directly uses the ImageDraw object methods to load the font once, and use it directly when calling the ImageDraw.text() method.  This eliminated the large lag.

I don't have time right now to create a pull request with a proposed change, but wanted to document this in case anyone else sees the lag.  I would also like to know if the authors are willing to consider a change that would allow a "current" font to be loaded and cached as part of the Display() object so that the Display.text_*() calls can still be used?  Or maybe, there is a different mechanism that would fit better with your overall design?

Thanks for a great system.

fractal13:
This will allow the users to preload a font and eliminate the constant
reloading of fonts, which can produce lag when displaying in a loop.

Maintainer edit: Fixes #600

### DIFF
--- a/ev3dev2/display.py
+++ b/ev3dev2/display.py
@@ -365,6 +365,7 @@ class Display(FbMem):
         Display `text` starting at pixel (x, y).
 
         The EV3 display is 178x128 pixels
+
         - (0, 0) would be the top left corner of the display
         - (89, 64) would be right in the middle of the display
 
@@ -374,10 +375,12 @@ class Display(FbMem):
         https://www.w3schools.com/colors/colors_names.asp
 
         'font' : can be any font displayed here
-            http://ev3dev-lang.readthedocs.io/projects/python-ev3dev/en/ev3dev-stretch/other.html#bitmap-fonts
+            http://ev3dev-lang.readthedocs.io/projects/python-ev3dev/en/ev3dev-stretch/display.html#bitmap-fonts
+
         - If font is a string, it is the name of a font to be loaded.
-        - If font is a Font object, returned from fonts.load(), then it is
+        - If font is a Font object, returned from :meth:`ev3dev2.fonts.load`, then it is
           used directly.  This is desirable for faster display times.
+
         """
 
         if clear_screen:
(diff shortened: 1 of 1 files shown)


ev3dev2/display.py
```diff
@@ -375,14 +375,19 @@ def text_pixels(self, text, clear_screen=True, x=0, y=0, text_color='black', fon
 
         'font' : can be any font displayed here
             http://ev3dev-lang.readthedocs.io/projects/python-ev3dev/en/ev3dev-stretch/other.html#bitmap-fonts
+        - If font is a string, it is the name of a font to be loaded.
+        - If font is a Font object, returned from fonts.load(), then it is
```
WasabiFan: Can you make this a docs link? Something like this, I think:

    :meth:`ev3dev2.fonts.load`

WasabiFan (commented):
Thanks! This looks pretty good to me.

The documentation generator (Sphinx) is unhappy about list formatting:

> /home/travis/build/ev3dev/ev3dev-lang-python/ev3dev2/display.py:docstring of ev3dev2.display.Display.text_pixels:14:Definition list ends without a blank line; unexpected unindent.

ev3dev2/display.py
fractal13: I added this reference link, and fixed the links to fonts, since they are not longer in the other.html, but in display.html.

fractal13:
Now passes the sphinx-build.  The problem was with the list in the docstring not having a blank line after it.

WasabiFan (approved):
Thanks!

WasabiFan:
This will be available in the next release.

How would you like to appear in the [changelog](https://github.com/ev3dev/ev3dev-lang-python/blob/ev3dev-stretch/debian/changelog)? (name or username?)

fractal13:
Name: [Curtis Larsen]

Issue #605, opened 2019-03-24
Revert commit for #548


dwalton76:
Originally the brickpi3 would set the motor state to ``running holding`` when applying the brake but the motor was not moving.  Only the brickpi3 and pistorms behaved this way but we commited a fix to account for it here (it was for issue #548 )
https://github.com/ev3dev/ev3dev-lang-python/commit/95ccb36053db5099dab9a47f9b7cf3b5d7f4e3ea

David changed the brickpi3 behavior to be consistent with other platforms:
https://github.com/ev3dev/ev3dev/issues/1188

We should back out the fix for #548

PR #607, opened 2019-03-24
Revert "wait_until_not_moving should consider 'running holding' as 'n…

dwalton76:


dwalton76:
Resolves #605

debian/changelog
```diff
@@ -11,7 +11,6 @@ python-ev3dev2 (2.0.0~beta3) stable; urgency=medium
   * Sound: removed play() in favor of play_file()
   * suppress unittest ResourceWarning messages
   * Added DistanceValue classes
-  * wait_until_not_moving should consider "running holding" as "not moving"
```
WasabiFan: We shouldnt remove things from the released changelog...

WasabiFan (approved):
Should add a new changelog entry

dwalton76:
> Should add a new changelog entry

what should the new header be in the changelog?  Or do we keep the current one?

WasabiFan:
Create a new one and replace "stable" with "UNRELEASED" for now.

WasabiFan (approved):

PR #609, opened 2019-03-25
Fix GyroAngle.reset()

### LINKED ISSUE #510: gyro.reset() error - needs bytes-like object not int
- **ev3dev version: 4.14.61-ev3dev-2.2.2-ev3
- **ev3dev-lang-python version:
Desired=Unknown/Install/Remove/Purge/Hold
| Status=Not/Inst/Conf-files/Unpacked/halF-conf/Half-inst/trig-aWait/Trig-pend
|/ Err?=(none)/Reinst-required (Status,Err: uppercase=bad)
||/ Name                      Version           Architecture      Description
+++-=========================-=================-=================-=======================================================
ii  python3-ev3dev            1.2.0             all               Python language bindings for ev3dev
ii  python3-ev3dev2           2.0.0~beta1       all               Python language bindings for ev3dev

To reset a lego mindstorms gyro, one needs to write the byte 17 directly to the UART.
sensor/lego.py attempts to do this by doing:

self.set_attr_raw(self._direct, 'direct', 17)

This maps to a file IO write, which can only take a bytes-like object and not the int 17.

Changing the "17" to "bytes((17,))" seems to fix the problem. "bytes([17])" would also work, but is less efficient.

dlech:
`Sensor.set_attr_raw()` takes a bytes-like object for the `value` parameter.

Also add a comment about where the value comes from while we are at it.

Note: this requires a kernel driver fix before the reset will actually work See ev3dev/ev3dev#1236.

Suggested-by: @matskew 
Fixes: #510

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -4,6 +4,9 @@ python-ev3dev2 (2.0.0~beta4) UNRELEASED; urgency=medium
   * Sound: fallback to regular case of 'note' if uppercase is not found
   * wait_until_not_moving should consider "running holding" as "moving"
 
+  [David Lechner]
+  * Fix and document ev3dev2.sensors.lego.GyroSensor.reset()
+
  -- Kaelin Laundry <wasabifan@outlook.com>  Sun, 24 Mar 2019 00:25:00 -0800
 
 python-ev3dev2 (2.0.0~beta3) stable; urgency=medium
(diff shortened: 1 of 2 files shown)


WasabiFan (approved):
Thanks! Can you add a changelog entry?

dlech:
done

WasabiFan:
Thanks!

PR #610, opened 2019-03-26
HDMI display support for raspberry pi based platforms

dwalton76:


dwalton76:
Will also fix this while in this area
```
Exception ignored in: <bound method FbMem.__del__ of <ev3dev2.display.Display object at 0x76535480>>
  File "/home/robot/ev3dev-lang-python/ev3dev2/display.py", line 158, in __del__
AttributeError: 'NoneType' object has no attribute '_close_fbdev'
```

dlech:
BirckPi3 doesn't have a display.

dwalton76:
Poorly worded title on my part..."HDMI display support for raspberry pi based platforms" maybe?  Even then though the code itself is checking that ``platform`` is ``brickpi3`` so feels kinda clunky.

dlech:
Ah, so really the Raspberry Pi HDMI out.

dwalton76:
This ``__del__`` in FbMem is problematic.  Even after adding the ``if FbMem`` check that is sometimes true only for us to barf on line 186 because the os module has been purged

```
183     @staticmethod
184     def _close_fbdev(fbfid):
185         """Close the framebuffer file descriptor."""
186         os.close(fbfid)
```

traceback
```
Exception ignored in: <bound method FbMem.__del__ of <ev3dev2.display.Display object at 0x766d8c90>>
  File "/home/robot/ev3dev-lang-python/ev3dev2/display.py", line 169, in __del__
  File "/home/robot/ev3dev-lang-python/ev3dev2/display.py", line 186, in _close_fbdev
TypeError: 'NoneType' object is not callable
```

I am not familiar with the FbMem history, anyone know why we have this ``__del__``?

ev3dev2/display.py
```diff
@@ -211,10 +231,12 @@ def __init__(self, desc='Display'):
             im_type = "1"
         elif self.var_info.bits_per_pixel == 16:
             im_type = "RGB"
-        elif self.platform == "ev3" and self.var_info.bits_per_pixel == 32:
+        elif ((self.platform == "ev3" or self.platform in platforms_with_hdmi_display) and
```
dlech: I don't think we need to check for "platform" by name here at all. Just check for 32 bits per pixel and not EV3. `im_type = "L"` means grayscale, so I don't think we want this for HDMI otherwise you won't be able to do anything in color.

Maybe something like this?

```
        elif self.platform == "ev3" and self.var_info.bits_per_pixel == 32:
            im_type = "L"
        elif self.var_info.bits_per_pixel == 16 or self.var_info.bits_per_pixel == 32:
            im_type = "RGB"
```

ev3dev2/display.py
```diff
@@ -297,10 +319,12 @@ def update(self):
             self.mmap[:len(b)] = b
         elif self.var_info.bits_per_pixel == 16:
             self.mmap[:] = self._img_to_rgb565_bytes()
-        elif self.platform == "ev3" and self.var_info.bits_per_pixel == 32:
+        elif ((self.platform == "ev3" or self.platform in platforms_with_hdmi_display) and
```
dlech: we can probably just drop the `self.platform == "ev3"` part here.

```
        elif self.var_info.bits_per_pixel == 32:
            self.mmap[:] = self._img.convert("RGB").tobytes("raw", "XRGB")
```

WasabiFan:
That `__del__` implementation has been there since initial display support: https://github.com/ev3dev/ev3dev-lang-python/commit/ab003c65b5ff137bd42754a0a4d9d1e2ae9af761

Not sure how it is supposed to work exactly.

This PR looks good after that's sorted out though.

dwalton76:
> That `__del__` implementation has been there since initial display support: [ab003c6](https://github.com/ev3dev/ev3dev-lang-python/commit/ab003c65b5ff137bd42754a0a4d9d1e2ae9af761)
> 
> Not sure how it is supposed to work exactly.
> 
> This PR looks good after that's sorted out though.

From reading this:
https://stackoverflow.com/questions/1481488/what-is-the-del-method-how-to-call-it

It sounds like we should not be using ``__del__``.  I am going to remove it and do some testing (on brickpi3 and EV3) and see if anything crops up.

ev3dev2/display.py
dwalton76: Whether ``L`` or ``RGB`` is used for brickpi3 the screen has this yellow/green color
IMG_7464

via
```
display = Display()

display_font = "luBS24"
x_grid = 4
y_grid = 4
display.text_grid("BAR", clear_screen=True, x=x_grid, y=y_grid, font=display_font)
display.update()
sys.exit(0)
```

Not a huge deal at this point as any display at all is a step in the right direction but I haven't figured out yet where this color is coming from.

ev3dev2/display.py
```diff
@@ -143,11 +153,6 @@ def __init__(self, fbdev=None):
         self.var_info = FbMem._get_var_info(fid)
         self.mmap = fbmmap
 
-    def __del__(self):
-        """Close the FbMem framebuffer memory object."""
-        self.mmap.close()
-        FbMem._close_fbdev(self.fid)
-
```
dwalton76, line 150: Have been testing on an EV3 and all seems fine without this ``__del__``

ev3dev2/display.py
WasabiFan: I recall there's a code path which uses something like a 5-6-5 bit width pattern... is it possible that that's somehow involved?

You could probably get a good clue about what's going on if you looked at a dump of the raw data it's writing to the framebuffer.

dwalton76:
No luck figuring out where that green/yellow color is coming from.  It isn't that important to me right now and this is a step in the right direction so am going to go ahead and merge it.  We can dig into the green/yellow thing later if it becomes a more significant issues for someone.

PR #611, opened 2019-03-30
Initial draft of MicroPython docs

WasabiFan:
This is a work in progress... I still need to validate that the table isn't lying.


dwalton76 (approved):
LGTM

staadecker:
I'm looking forward to this! How come Button is not supported? Also is Sound supported? Any workarounds for that? The reason I ask is this would speed up development greatly. However, for the WRO contest, I use the button to start the robot after the program has loaded so I need to be able to use buttons. Also having audio cues helps a lot while testing.

WasabiFan:
For buttons, see https://github.com/ev3dev/ev3dev-lang-python/issues/571.

I'm not sure about sound; my guess is that some works and some doesn't.

docs/micropython.md
```diff
@@ -0,0 +1,45 @@
...
+MicroPython, we recommend you run your code with it for improved performance.
+
+## Module support
+
+```eval_rst
+==============================  =================
+Module                          Support status
+==============================  =================
+`ev3dev2.button`                ❌
+`ev3dev2.control` [1]_          ⚠️
```
dwalton76, line 14: For the control part I created issues for the areas that need work:
- #636 
- #637 
- #638

staadecker:
Sound doesn't work because micropython can't run the imports at the start of the file. See #628 . So maybe mark as X. If #632 is merged then Sound will be supported.

WasabiFan:
I'll merge this once #639 is merged.

dwalton76:
@WasabiFan #639 is in now.  I had not noticed it before but this PR had the same test failures that I was seeing due to shutil.rmtree not working.

dwalton76:
Pulled in latest changes from ev3dev-stretch to fix shutil.rmtree and tests are passing now

PR #615, opened 2019-04-09
MoveDifferential odometry support

dwalton76:
For issue #235 

This is not ready to be merged yet...needs some discussion first.


dwalton76:
@WasabiFan say I have a robot that starts at (0,0) and I tell it to drive forward in a straight line...so we will see x stay at 0 and y increase.  Should the robot's angle be considered 0 or 90?  The current code (I ported a C implementation I found) considers this to be an angle of 0 but in my head it seems like it should be 90.

dwalton76:
PS `on_to_coordinates` and and `turn_to_angle` have not been tested yet

WasabiFan:
For the angle, I'd suggest we maintain 90 being to the right (positive X) so that triggered functions work as they would on a typical unit circle layout.

ev3dev2/motor.py
```diff
@@ -2182,6 +2189,138 @@ def turn_left(self, speed, degrees, brake=True, block=True):
...
+                right_mm = float(right_rotations * self.wheel.circumference_mm)
+
+                # calculate distance we have traveled since last sampling
+                mm = (left_mm + right_mm) / 2.0
+
+                # accumulate total rotation around our center
+                self.theta += (left_mm - right_mm) / self.wheel_distance_mm
+
+                # and clip the rotation to plus or minus 360 degrees
+                self.theta -= float(int(self.theta/TWO_PI) * TWO_PI)
```
WasabiFan, line 2276: This seems strange... If we to above 2pi it's shifted down to be greater than 0, but we can go below zero until we get to -2pi... In other words, we allow a range of `[-2pi,2pi]` which is 2 rotations. Shouldn't this me a mod here to give us `(0,2pi]`?

ev3dev2/motor.py
```diff
@@ -2182,6 +2189,138 @@ def turn_left(self, speed, degrees, brake=True, block=True):
         """
         self._turn(speed, abs(degrees) * -1, brake, block)
 
+    def odometry_coordinates_log(self):
+        log.info("%s: odometry angle %s at (%d, %d)" %
+            (self, math.degrees(self.theta), self.x_pos_mm, self.y_pos_mm))
+
+    def odometry_start(self, theta_degrees_start=0.0, x_pos_start=0.0, y_pos_start=0.0):
```
WasabiFan: What do you think about breaking this out into a separate `WheelOdometry` class which can also be used independently? As in, you could instantiate a class instance and call `.start()` and it would track position for you? This class would then store an instance of that odom class when needed.

ev3dev2/motor.py
```diff
@@ -2182,6 +2189,138 @@ def turn_left(self, speed, degrees, brake=True, block=True):
         """
         self._turn(speed, abs(degrees) * -1, brake, block)
 
+    def odometry_coordinates_log(self):
+        log.info("%s: odometry angle %s at (%d, %d)" %
```
WasabiFan: Do you know what loop times we get on the EV3?

ev3dev2/motor.py
```diff
@@ -2182,6 +2189,138 @@ def turn_left(self, speed, degrees, brake=True, block=True):
...
+        which will set odometry_thread_run to False
+        """
+
+        def _odometry_monitor():
+            left_previous = 0
+            right_previous = 0
+            self.theta = math.radians(theta_degrees_start)  # robot heading
+            self.x_pos_mm = x_pos_start  # robot X position in mm
+            self.y_pos_mm = y_pos_start  # robot Y position in mm
+            SLEEP_TIME = 0.05  # 50ms
```
WasabiFan: This seems long... Is this just as fast as we can go? Or was that what the source you referenced uses?

ev3dev2/motor.py
```diff
@@ -2182,6 +2189,138 @@ def turn_left(self, speed, degrees, brake=True, block=True):
...
+        Signal the odometry thread to exit and wait for it to exit
+        """
+
+        if self.odometry_thread_id:
+            self.odometry_thread_run = False
+
+            while self.odometry_thread_id:
+                pass
+
+    def turn_to_angle(self, speed, angle_target_degrees, brake=True, block=True):
```
WasabiFan, line 2301: We should make it clear with docs and maybe in the name that this is for odometry and uses absolute angles.

ev3dev2/motor.py
```diff
@@ -2182,6 +2189,138 @@ def turn_left(self, speed, degrees, brake=True, block=True):
...
+
+        if angle_target_degrees < 0:
+            angle_target_degrees += 360
+
+        angle_current_degrees = math.degrees(self.theta)
```
WasabiFan, line 2311: What if I call this without odometry running? Should probably error if that's the case.

ev3dev2/motor.py
```diff
@@ -2182,6 +2189,138 @@ def turn_left(self, speed, degrees, brake=True, block=True):
...
+            angle_current_degrees += 360
+
+        angle_delta = angle_current_degrees - angle_target_degrees
+
+        log.info("%s: turn_to_angle %s, current angle %s, delta %s" %
+            (self, angle_target_degrees, angle_current_degrees, angle_delta))
+        self.odometry_coordinates_log()
+
+        # Do we rotate left or right and by how many degrees?
+        if angle_current_degrees > angle_target_degrees:
```
WasabiFan: I think this needs to do a more intelligent conversion by finding the shortest difference around the circle... What happens if I'm facing 5 degrees and request 355? Right now it looks like it will turn 350 degrees clockwise rather than 10 counterclockwise.

ev3dev2/motor.py
```diff
@@ -2182,6 +2189,138 @@ def turn_left(self, speed, degrees, brake=True, block=True):
...
+            self.turn_right(speed, angle_delta, brake, block)
+        else:
+            log.info("%s: turn left %s degrees" % (self, abs(angle_delta)))
+            self.turn_left(speed, angle_delta, brake, block)
+
+        self.odometry_coordinates_log()
+        log.info("\n\n\n")
+        # log.info("%s: angle is %.5f radians (%.5f degrees)." % (self, angle_rad, angle_target_degrees))
+
+    def on_to_coordinates(self, speed, x_target_mm, y_target_mm, brake=True, block=True):
```
WasabiFan, line 2340: It would be cool if we had one which curved to put you at a point facing a direction... I have some algorithm ideas. That can wait though.

ev3dev2/motor.py
dwalton76: I'm not sure...this is a straight port of the C implementation I found

dwalton76: Once I get further along with this I can test both methods and see which one looks correct

ev3dev2/motor.py
dwalton76: I kinda like it as part of `MoveDifferential` myself

ev3dev2/motor.py
dwalton76: This was just from the C port...will do some testing later to figure out how short we can realistically make this sleep on an ev3.

ev3dev2/motor.py
dwalton76: sorry for the confusion, this wasn't ready for review. I was posting the patch to discuss the angle thing to try to get my head wrapped around that part.

ev3dev2/motor.py
dwalton76: yep, can add that

ev3dev2/motor.py
dwalton76: yeah telling the robot to drive in a arc to reach the destination is very doable (we have arc_left and arc_right already).

ev3dev2/motor.py
dwalton76: ack

ev3dev2/motor.py
dwalton76: The reason being I don't think we have any other classes where we could do odometry.

ev3dev2/motor.py
dwalton76: > For the angle, I'd suggest we maintain 90 being to the right (positive X) so that triggered functions work as they would on a typical unit circle layout

To clarify here, if the robot drives forward in a straight line `self.theta` will be 0 because that is the initial value of `self.theta`. Are you saying it is correct at 0 or it should be 90?  If we change the intiial value to 90 we will need to tweak some of the other match in this loop (I had tried this and using 90 as the initial value causes x to increase while y remains 0 when driving in a straight line).

One option here would be to store two values:
- `self.theta` can remain as is, this will be a measure of how many degrees the robot has turned
- A new variable `self.angle_direction` (something like that) that would start at 90 and would tell you which way the robot is facing

WasabiFan: If we want to make that change, it should just be a matter of changing the initial angle to 90 degrees and then swapping sin with cos and vice versa.

WasabiFan: If we want to make that change, it should just be a matter of changing the initial angle to 90 degrees and then swapping sin with cos and vice versa.

ev3dev2/motor.py
dwalton76: Testing now on a pi3 and with no sleep at all there are normally 2ms to 5ms between lines of log output.  Will test later on EV3.

dwalton76: On an EV3 it is hard to tell...just the log.info() call adds 30ms :(

ev3dev2/motor.py
dwalton76: I fixed all of this...we start at 90 degrees now.

ev3dev2/motor.py
dwalton76: I still needs to add docstrings

ev3dev2/motor.py
WasabiFan: If you'd prefer it within this class, that's fine with me :+1:

ev3dev2/motor.py
WasabiFan: 2-5ms seems reasonable given the constraints... RE: the thread about sleep time, I do feel like the time should be shorter, but if it works as-is that isn't really an issue.

dwalton76:
@WasabiFan I think this one is ready now

WasabiFan (approved):
If this is tested and works, LGTM

Issue #583, opened 2018-12-20
Race condition in "wait()" causes call to block infinitely


WasabiFan:
- **ev3dev version:**

**System info (from `ev3dev-sysinfo`)**
```
Image file:         ev3dev-stretch-ev3-generic-2018-07-15
Kernel version:     4.14.71-ev3dev-2.3.0-ev3
Brickman:           0.10.0
BogoMIPS:           148.88
Board:              board0
BOARD_INFO_HW_REV=7
BOARD_INFO_MODEL=LEGO MINDSTORMS EV3
BOARD_INFO_ROM_REV=6
BOARD_INFO_SERIAL_NUM=00165340720B
BOARD_INFO_TYPE=main
```
- **ev3dev-lang-python version:** current `ev3dev-stretch` branch

From https://github.com/ev3dev/ev3dev/issues/1200.

When running a series of small movements where the motor will almost immediately reach the target, we sometimes hit a case where our wait logic gets stuck. This is improved by the re-ordering we did recently of the relevant wait code, but nonetheless still happens (in a slightly different way).

https://github.com/ev3dev/ev3dev-lang-python/blob/056f721c383a8a4331f7cc7f9958ffdcf794e573/ev3dev2/motor.py#L849-L874

From what I can tell, we get through the check on line 868 and determine we can't immediately return; so, we then call `poll` to block until something changes. In the time between when that happens and when the `poll` internals actually get around to checking, the `state` has been updated from `running` to `holding` and we've missed it. This causes the method to block indefinitely.

I don't see an obvious solution to this. We could add a timeout when waiting for the motor to stop (so it would at least not freeze). Alternately, perhaps we could modify the code where we call `poll` so that it always uses a small timeout so we manually check occasionally even if `poll` missed it. This latter approach would degrade responsiveness in the general case but improve the edge case.

dlech:
Off topic: `None if timeout is None else timeout` is no different than just `timeout`.

On topic: a timeout could always be passed to `poll()` even when the `timeout` argument is `None` to force checking the state again.

strichter:
This issue greatly limits the utility of the new library at this point, since any quick action will cause a deadlock.

I think the solution will be to pass the state upon original checking to the poll itself and instead of relying on the attribute access later on. I would add a line before 868 saying `state = self.state` and then pass `state` into the polling somehow. (I have not worked with this code at all yet.) That would completely eliminate the race condition without degrading performance.

[...]

PR #622, opened 2019-04-21
Avoid poll(None) if no timeout specified

dwalton76:
Potential fix for #583

### DIFF
--- a/ev3dev2/motor.py
+++ b/ev3dev2/motor.py
@@ -865,14 +865,22 @@ class Motor(Device):
             self._poll = select.poll()
             self._poll.register(self._state, select.POLLPRI)
 
+        # Set poll timeout to something small
+        if timeout:
+            poll_tm = min(timeout, 100)
+        else:
+            poll_tm = 100
+
         while True:
+            # This check is now done every poll_tm even if poll has nothing to report:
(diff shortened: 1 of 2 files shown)


ev3dev2/motor.py
```diff
@@ -865,14 +865,22 @@ def wait(self, cond, timeout=None):
             self._poll = select.poll()
             self._poll.register(self._state, select.POLLPRI)
 
+        # Set poll timeout to something small
```
WasabiFan: Can you add a comment with a link to the original issue for future reference?

WasabiFan (approved):
I'd like to see verification that it solves the problem, but this looks good to me in theory.

dwalton76:
I think I had worked around this problem in my cubesolver by putting multi second limits just about everywhere I use wait...so I don't have a good way to test/verify that this fixes the issue. I would be fine if we go ahead and commit it though, the fix looks pretty straightforward and should fix the issue.

WasabiFan:
As long as it doesn't make behavior worse, I'm OK with merging.

dwalton76:
I did some basic testing with python3 and micropython and it seems fine...will merge.

Issue #625, opened 2019-04-26
PID line follower


dwalton76:
It would be cool to support a PID line follower such as:
- https://www.youtube.com/watch?v=AMBWV_HGYj4
- https://github.com/CS2098/Line-Follower/blob/master/line_follower.py

PR #630, opened 2019-05-11
PID Line follower

dwalton76:
For #625

### DIFF
--- a/ev3dev2/motor.py
+++ b/ev3dev2/motor.py
@@ -2010,7 +2010,7 @@ class MoveTank(MotorSet):
         ``white`` is the reflected_light_intensity that is used to determine if we have
             lost the line
 
-        ``OFF_LINE_COUNT`` is how many consecutive times through the loop the
+        ``off_line_count_max`` is how many consecutive times through the loop the
             reflected_light_intensity must be greater than ``white`` before we
             declare the line lost and raise an exception
 
--- a/ev3dev2/motor.py
+++ b/ev3dev2/motor.py
@@ -48,7 +48,6 @@ log = getLogger(__name__)
 # update to 'running' in the "on_for_XYZ" methods of the Motor class
 WAIT_RUNNING_TIMEOUT = 100
 
-stopwatch = None
 
 # OUTPUT ports have platform specific values that we must import
 platform = get_current_platform()
@@ -1792,7 +1791,7 @@ class MotorSet(object):
 
 
 # line follower classes
-class LineFollowError(Exception):
+class LineFollowErrorLostLine(Exception):
     """
     Raised when a line following robot has lost the line
     """
@@ -1808,26 +1807,24 @@ class LineFollowErrorTooFast(Exception):
 
 
 # line follower functions
-def follow_for_forever(_tank):
+def follow_for_forever(tank):
     """
-    ``_tank``: the MoveTank object that is following a line
+    ``tank``: the MoveTank object that is following a line
     """
     return True
 
 
-def follow_for_ms(_tank, ms):
+def follow_for_ms(tank, ms):
     """
-    ``_tank``: the MoveTank object that is following a line
+    ``tank``: the MoveTank object that is following a line
     ``ms`` : the number of milliseconds to follow the line
     """
-    global stopwatch
+    if not hasattr(tank, 'stopwatch') or tank.stopwatch is None:
+        tank.stopwatch = StopWatch()
+        tank.stopwatch.start()
 
-    if stopwatch is None:
-        stopwatch = StopWatch()
-        stopwatch.start()
-
-    if stopwatch.value_ms >= ms:
-        stopwatch = None
+    if tank.stopwatch.value_ms >= ms:
+        tank.stopwatch = None
         return False
     else:
         return True
@@ -2091,7 +2088,7 @@ class MoveTank(MotorSet):
 
                 if off_line_count >= off_line_count_max:
                     self.stop()
-                    raise LineFollowError("we lost the line")
+                    raise LineFollowErrorLostLine("we lost the line")
             else:
                 off_line_count = 0
 
--- a/ev3dev2/motor.py
+++ b/ev3dev2/motor.py
@@ -76,14 +76,14 @@ else:
     raise Exception("Unsupported platform '%s'" % platform)
 
 
-class LineFollowFailed(Exception):
+class LineFollowError(Exception):
     """
     Raised when a line following robot has lost the line
     """
     pass
 
 
-class LineFollowTooFast(Exception):
+class LineFollowErrorTooFast(Exception):
     """
     Raised when a line following robot has been asked to follow
     a line an an unrealistic speed
@@ -102,7 +102,7 @@ class SpeedValue(object):
         return self.to_native_units() == other.to_native_units()
 
     def __ne__(self, other):
-        return self.to_native_units() != other.to_native_units()
+        return not self.__eq__(other)
 
     def __lt__(self, other):
         return self.to_native_units() < other.to_native_units()
@@ -1961,7 +1961,7 @@ class MoveTank(MotorSet):
             follow_left_edge=True,
             WHITE=60,
             OFF_LINE_COUNT_MAX=20,
-            SLEEP_SEC=None
+            SLEEP_TIME=0.01
         ):
         """
         PID line follower
@@ -1986,7 +1986,7 @@ class MoveTank(MotorSet):
             reflected_light_intensity must be greater than ``WHITE`` before we
             declare the line lost and raise an exception
 
-        ``SLEEP_SEC`` is how many seconds we sleep on each pass through
+        ``SLEEP_TIME`` is how many seconds we sleep on each pass through
             the loop.  This is to give the robot a chance to react
             to the new motor settings. This should be something small such
             as 0.01 (10ms).
@@ -2011,23 +2011,21 @@ class MoveTank(MotorSet):
             last_error = error
             turn_native_units = (kp * error) + (ki * integral) + (kd * derivative)
 
-            if follow_left_edge:
-                left_speed = SpeedNativeUnits(speed_native_units - turn_native_units)
-                right_speed = SpeedNativeUnits(speed_native_units + turn_native_units)
+            if not follow_left_edge:
+                turn_native_units *= -1
 
-            else:
-                left_speed = SpeedNativeUnits(speed_native_units + turn_native_units)
-                right_speed = SpeedNativeUnits(speed_native_units - turn_native_units)
+            left_speed = SpeedNativeUnits(speed_native_units - turn_native_units)
+            right_speed = SpeedNativeUnits(speed_native_units + turn_native_units)
 
             if left_speed > MAX_SPEED:
                 log.info("%s: left_speed %s is greater than MAX_SPEED %s"  % (self, left_speed, MAX_SPEED))
                 self.stop()
-                raise LineFollowTooFast("The robot is moving too fast to follow the line")
+                raise LineFollowErrorTooFast("The robot is moving too fast to follow the line")
 
             if right_speed > MAX_SPEED:
                 log.info("%s: right_speed %s is greater than MAX_SPEED %s"  % (self, right_speed, MAX_SPEED))
                 self.stop()
-                raise LineFollowTooFast("The robot is moving too fast to follow the line")
+                raise LineFollowErrorTooFast("The robot is moving too fast to follow the line")
 
             # Have we lost the line?
             if reflected_light_intensity >= WHITE:
@@ -2035,12 +2033,12 @@ class MoveTank(MotorSet):
 
                 if off_line_count >= OFF_LINE_COUNT_MAX:
                     self.stop()
-                    raise LineFollowFailed("we lost the line")
+                    raise LineFollowError("we lost the line")
             else:
                 off_line_count = 0
 
-            if SLEEP_SEC:
-                sleep(SLEEP_SEC)
+            if SLEEP_TIME:
+                time.sleep(SLEEP_TIME)
 
             self.on(left_speed, right_speed)
 
--- a/ev3dev2/motor.py
+++ b/ev3dev2/motor.py
@@ -1827,9 +1827,9 @@ def follow_for_ms(_tank, ms):
         stopwatch.start()
 
     if stopwatch.value_ms >= ms:
+        stopwatch = None
         return False
     else:
-        stopwatch = None
         return True
 
 
@@ -1992,7 +1992,7 @@ class MoveTank(MotorSet):
             off_line_count_max=20,
             sleep_time=0.01,
             follow_for=follow_for_forever,
-            **kwargs,
+            **kwargs
         ):
         """
         PID line follower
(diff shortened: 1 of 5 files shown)


ev3dev2/motor.py
```diff
@@ -76,16 +76,46 @@
     raise Exception("Unsupported platform '%s'" % platform)
 
 
+class LineFollowFailed(Exception):
```
WasabiFan: I think we should suffix "Error" on these: https://www.python.org/dev/peps/pep-0008/#exception-names

ev3dev2/motor.py
```diff
@@ -76,16 +76,46 @@
...
     """
     A base class for other unit types. Don't use this directly; instead, see
     :class:`SpeedPercent`, :class:`SpeedRPS`, :class:`SpeedRPM`,
     :class:`SpeedDPS`, and :class:`SpeedDPM`.
     """
 
+    def __eq__(self, other):
+        return self.to_native_units() == other.to_native_units()
+
+    def __ne__(self, other):
```
WasabiFan, line 90: This should probably invoke `__eq__` to ensure it's always the inverse of `==`

ev3dev2/motor.py
```diff
@@ -1798,6 +1828,9 @@ def __init__(self, left_motor_port, right_motor_port, desc=None, motor_class=Lar
         self.right_motor = self.motors[right_motor_port]
         self.max_speed = self.left_motor.max_speed
 
+        # color sensor used by follow_line()
+        self.cs = None
```
WasabiFan, line 1858: What's the thought process behind building this into `MoveTank` rather than having a separate class? I feel like having a `LineFollower` class which can be constructed with a custom `MoveTank` instance if needed would be better, even if just for code separation. I'm a bit scared by classes like this gaining a many "optional" features which have to coexist.

ev3dev2/motor.py
```diff
@@ -1921,6 +1954,98 @@ def on(self, left_speed, right_speed):
         self.left_motor.run_forever()
         self.right_motor.run_forever()
 
+    def follow_line(self,
+            kp, ki, kd,
+            speed, keep_following_function,
```
WasabiFan: `predicate`?

ev3dev2/motor.py
```diff
@@ -1921,6 +1954,98 @@ def on(self, left_speed, right_speed):
         self.left_motor.run_forever()
         self.right_motor.run_forever()
 
+    def follow_line(self,
+            kp, ki, kd,
```
WasabiFan, line 1984: Do we have defaults for these?

ev3dev2/motor.py
```diff
@@ -1921,6 +1954,98 @@ def on(self, left_speed, right_speed):
...
+        while keep_following_function(self):
+            reflected_light_intensity = self.cs.reflected_light_intensity
+            error = target_light_intensity - reflected_light_intensity
+            integral = integral + error
+            derivative = error - last_error
+            last_error = error
+            turn_native_units = (kp * error) + (ki * integral) + (kd * derivative)
+
+            if follow_left_edge:
+                left_speed = SpeedNativeUnits(speed_native_units - turn_native_units)
```
WasabiFan: I think you can just do

```python
if not follow_right_edge:
    turn_native_units *= -1
```

before and have one version of this.

ev3dev2/motor.py
```diff
@@ -1921,6 +1954,98 @@ def on(self, left_speed, right_speed):
...
+            if reflected_light_intensity >= WHITE:
+                off_line_count += 1
+
+                if off_line_count >= OFF_LINE_COUNT_MAX:
+                    self.stop()
+                    raise LineFollowFailed("we lost the line")
+            else:
+                off_line_count = 0
+
+            if SLEEP_SEC:
```
WasabiFan: No sleep at all by default? I would suggest making sure that there's always a consistent delay otherwise the derivative and integral behavior will by inconsistent and unpredictable. Alternately, we could use real units for derivative and integral (i.e., divide by time delta).

utils/stop_all_motors.py
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env python3
+#!/usr/bin/env micropython
```
WasabiFan, line 1: I don't think we want to change this globally...

utils/line-follower-find-kp-ki-kd.py
```diff
@@ -0,0 +1,152 @@
...
+The default speed for this program is SpeedPercent(30).  You can use whatever
+speed you want, just search for "speed = SpeedPercent(30)" in this file and
+modigy that line.
+
+You can use the pdf from this site to print a line that makes an oval, just
+have your robot follow that oval when running this program.
+http://robotsquare.com/2012/11/28/line-following/
+"""
+
+from ev3dev2.motor import OUTPUT_A, OUTPUT_B, MoveTank, SpeedPercent, LineFollowFailed
```
WasabiFan: Cool script!

ev3dev2/motor.py
dwalton76: yeah something I have been contemplating is do we just make a `FirstLegoLeagueRobot` class/file (open for suggestions on the name) where we have a child class of `MoveTank` that holds these sorts of things.  All of the odometry functionality in `MoveTank` could go here as well.

ev3dev2/motor.py
dwalton76: ??

ev3dev2/motor.py
dwalton76: They are so dependent on the individual robot I don't think we could pick defaults that would work well.

ev3dev2/motor.py
dwalton76: I'll do some testing with 10ms and see how that behaves.  This is another one of those things that might be highly dependent on the type of robot.

utils/stop_all_motors.py
dwalton76: It stops the motors so much faster though...what's the downside?

ev3dev2/motor.py
dwalton76: will test

ev3dev2/motor.py
WasabiFan: Why not just make individual classes for each behavior, so I can do this?

```python
drive = MoveTank(OUTPUT_A, OUTPUT_B)

# get other utilities if I want them
line_follower = PidLineFollower(drive)
odometry = Odometry(drive) # or just take in motors directly

line_follower.follow_line()
# use odometry...

```

ev3dev2/motor.py
WasabiFan: It's just not clear from the name what `keep_following_function` is supposed to do... it's really supposed to be a "predicate" with no side-effects. As it is I might interpret the name to mean I have to give a function that keeps following the line. The naming issue is pretty minor though.

utils/stop_all_motors.py
WasabiFan: Mhmmm... I guess that makes sense. The code's all the same so people can still refer to it, so I don't see any real harm.

ev3dev2/motor.py
dwalton76: I just tend to think in terms of adding functionality via methods in child classes.  To me that is cleaner and easier to follow than having a standalone function that can only accept certain types of objects to implement the desired functionality.

dwalton76:
@WasabiFan I should be able to get back to this PR this weekend

ev3dev2/motor.py
dwalton76: I am struggling to come up with a better name here :(  The docstring explains the behavior
```
        ``keep_following_function`` is called to determine if we should keep
            following the line or stop
```

I'm open to renaming it if you have suggestions.

ev3dev2/motor.py
dwalton76: yep that worked

ev3dev2/motor.py
dwalton76: 10ms seemed fine

dwalton76:
FYI tested with this
```
from ev3dev2.motor import OUTPUT_A, OUTPUT_B, MoveTank, SpeedRPM, SpeedPercent, follow_for_ms, follow_for_forever
from ev3dev2.sensor.lego import ColorSensor
from time import sleep
import logging
import sys


# logging


tank = MoveTank(OUTPUT_A, OUTPUT_B)
tank.cs = ColorSensor()

try:
    log.info("start")
    tank.follow_line(kp=11.3, ki=0.05, kd=3.2,
        speed=SpeedRPM(60),
        # follow_for=follow_for_forever,
        follow_for=follow_for_ms,
        ms=4500,
    )
    log.info("end")
except Exception:
    tank.stop()
    raise

tank.stop()

```

ev3dev2/motor.py
```diff
@@ -76,16 +76,46 @@
...
 
 
+class LineFollowError(Exception):
+    """
+    Raised when a line following robot has lost the line
+    """
+    pass
+
+
+class LineFollowErrorTooFast(Exception):
```
WasabiFan: We probably need an example in the method's docstring which uses it and shows how to handle these.

ev3dev2/motor.py
```diff
@@ -1921,6 +1954,96 @@ def on(self, left_speed, right_speed):
...
+            SLEEP_TIME=0.01
+        ):
+        """
+        PID line follower
+
+        ``kp``, ``ki``, and ``kd`` are the PID constants.
+
+        ``speed`` is the desired speed of the midpoint of the robot
+
+        ``keep_following_function`` is called to determine if we should keep
```
WasabiFan: What are the expected scenarios here? I'd imagine the big ones are "follow for a distance" and "follow for a time"; could we provide easy ways to do that? Either provide functions which return the appropriate predicate, or (my preferred solution) also have keyword parameters for `follow_time` and `follow_distance`. And perhaps specifying none of them defaults to going forever.

We should also have an example of how to use this to specify common conditions.

ev3dev2/motor.py
```diff
@@ -1921,6 +1954,96 @@ def on(self, left_speed, right_speed):
...
         self.right_motor.run_forever()
 
+    def follow_line(self,
+            kp, ki, kd,
+            speed, keep_following_function,
+            target_light_intensity=None,
+            follow_left_edge=True,
+            WHITE=60,
+            OFF_LINE_COUNT_MAX=20,
+            SLEEP_TIME=0.01
```
WasabiFan: I think these should probably be lower_camel_case.

ev3dev2/motor.py
WasabiFan: Should probably document the argument it's passed.

ev3dev2/motor.py
```diff
@@ -1921,6 +1954,96 @@ def on(self, left_speed, right_speed):
...
+        last_error = 0.0
+        derivative = 0.0
+        off_line_count = 0
+        speed_native_units = speed.to_native_units(self.left_motor)
+        MAX_SPEED = SpeedNativeUnits(self.max_speed)
+
+        while keep_following_function(self):
+            reflected_light_intensity = self.cs.reflected_light_intensity
+            error = target_light_intensity - reflected_light_intensity
+            integral = integral + error
```
WasabiFan, line 2064: Does the implementation you were referencing not have an accumulator cap? I'm a bit surprised but I guess if it works it's good. Probably if it accumulates too much it's lost the line anyway so it doesn't much matter.

ev3dev2/motor.py
```diff
@@ -1921,6 +1954,96 @@ def on(self, left_speed, right_speed):
...
+                log.info("%s: right_speed %s is greater than MAX_SPEED %s"  % (self, right_speed, MAX_SPEED))
+                self.stop()
+                raise LineFollowErrorTooFast("The robot is moving too fast to follow the line")
+
+            # Have we lost the line?
+            if reflected_light_intensity >= WHITE:
+                off_line_count += 1
+
+                if off_line_count >= OFF_LINE_COUNT_MAX:
+                    self.stop()
```
WasabiFan, line 2090: good thinking to stop it :+1:

WasabiFan:
I'm liking the added functionality! it would be nice to have this easily accessible.

bpmerkel (changes requested):
In utils/line-follower-find..., the 2nd call to find_kp_ki_kd needs the speed parameter.

bpmerkel:
> In utils/line-follower-find..., the 2nd call to find_kp_ki_kd needs the speed parameter.

Also, in the same file... can it be refactored to use time instead of the datetime class? As datetime doesn't seem to work correctly with micropython.

ev3dev2/motor.py
dwalton76: It did not, I mainly followed this video as a guide
https://www.youtube.com/watch?v=AMBWV_HGYj4&t=1127s

dwalton76:
> In utils/line-follower-find..., the 2nd call to find_kp_ki_kd needs the speed parameter.

thanks...fixed

dwalton76:
> > In utils/line-follower-find..., the 2nd call to find_kp_ki_kd needs the speed parameter.
> 
> Also, in the same file... can it be refactored to use time instead of the datetime class? As datetime doesn't seem to work correctly with micropython.

I'll have this use the StopWatch class...it is a good abstraction for this sort of thing (it works with both python and micropython).

ev3dev2/motor.py
dwalton76: We could provide some basic `follow_` functions here in motor.py:
- `follow_forever`
- `follow_for_ms`
- `follow_for_distance`

Where the user passes in the name of the function for the one they wish to use.  We need to make this flexible though...am thinking of the FLL scenario where they need to do something more complicated like `follow_until_intersection_with_another_line`.

dwalton76: Actually we cannot do `follow_for_distance` here because this code is all in `MoveTank` where we do not know the wheel dimensions.  The user could always do a custom `follow_for_` function where they implement that.

ev3dev2/motor.py
```diff
@@ -1774,6 +1791,48 @@ def _block(self):
         self.wait_until_not_moving()
 
 
+# line follower classes
+class LineFollowError(Exception):
```
WasabiFan: `LineFollowErrorLostLine`? It's weird that this name is a substring of the name of another error but neither extends the other.

ev3dev2/motor.py
```diff
@@ -1774,6 +1791,48 @@ def _block(self):
...
+    """
+    return True
+
+
+def follow_for_ms(_tank, ms):
+    """
+    ``_tank``: the MoveTank object that is following a line
+    ``ms`` : the number of milliseconds to follow the line
+    """
+    global stopwatch
```
WasabiFan: this is scary... can the tank object store a stopwatch which is started when line following is started and provide an accessor?

WasabiFan (approved):
Minor comments, but otherwise looks good!

ev3dev2/motor.py
dwalton76: We could pass a stopwatch object (one that is created at the start of `follow_line`) to the `follow_for_XYZ` function...I like that more than the global variable :)

ev3dev2/motor.py
dwalton76: `LineFollowErrorLostLine` sounds good to me.  We could also have both of these inherit from a base `LineFollowError` class.

Issue #628, opened 2019-05-10
Micropython support for ev3dev2.sound


staadecker:
ev3dev2.sound does not support micropython. This is an issue because audio keys are useful while testing. For now I can use the following to make the robot beep.
 
```
import os
os.system("/usr/bin/beep")
```

However it would be nice to have support. I found that what is breaking the module is
`from subprocess import check_output, Popen, PIPE`.

dwalton76:
yeah the issue is with Sound.py using subprocess...I was reading this:
https://forum.micropython.org/viewtopic.php?t=5055

It sounds like you should be able to replace the subprocess calls with `os.fork()` or `os.system()` calls.  If you want to work on it assign this issue to yourself (or if you can't let me know and I can assign it to you).

staadecker:
Unfortunately, this is really not my area of expertise. I'm confused when looking at the code. For now I will use `os.system("/usr/bin/beep")`

JunOllyLi:
I had an experiment project, to improve ev3dev2 support in Micropython.
You guys can try it if you are interested.  I am not familiar with the Debian Packaging stuff, so this would require some hands on compiling.

https://github.com/robosaurus-ev3dev2-upy/upy-ev3dev2

So far I tested LEDs, sound, buttons and they work.  And of course the motors and external sensors were already supported.

dwalton76:
I'll take a stab at it

dwalton76:
@JunOllyLi wow you have made a lot of progress!! Have you thought about submitting pull requests for the changes you have made to the various repos?  It would be nice to pull your work in and have sound, buttons and LEDs all working in micropython for ev3dev-lang-python.

WasabiFan:
Yeah, @JunOllyLi we'd love to accept these changes into the official repos if you're interested. I haven't looked through it in depth, but I expect we'd be open to merging pretty much all of it. We'd really love the help for improved uPy support!

JunOllyLi:
Would love to, but if you looked at my changes, mostly they are in micropyton-lib and evdev. There only two very small quick hacks in ev3dev2, which I am not even satisfied although they work. At lease one of the changes should be fixed later with better PCRE implementation in Micropython-lib.
The change in evdev is a total re-written C code in Micropython module style, not going to be compatible with CPython code at all. Not sure if evdev guys would like to take it.
[...]

PR #632, opened 2019-05-13
micropython support for Sound

dwalton76:
For #628 

Tested micropython and python3 via
```

VOLUME = 30

sound = Sound()
sound.speak("hello world", volume=VOLUME, play_type=Sound.PLAY_WAIT_FOR_COMPLETE)

sound.play_song((
    ('Db4', 'e3'),
))

# Star Wars via play_song()
sound.play_song((
    ('D4', 'e3'),      # intro anacrouse
    ('D4', 'e3'),
    ('D4', 'e3'),
    ('G4', 'h'),       # meas 1
    ('D5', 'h'),
    ('C5', 'e3'),      # meas 2
    ('B4', 'e3'),
    ('A4', 'e3'),
    ('G5', 'h'),
    ('D5', 'q'),
    ('C5', 'e3'),      # meas 3
    ('B4', 'e3'),
    ('A4', 'e3'),
    ('G5', 'h'),
    ('D5', 'q'),
    ('C5', 'e3'),      # meas 4
    ('B4', 'e3'),
    ('C5', 'e3'),
    ('A4', 'h.'),
))

# Emperial March via tone()
sound.tone([
    (392, 350, 100), (392, 350, 100), (392, 350, 100), (311.1, 250, 100),
    (466.2, 25, 100), (392, 350, 100), (311.1, 250, 100), (466.2, 25, 100),
    (392, 700, 100), (587.32, 350, 100), (587.32, 350, 100),
    (587.32, 350, 100), (622.26, 250, 100), (466.2, 25, 100),
    (369.99, 350, 100), (311.1, 250, 100), (466.2, 25, 100), (392, 700, 100),
    (784, 350, 100), (392, 250, 100), (392, 25, 100), (784, 350, 100),
    (739.98, 250, 100), (698.46, 25, 100), (659.26, 25, 100),
    (622.26, 25, 100), (659.26, 50, 400), (415.3, 25, 200), (554.36, 350, 100),
    (523.25, 250, 100), (493.88, 25, 100), (466.16, 25, 100), (440, 25, 100),
    (466.16, 50, 400), (311.13, 25, 200), (369.99, 350, 100),
    (311.13, 250, 100), (392, 25, 100), (466.16, 350, 100), (392, 250, 100),
    (466.16, 25, 100), (587.32, 700, 100), (784, 350, 100), (392, 250, 100),
    (392, 25, 100), (784, 350, 100), (739.98, 250, 100), (698.46, 25, 100),
    (659.26, 25, 100), (622.26, 25, 100), (659.26, 50, 400), (415.3, 25, 200),
    (554.36, 350, 100), (523.25, 250, 100), (493.88, 25, 100),
    (466.16, 25, 100), (440, 25, 100), (466.16, 50, 400), (311.13, 25, 200),
    (392, 350, 100), (311.13, 250, 100), (466.16, 25, 100),
    (392.00, 300, 150), (311.13, 250, 100), (466.16, 25, 100), (392, 700)
])

sound.play_file("/usr/share/sounds/ev3dev/mechanical/airbrake.wav", volume=VOLUME, play_type=Sound.PLAY_NO_WAIT_FOR_COMPLETE)
```

### DIFF
--- a/ev3dev2/sound.py
+++ b/ev3dev2/sound.py
@@ -28,8 +28,13 @@ import sys
 if sys.version_info < (3, 4):
     raise SystemError('Must be using Python 3.4 or higher')
 
+from ev3dev2 import is_micropython
 import os
 import re
+
+if not is_micropython():
+    import shlex
+    from subprocess import Popen, PIPE
 
 
 def _make_scales(notes):
@@ -40,6 +45,25 @@ def _make_scales(notes):
         for n in note.split('/'):
             res[n] = freq
     return res
+
+
+
+def get_command_processes(command):
+    # We must split command into sub-commands to support pipes
+    if "|" in command:
+        command_parts = command.split("|")
+    else:
+        command_parts = [command]
+
+    processes = []
+
+    for command_part in command_parts:
+        if processes:
+            processes.append(Popen(shlex.split(command_part), stdin=processes[-1].stdout, stdout=PIPE, stderr=PIPE))
+        else:
+            processes.append(Popen(shlex.split(command_part), stdin=None, stdout=PIPE, stderr=PIPE))
+
+    return processes
 
 
 class Sound(object):
@@ -85,6 +109,46 @@ class Sound(object):
         assert play_type in self.PLAY_TYPES, \
             "Invalid play_type %s, must be one of %s" % (play_type, ','.join(str(t) for t in self.PLAY_TYPES))
 
+    def _audio_command(self, command, play_type):
+        if is_micropython():
+
+            if play_type == Sound.PLAY_WAIT_FOR_COMPLETE:
+                os.system(command)
+
+            elif play_type == Sound.PLAY_NO_WAIT_FOR_COMPLETE:
+                os.system('{} &'.format(command))
+
+            elif play_type == Sound.PLAY_LOOP:
+                while True:
+                    os.system(command)
+
+            else:
+                raise Exception("invalid play_type " % play_type)
+
+            return None
+
+        else:
+            with open(os.devnull, 'w') as n:
+
+                if play_type == Sound.PLAY_WAIT_FOR_COMPLETE:
+                    processes = get_command_processes(command)
+                    processes[-1].communicate()
+                    processes[-1].wait()
+                    return None
+
+                elif play_type == Sound.PLAY_NO_WAIT_FOR_COMPLETE:
+                    processes = get_command_processes(command)
+                    return processes[-1]
+
+                elif play_type == Sound.PLAY_LOOP:
+                    while True:
+                        processes = get_command_processes(command)
+                        processes[-1].communicate()
+                        processes[-1].wait()
+
+                else:
+                    raise Exception("invalid play_type " % play_type)
+
     def beep(self, args='', play_type=PLAY_WAIT_FOR_COMPLETE):
         """
         Call beep command with the provided arguments (if any).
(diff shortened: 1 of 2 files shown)


WasabiFan:
This changes the API, right? It removes one of the "play type" options and removes the handle returns?

dwalton76:
All of the `play_type` options are still the same.  The return value has changed if you are doing `play_type=Sound.PLAY_NO_WAIT_FOR_COMPLETE`...that used to return the PID of the subprocess but now we just return `None`.  Our default is `Sound.PLAY_WAIT_FOR_COMPLETE`...my gut says there probably aren't a ton of folks using the `NO_WAIT` option and doing much with that returned PID.  With the proposed change they could always figure out the PID by looking at the processes that are currently running.

WasabiFan:
Ok, so am I correct in thinking that with this change there isn't an API to play a sound asynchronously (i.e., start it then keep going while it plays)? I feel like that behavior could be useful, as it might be nice to speak/beep some status while your robot is still going.

dwalton76:
No, that functionality is still there, to do that you use `play_type=Sound.PLAY_NO_WAIT_FOR_COMPLETE`.

WasabiFan:
I'm still a bit concerned about the breaking change on the return type. Perhaps this is unfounded, but it feels like this might cause problems for existing consumers who do something more advanced.

I think we could instead split the `Popen` stuff into its own method (i.e., just have it as a generic "run audio command" method) and then have it fall back to `os.system` when needed. That would probably be less of a code change, would preserve the existing behavior, and introduce support for the additional platform.

I think this would probably work:

```python
def _audio_command(self, command, play_type=PLAY_WAIT_FOR_COMPLETE):
    if is_micropython():
        if play_type == Sound.PLAY_WAIT_FOR_COMPLETE:
            os.system(command)
        else:
            os.system('{} &'.format(command))
        
        return None
    else:
        with open(os.devnull, 'w') as n:
            subprocess = Popen(shlex.split(command), stdout=n)
            if play_type == Sound.PLAY_WAIT_FOR_COMPLETE:
                subprocess.wait()
                return None
            else:
                return subprocess
```

It should be fairly straightforward to rewrite the other functions in terms of this, and that would get all the behavior we need.

dwalton76:
I think this is backwards compatible now.  The only quirk is that the `return` behavior documented in the docstrings is incorrect for micropython...it will not return the PID of the subprocess.  Do we want to document python3 vs micropython behavior?  I don't know that we do that elsewhere.

ev3dev2/sound.py
```diff
@@ -44,15 +47,29 @@ def _make_scales(notes):
     return res
 
 
+
+def get_command_processes(command):
```
WasabiFan, line 51: Ah, got it; this is for the speak method which uses pipes?

ev3dev2/sound.py
```diff
@@ -370,10 +388,10 @@ def get_volume(self, channel=None):
         if channel is None:
             channel = self._get_channel()
 
-        out = check_output(['amixer', 'get', channel]).decode()
-        m = re.search(r'\[(?P<volume>\d+)%\]', out)
+        out = os.popen(['/usr/bin/amixer', 'get', channel]).read()
```
WasabiFan, line 396: this is supported on Micropython as well?

WasabiFan (approved):
Nice! This seems reasonable to me; I think it would make the most sense to add a note to those doc comments saying that it doesn't return the object on Micropython.

ev3dev2/sound.py
dwalton76: yep

ev3dev2/sound.py
dwalton76: yep, I tested every Sound feature I could find and they all work with python3 and micropython

dwalton76:
Will update the docstrings to mention about the return values when using python3 vs micropython and then merge this one.

PR #634, opened 2019-05-13
micropython support for LED animations

dwalton76:
For #633 

Tested python3 and micropython via
```
from ev3dev2.led import Led, Leds
from time import sleep
import logging
import sys

# logging

my_leds = Leds()

for x in range(4):
    if x == 0:
        my_leds.animate_police_lights('RED', 'GREEN')
    elif x == 1:
        my_leds.animate_flash('AMBER')
    elif x == 2:
        my_leds.animate_cycle(('RED', 'GREEN', 'AMBER'))
    elif x == 3:
        my_leds.animate_rainbow()

my_leds.reset()
```

### DIFF
--- a/ev3dev2/stopwatch.py
+++ b/ev3dev2/stopwatch.py
@@ -5,7 +5,7 @@ A StopWatch class for tracking the amount of time between events
 try:
     import datetime as dt
     micropython = False
-except ImportError:
+except Exception:
     import utime
     micropython = True
 
--- a/ev3dev2/stopwatch.py
+++ b/ev3dev2/stopwatch.py
@@ -2,16 +2,16 @@
 A StopWatch class for tracking the amount of time between events
 """
 
-try:
-    import datetime as dt
-    micropython = False
-except Exception:
+from ev3dev2 import is_micropython
+
+if is_micropython():
     import utime
-    micropython = True
+else:
+    import datetime as dt
 
 
 def get_ticks_ms():
-    if micropython:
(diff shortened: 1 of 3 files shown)


ev3dev2/stopwatch.py
```diff
@@ -5,7 +5,7 @@
 try:
     import datetime as dt
     micropython = False
-except ImportError:
```
WasabiFan, line 8: We sure we want to do that?

ev3dev2/stopwatch.py
```diff
@@ -5,7 +5,7 @@
...
+except Exception:
     import utime
     micropython = True
```
WasabiFan: isn't there already a function we have to check this? And/or a uPy builtin?

ev3dev2/stopwatch.py
dwalton76: yeah it is a bit odd but on my laptop (ubuntu) it raises `ImportError` but on ev3dev it raised something else (`AttributeError` maybe) so to play it safe I just caught all Exceptions.

ev3dev2/stopwatch.py
dwalton76: ah you are right, we have
```
def is_micropython():
    return sys.implementation.name == "micropython"
```

Will use that

WasabiFan (approved):
Looks good to me, thanks for figuring these things out!

Issue #571, opened 2018-11-29
Micropython support for ev3dev2 Button class


Dilbert76:
I'd like to propose to add ev3dev2.Button support for Micropython.

Buttons are often used by WRO or FLL teams to calibrate motors. Furthermore WRO Regular rules allow to start a program beforehand and start the actual run by pushing the center button on the EV3 brick. This trick would allow WRO teams to initialise/construct all motor and sensor objects (which still takes several seconds) before the actual run, so robot would move instantaneously  as soon as center button is pressed without having to wait a couple of seconds before the robot actually starts doing something. Without Button support I'm afraid that it might be tough for WRO teams to use Python...

WasabiFan:
@dlech On the technical side, do you have any suggestions on the best way to do this? The button module uses `fcntl` which should be fairly easy to replace with the micropython equivalent, but I expect `evdev` (for device discovery and events) to be less doable. Is there a logical way that you see?

dlech:
If `micropython-lib` has something compatible with the standard `struct` module, we could probably skip using the existing python `evdev` package and roll our own simplified version.

See <https://github.com/dlech/micropython-ev3dev/blob/ev3dev-stetch/src/uev3dev/button.py>. This isn't compatible with standard python, but it could possibly be adapted.

staadecker:
Any news on this? @dlech Is there any way for me to use the button class from the link since it works. What if I want my code to be compatible with both normal and micro. Should I have an if statement to differentiate between which modules to use?

dlech:
IMHO, trying to make code that is compatible with both Python 3 and MicroPython is not worth the effort. They are different programming languages and there are too many gotchas to deal with because of all of the little differences.

staadecker:
Ok makes sense. Any plan to merge your code from micropython-ev3dev into the main repo and then people can import one library or the other depending on their needs? As of now, would the best be to clone the micropython repo onto the ev3 and add it to the path?

JunOllyLi:
I had an experiment project, to improve ev3dev2 support in Micropython.
You guys can try it if you are interested.  I am not familiar with the Debian Packaging stuff, so this would require some hands on compiling.

https://github.com/robosaurus-ev3dev2-upy/upy-ev3dev2

So far I tested LEDs, sound, buttons and they work.  And of course the motors and external sensors were already supported.

dwalton76:
@JunOllyLi had you worked on Display support in your project?

[...]

PR #639, opened 2019-05-24
Micropython buttons support

dwalton76:
For issue #571


dwalton76:
Huge credit with this patch goes to David's work at https://github.com/dlech/micropython-ev3dev/blob/ev3dev-stetch/src/uev3dev/button.py

dwalton76:
Not sure why the test cases are failing for micropython, they pass when I run them locally via `micropython tests/api_tests.py`.  Will dig some more later...

ev3dev2/button.py
```diff
@@ -337,3 +197,268 @@ def backspace(self):
...
+        raise Exception("micropython button support has not been implemented for '%s'" % platform)
+
+
+    def _test_bit(buf, index):
+        byte = buf[int(index >> 3)]
+        bit = byte & (1 << (index % 8))
+        return bool(bit)
+
+
+    class ButtonBase(ButtonCommon):
```
WasabiFan, line 223: What's this for?

WasabiFan (approved):
Nice! To make sure the API didn't change I [diffed the docs output](https://gist.github.com/WasabiFan/d3c3d7c93ffd0d140655bde815decd6a) and aside from some reordering and wording changes it looks like the API surface is the same. 

So once you figure out what happened to the CI build (maybe an independent package published an update that broke us?) I think this looks OK. Just make sure the general stuff has been tested on both platforms. :+1:

Thanks!

ev3dev2/button.py
dwalton76: The InfraredSensor class inherits from this.

dwalton76: yep I tested all the basics on python3 and Micropython. Will look into the CI issues tonight.

dwalton76:
For some reason the `dst` directory isn't getting cleaned up after the first test runs (I have no clue as to why yet) so all of the others fail at `shutil.copytree(src, dst)` because the `dst` directory is already there.  I submitted a PR here:
https://github.com/ev3dev/ev3dev-lang-fake-sys/pull/14

to delete the `dst` directory if it is already there.

dwalton76:
Long story short `shutil.rmtree` is broken in micropython v1.9.4 and that is why the directory isn't cleaned up.  Possible work-around in https://github.com/ev3dev/ev3dev-lang-fake-sys/pull/15

dwalton76:
woohoo finally a clean test run!! :)

PR #641, opened 2019-05-27
Update fake-sys with EEXIST fix

WasabiFan:


WasabiFan:
@dwalton76 It looks like the fix has a bug: https://travis-ci.org/ev3dev/ev3dev-lang-python/jobs/537604548

dwalton76:
Is it that `shutil.rmtree` in micropython is just broken?

dwalton76:
The fix for this is https://github.com/ev3dev/ev3dev-lang-fake-sys/pull/15 which we will pick up via https://github.com/ev3dev/ev3dev-lang-python/pull/639 so that the test cases for that PR will pass.  Will close this PR.

Issue #635, opened 2019-05-13
Micropython support for Display


dwalton76:
Take a look at what David did here:
https://github.com/dlech/micropython-ev3dev/blob/ev3dev-stetch/src/uev3dev/display.py

and see if we can bring that in to add Display support for micropython

dwalton76:
@WasabiFan I am going to experiment with David’s Micropython Display code. Maybe this will be like the buttons where we can add this without a ton of effort :)

WasabiFan:
The major issue I see here is that most of our current display API is really just exposing the PILlow drawing object. I think we do still have wrapper methods around it though, so maybe we can fill in some holes there and then have the PIL object only available on CPython.

dwalton76:
I think if we could get just text to the display working that would be worth committing.  Drawing shapes, etc could all come later if needed.

Am doing my work here if you want to give it a try. I can't actually get anything to display though (using `Display.text_pixels`)...not sure why yet
https://github.com/dwalton76/ev3dev-lang-python/tree/micropython-display

dwalton76:
Hmm this looks interesting
https://docs.micropython.org/en/latest/library/framebuf.html
```
import framebuf

# FrameBuffer needs 2 bytes for every RGB565 pixel
fbuf = FrameBuffer(bytearray(10 * 100 * 2), 10, 100, framebuf.RGB565)

fbuf.fill(0)
fbuf.text('MicroPython!', 0, 0, 0xffff)
fbuf.hline(0, 10, 96, 0xffff)
```

dwalton76:
oh wait David's code was already using that...doh!!

bpmerkel:
Okay. Just in case anyone is watching this space... You can display and position text on the LCD screen using escape code sequences supported by the EV3DEV console driver (appears to support the VT100 sequences). Here are some code snippets that'll get you started:
~~~~
def set_cursor(on=False):
    # use escape code to turn cursor on or off
    if on:
        print("\x1b[?25h", end='')
    else:
        print("\x1b[?25l", end='')


def clear_screen():
    print("\x1b[2J\x1b[H", end='')
    set_cursor(False)


def text_at(row, col, text='', reverse=False):
    if reverse:
        text = "\x1b[7m%s\x1b[m" % (text)

    print("\x1b[%d;%dH%s" % (row, col, text), end='')


def enable_big_text(font="Lat15-TerminusBold24x12"):
    """
[...]

PR #647, opened 2019-07-20
Display() text-only implementation for micropython

bpmerkel:
For issue #635 `Micropython support for Display`

This is a text-only implementation using the EV3 LCD console ANSI code support to position the cursor, clear the screen, show text with different fonts, and use reverse (white text on black background, versus the default black text on white background). See the file `utils/display_fonts.py` to see the fonts supported with sample calls to the `ev3dev2.Display()` class.


bpmerkel:
I also included a Makefile so folks can `git clone`, build and install the latest ev3dev2 modules to the /usr/lib/micropython directory.

PR #648, opened 2019-07-20
Display() text-only implementation for micropython

bpmerkel:
For issue #635 Micropython support for Display

This is a text-only implementation using the EV3 LCD console ANSI code support to position the cursor, clear the screen, show text with different fonts, and use reverse (white text on black background, versus the default black text on white background). See the file utils/display_fonts.py to see the fonts supported with sample calls to the ev3dev2.Display() class.

Also included a Makefile so folks can git clone, build and install the latest ev3dev2 modules to the /usr/lib/micropython directory on their EV3.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,13 @@
+OUT := build
+MPYCROSS := /usr/bin/mpy-cross
+MPYC_FLAGS := -v -v -mcache-lookup-bc
+PYS := $(shell find ./ev3dev2 -type f \( -iname "*.py" ! -iname "setup.py" \))
+MPYS := $(PYS:./%.py=${OUT}/%.mpy)
+vpath %.py .
+
+${OUT}/%.mpy: %.py
+	@mkdir -p $(dir $@)
+	${MPYCROSS} ${MPYC_FLAGS} -o $@ $<
+
+install: ${MPYS}
+	cp -R $(OUT)/* /usr/lib/micropython/
--- a/docs/micropython.md
+++ b/docs/micropython.md
@@ -12,8 +12,8 @@ Module                          Support status
 ==============================  =================
 `ev3dev2.button`                ️️✔️
 `ev3dev2.control` [1]_          ⚠️
-`ev3dev2.display`               ❌
-`ev3dev2.fonts` [2]_            ⚠️
+`ev3dev2.display` [2]_          ⚠️
+`ev3dev2.fonts`   [3]_          ❌
 `ev3dev2.led`                   ✔️
 `ev3dev2.motor`                 ✔️
 `ev3dev2.port`                  ✔️
@@ -25,7 +25,8 @@ Module                          Support status
 ==============================  =================
 
 .. [1] Untested/low-priority, but some of it might work.
-.. [2] It might work, but isn't useful without ``ev3dev2.display``.
+.. [2] Display() works for text only, using ANSI codes to the console LCD screen.
+.. [3] The fonts module is replaced with the ability to change the console font using the Display() class.
 ```
 
 ## Differences from standard Python (CPython)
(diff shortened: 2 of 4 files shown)


Makefile
```diff
@@ -0,0 +1,13 @@
...
+MPYC_FLAGS := -v -v -mcache-lookup-bc
+PYS := $(shell find ./ev3dev2 -type f \( -iname "*.py" ! -iname "setup.py" \))
+MPYS := $(PYS:./%.py=${OUT}/%.mpy)
+vpath %.py .
+
+${OUT}/%.mpy: %.py
+	@mkdir -p $(dir $@)
+	${MPYCROSS} ${MPYC_FLAGS} -o $@ $<
+
+install: ${MPYS}
```
dwalton76, line 12: Maybe call this `micropython-install` so  `install` could do `python3 setup.py install`?

docs/micropython.md
```diff
@@ -36,10 +37,42 @@ See [the MicroPython differences page](http://docs.micropython.org/en/latest/gen
...
+
+### Building and installing the latest EV3DEV2 module on your EV3
+
+In an SSH Terminal window with an EV3 with Internet access, run the following commands:
+(recall that the `sudo` password is `maker`)
+
+```shell
+git clone https://github.com/ev3dev/ev3dev-lang-python.git
+cd ev3dev-lang-python
+sudo make
```
dwalton76, line 69: `sudo make micropython-install`

docs/micropython.md
```diff
@@ -36,10 +37,42 @@ See [the MicroPython differences page](http://docs.micropython.org/en/latest/gen
...
+```
+
+To update the module, use the following commands:
+
+```shell
+cd ev3dev-lang-python
+git pull
+sudo make
```
dwalton76, line 77: same here

ev3dev2/display.py
```diff
@@ -23,420 +23,519 @@
...
+
+        def set_font(self, font):
+            """
+            Set the LCD console font
+            Parameter:
+                `font` (string): Font name, as found in /usr/share/consolefonts/
+           """
+            self._font = font
+            os.system("setfont %s" % (font))
+            self.reset_screen()
```
dwalton76, line 49: Can we avoid the `reset_screen()` here...wouldn't than open the door for the user drawing in different fonts on different sections of the screen?

ev3dev2/display.py
```diff
@@ -23,420 +23,519 @@
...
+                `on` (bool): ``True`` to turn on the screen cursor; default is ``False``
+            """
+            # use escape code to turn cursor on or off
+            print("\x1b[?25%s" % ('h' if on else 'l'), end='')
+
+        def update(self):
+            """
+            NOTE: Not implemented for micropython
+            """
+            pass
```
dwalton76, line 64: For all of the methods that are not implemented I think it would be good to have a
```
log.info("XYZ not implemented for micropython")
```
that way the user is aware they are trying to use something that is not supported.

ev3dev2/display.py
```diff
@@ -23,420 +23,519 @@
...
+
+        def text_pixels(self, text, clear_screen, x, y, text_color=False, font=None):
+            """
+            NOTE: Not implemented for micropython
+            Draw text on the screen
+
+            Parameters:
+                text (str): The text to draw
```
dwalton76, line 72: I'd be fine if we dropped all of the parameters definitions in the docstrings for the ones that are not implemented.

dwalton76: If we do keep these (not a biggie to me either way) we should make the use of the backtick ` consistent.  This docstring isn't using them for the variables but others are...not all of them though.

ev3dev2/display.py
```diff
@@ -23,420 +23,519 @@
...
+
+        def text_grid(self, text, clear_screen=False, x=1, y=1, text_color=False, font=None):
+            """
+            Display 'text' (string) starting at grid (x, y). Note that the grid location is 1-based (not 0-based).
+
+            Depending on the font, the EV3 LCD display can show text within 4 to 21 rows, and 11 to 44 columns.
+            The default font results in a grid that is 14 columns wide and 5 rows tall.
+
+            Parameters:
+                'text' (string): Text to display
```
dwalton76, line 90: This one is using a single quote instead of a backtick

ev3dev2/display.py
```diff
@@ -23,420 +23,519 @@
...
+
+    import mmap
+    import ctypes
+    import logging
+    from PIL import Image, ImageDraw
+    from . import fonts
+    from . import get_current_platform, library_load_warning_message
+    from struct import pack
+    
+    log = logging.getLogger(__name__)
```
dwalton76, line 139: Would keep the logging parts at the top since those will be common to both

dwalton76:
I forgot to mention you should add an entry to `debian/changelog` for your change

Makefile
bpmerkel: Good. Done. But committed in the Console() PR #649.

docs/micropython.md
bpmerkel: Good. Done. But committed in the Console() PR #649.

docs/micropython.md
bpmerkel: Good. Done. But committed in the Console() PR #649.

ev3dev2/display.py
bpmerkel: Agreed. Fixed in the Console() PR #649.

bpmerkel:
Closing this PR in favor of Console() PR #649. The Console() class is better off as a separate class, as it works for both Python and MicroPython, and thus doesn't quite fit within the Display() class for cross-language support.

PR #654, opened 2019-08-06
Console() rows/columns, echo, and cursor properties; doc corrections

bpmerkel:
Enhanced Console() to add `rows`, `columns`, and `echo` properties; convert the `show_cursor` method to a `cursor` property; and corrected the docs to reflect the changes, as well as the main README to properly render the link to the official API doc site. Also added a `console_menu` utility to demonstrate the Console(), Led() and Button() classes to create a fast mission-switcher for FIRST LEGO League teams to save precious time between missions (without having to hunt and peck at Brickman to find their next file to run).


ev3dev2/console.py
```diff
@@ -38,7 +38,62 @@ def __init__(self, font="Lat15-TerminusBold24x12"):
 
         """
         self._font = None
-        self.set_font(font, False)  # don't reset the screen during construction
+        self._columns = 0
+        self._rows = 0
+        self._echo = False
+        self._cursor = False
+        self.set_font(font, reset_console=False)  # don't reset the screen during construction
+        self.cursor = False
```
WasabiFan, line 46: Is there a reason to set both `_cursor` and `cursor`?

ev3dev2/console.py
```diff
@@ -38,7 +38,62 @@ def __init__(self, font="Lat15-TerminusBold24x12"):
...
+
+    @property
+    def echo(self):
+        """
+        Return (bool) whether the console echo mode is enabled.
+        """
+        return self._echo
+
+    @rows.setter
+    def echo(self, value):
```
WasabiFan, line 71: `rows.setter`? Shouldn't that be `echo` (here and below)?

ev3dev2/console.py
```diff
@@ -105,21 +160,13 @@ def set_font(self, font, reset_console=True):
         if font is not None and font != self._font:
             self._font = font
             os.system("setfont {}".format(font))
+            rows, columns = os.popen('stty size').read().strip().split(" ")
```
WasabiFan, line 160: :+1: nice!

utils/console_menu.py
```diff
@@ -0,0 +1,209 @@
...
+def menu(choices, before_run_function=None, after_run_function=None):
+    """
+    Console Menu that accepts choices and corresponding functions to call.
+    The user must press the same button twice: once to see their choice highlited,
+    a second time to confirm and run the function. The EV3 LEDs show each state change:
+    Green = Ready for button, Amber = Ready for second button, Red = Running
+
+    Parameters:
+
+    - `choices` a dictionary of tuples "button-name": ("mission-name", function-to-call)
```
WasabiFan: I don't think this has to be specific to missions; one could use this for other menus, right?

utils/console_menu.py
```diff
@@ -0,0 +1,209 @@
+#!/usr/bin/env micropython
+from time import sleep
+from sys import stderr
```
WasabiFan: What's the expected way people will use this -- is it an example template to start from, or should people be importing it? It seems like the former, given the main function and template code; perhaps we should move it to the demo repo. But I'm open to discussion.

WasabiFan (commented):
Thanks for the contributions! Some miscellaneous comments...

WasabiFan:
In #655 @dlech suggested we publish a new release this week; I'd love to get these changes in over the next few days to include. And since I haven't gotten to actually test the `Console` I'll get to do that too now that I'll be back home. I can help out with some of the changes here, but others are questions which I wasn't sure of; let me know how I can help.

ev3dev2/console.py
bpmerkel: My linter says that class-level fields should be instantiated/initialized in the constructor. Agree that it's redundant.

ev3dev2/console.py
bpmerkel: Good catch! Yes. Fix is imminent. Looks like a copy pasta mistake.

ev3dev2/console.py
bpmerkel: LOL. Would rather use python-esque termio, but termio isn't yet implemented for micropython.

utils/console_menu.py
bpmerkel: Yes. It's an example utility to demonstrate a way to create a 5-button menu sequencer--which will be in high demand for _FIRST_ LEGO League teams; as they are confined to 2 and a half minutes to switch quickly between "mission" routines... The hunt-and-peck of EV3-G and Brickman significantly impact points-per-second metric... so a "menu" is essential--especially with the long launch times of python and (lesser) micropython programs.

utils/console_menu.py
bpmerkel: Sure... its a nifty routine to tie-in functions to the EV3 buttons... and uses the console and LEDs to provide feedback when selecting and confirming a choice. For _FIRST_ LEGO League, we refer to the scoring runs as "missions"... and they can chain them together to accomplish multiple models in the same "sortie".

utils/console_menu.py
bpmerkel: It would be cool to craft an "advanced" module for these types of classes that reference multiple sensors, other classes, etc. I suspect that folks will indeed compose their own "Robot()" class that closely reflects their specific robot (i.e. my EV3 robot from last year has large motors in B and C, medium motors in A and D, Gyro in 1, Color Sensors in 2 and 3, and a button in 4.) I envision teams implementing their own custom logic in their Robot() class that integrates abstract and advanced capabilities (as follow-line right but stop when left color sensor sees black; or go straight for 2.5 rotations, but ramp up speed, and ramp down near the end to avoid having the robot 'jerk'.)

WasabiFan: What I was asking was more about this particular PR: is this code really appropriate for `utils`, or should it instead go somewhere else, given that it isn't intended to be run standalone/unmodified?

I get the use case for FLL -- I previously was on an FLL team and  recognize the need. I'm just trying to figure out the most appropriate code organization to not be confusing.

WasabiFan:
Referring to this thread: https://github.com/ev3dev/ev3dev-lang-python/pull/654#discussion_r314122966 

After thinking about it some more, I think the new menu code should live either in the [demo repo](https://github.com/ev3dev/ev3dev-lang-python-demo) or in the docs. We could make an importable `Menu` class and provide it built-in to the library, which I think would be a good idea (for a future PR). But given that the `console_menu.py` in this PR requires code modification before use in one's own project, I don't think it is a good fit for the `utils/` dir in this repo as it stands.

bpmerkel:
> Referring to this thread: [#654 (comment)](https://github.com/ev3dev/ev3dev-lang-python/pull/654#discussion_r314122966)
> 
> After thinking about it some more, I think the new menu code should live either in the [demo repo](https://github.com/ev3dev/ev3dev-lang-python-demo) or in the docs. We could make an importable `Menu` class and provide it built-in to the library, which I think would be a good idea (for a future PR). But given that the `console_menu.py` in this PR requires code modification before use in one's own project, I don't think it is a good fit for the `utils/` dir in this repo as it stands.

I concur. Exclude it from the PR. I can submit a different PR as a new class in ev3dev2, or for the demo repo... I have other demo code too with a new Robot() class wrapper to encapsulate a robot's specific ports, speeds, wheel sizes, with useful properties meant to serve as singleton references. I also have a more advanced drive module with Gyro proportional turns, Gyro-based moves, and moves with ramp-up/ramp-down to avoid annoying robot jerk from torque/momentum.

utils/console_menu.py
bpmerkel: Yes correct... I was thinking of a new util to be a better replacement to the clunky brickman show sensors tool... use the menu instead to show sensor values, operate the motors, etc. There should be a cool way to nest menus within a top-level menu to create a menu tree for quick navigation--may need a "single-press" vs. "double-press" option to make it even simpler.

WasabiFan:
Thanks! Those other items you mentioned also sound great -- I'd love to have real tested control capabilities and demos. Right now we don't get a lot of feedback on how our, e.g., driving primitives perform in real-world use cases -- your contributions and comments validate that at least most of it is doing the right thing.

PR #658, opened 2019-08-18
add console and stopwatch to rules

bpmerkel:
Found that console.mpy and stopwatch.mpy were excluded from the Beta4 release deployment. Suspect that adding them to the `rules` file will fix that!


WasabiFan:
Good catch! I'll probably publish a `beta5` now to include them.

PR #665, opened 2019-08-23
LED animation fix duration None

mvolfik:
- allow passing `duration=None` to `animate_police_lights` to animate forever, as stated in docs
- add animations reset to `all_off`, as without it the leds light up on the next animation cycle

### DIFF
--- a/ev3dev2/led.py
+++ b/ev3dev2/led.py
@@ -415,7 +415,7 @@ class Leds(object):
                     self.set_color(group1, color2)
                     self.set_color(group2, color1)
 
-                if self.animate_thread_stop or (duration is not None and stopwatch.value_ms >= duration_ms):
+                if self.animate_thread_stop or (duration_ms is not None and stopwatch.value_ms >= duration_ms):
                     break
 
                 even = not even
--- a/ev3dev2/led.py
+++ b/ev3dev2/led.py
@@ -403,8 +403,7 @@ class Leds(object):
         def _animate_police_lights():
             self.all_off()
             even = True
-            if duration is not None:
-                duration_ms = duration * 1000
+            duration_ms = duration * 1000 if duration is not None else None
             stopwatch = StopWatch()
             stopwatch.start()
 
--- a/ev3dev2/led.py
+++ b/ev3dev2/led.py
@@ -401,7 +401,8 @@ class Leds(object):
         def _animate_police_lights():
             self.all_off()
             even = True
-            duration_ms = duration * 1000
+            if duration != None:
+                duration_ms = duration * 1000
             stopwatch = StopWatch()
             stopwatch.start()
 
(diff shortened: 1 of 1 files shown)


ev3dev2/led.py
```diff
@@ -401,7 +403,8 @@ def animate_police_lights(self, color1, color2, group1='LEFT', group2='RIGHT', s
         def _animate_police_lights():
             self.all_off()
             even = True
-            duration_ms = duration * 1000
+            if duration is not None:
+                duration_ms = duration * 1000
```
WasabiFan: Can we make this variable `None` rather than leaving it undeclared? Something like 

```python
duration_ms = duration * 1000 if duration is not None else None
```

WasabiFan (approved):
Good catches! One minor comment and then I'll merge. Thanks!

ev3dev2/led.py
mvolfik: Ok

ev3dev2/led.py
```diff
@@ -413,7 +415,7 @@ def _animate_police_lights():
                     self.set_color(group1, color2)
                     self.set_color(group2, color1)
 
-                if self.animate_thread_stop or stopwatch.value_ms >= duration_ms:
+                if self.animate_thread_stop or (duration is not None and stopwatch.value_ms >= duration_ms):
```
dwalton76: checking for `duration_ms is not None and ...` would read a little better here since the next part of the `and` deals with `duration_ms`.

The rest looks good to me.  Thank you for the PR.

mvolfik: I was thinking about it too, but I decided to compare with duration, as "duration is none --> no time limit specified, don't check elapsed time". But if you both feel like this is better, I don't have problem with it. Edited

WasabiFan:
Two things, now that I look closer.

Most importantly, shouldn't this change be made to all the animations, not just police lights?

Second, can't we move the duration calculation to be within the if condition so we can eliminate the need for propagating `None`?

mvolfik:
> Most importantly, shouldn't this change be made to all the animations, not just police lights?

Oh, of course, I didn't notice this issue is in other functions too, I just came across this. when I used it. Will do it
> Second, can't we move the duration calculation to be within the if condition so we can eliminate the need for propagating None?

Yeah, it will simplify the code, however we would calculate the same thing in every cycle. Not a big deal, but my IT teacher would bite me :). Just tell me what you prefer and I will do it the same way in every animation function...

mvolfik:
Also shouldn't we issue warning/imply `block=False`/something when user tries to animate forever in main thread? Or that's responsibility we keep on the user?

WasabiFan:
> Just tell me what you prefer and I will do it the same way in every animation function...

I figure this is probably a common use case for the Stopwatch anyway (checking if a time has elapsed) so I opened #667 to add such a method directly to the StopWatch. Once that is reviewed and merged you can use the new methods here. Sorry for making this take so long, we appreciate the contribution!

> Also shouldn't we issue warning/imply block=False/something when user tries to animate forever in main thread? Or that's responsibility we keep on the user?

If it were clear-cut I might want to add such a warning, but consider this: in my main thread I spawn a new thread which runs my robot then my main thread just blocks running LEDs indefinetely. I'd say that's still valid (if a bit weird), so I'd prefer to just keep it simple here. If someone blocks their main thread unintentionally they'll either figure it out for themselves or ask for help and we can tell them.

PR #667, opened 2019-08-26
Add seconds and elapsed checks to stopwatch

WasabiFan:
Also did general clean-up and added documentation. Removed intermediate "value" so that accessors don't modify state.

There is one breaking change (I believe everything else is backwards-compatible): "value_hms" now returns the unstringified tuple, and a new "hms_str" returns the string version. Also changed the `__str__` format to include the hms string but I don't view that as a breaking change.

Useful for #665.

### DIFF
--- a/ev3dev2/stopwatch.py
+++ b/ev3dev2/stopwatch.py
@@ -25,7 +25,7 @@ class StopWatch(object):
         Initializes the StopWatch but does not start it.
 
         desc:
-            A string description to pring when stringifying.
+            A string description to print when stringifying.
         """
         self.desc = desc
         self._value = 0
--- a/ev3dev2/stopwatch.py
+++ b/ev3dev2/stopwatch.py
@@ -15,6 +15,13 @@ def get_ticks_ms():
     else:
         return int(dt.datetime.timestamp(dt.datetime.now()) * 1000)
 
+class StopWatchAlreadyStartedException(Exception):
+    """
+    Exception raised when start() is called on a StopWatch which was already start()ed and not yet
+    stopped.
+    """
+    pass
+
 class StopWatch(object):
     """
     A timer class which lets you start timing and then check the amount of time
@@ -38,7 +45,12 @@ class StopWatch(object):
     def start(self):
         """
         Starts the timer. If the timer is already running, resets it.
+
+        Raises a :py:class:`ev3dev2.stopwatch.StopWatchAlreadyStartedException` if already started.
         """
+        if self.is_started:
+            raise StopWatchAlreadyStartedException()
+
         self._stopped_total_time = None
         self._start_time = get_ticks_ms()
 
@@ -54,10 +66,16 @@ class StopWatch(object):
 
     def reset(self):
         """
-        Resets the timer, and starts it again.
+        Resets the timer and leaves it stopped.
         """
         self._start_time = None
         self._stopped_total_time = None
+    
+    def restart(self):
+        """
+        Resets and then starts the timer.
+        """
+        self.reset()
         self.start()
     
     @property
--- a/ev3dev2/stopwatch.py
+++ b/ev3dev2/stopwatch.py
@@ -9,76 +9,114 @@ if is_micropython():
 else:
     import datetime as dt
 
-
 def get_ticks_ms():
     if is_micropython():
         return utime.ticks_ms()
     else:
         return int(dt.datetime.timestamp(dt.datetime.now()) * 1000)
 
-
 class StopWatch(object):
-
+    """
+    A timer class which lets you start timing and then check the amount of time
+    elapsed.
+    """
     def __init__(self, desc=None):
+        """
+        Initializes the StopWatch but does not start it.
+
+        desc:
+            A string description to pring when stringifying.
+        """
         self.desc = desc
         self._value = 0
-        self.start_time = None
-        self.prev_update_time = None
+        self._start_time = None
+        self._stopped_total_time = None
 
     def __str__(self):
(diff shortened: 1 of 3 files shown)


ev3dev2/stopwatch.py
```diff
@@ -9,76 +9,114 @@
...
+    """
+    A timer class which lets you start timing and then check the amount of time
+    elapsed.
+    """
     def __init__(self, desc=None):
+        """
+        Initializes the StopWatch but does not start it.
+
+        desc:
+            A string description to pring when stringifying.
```
mvolfik:
```suggestion
            A string description to print when stringifying.
```

mvolfik:
Wouldn't `reset` and `start` after this do the same thing? I would suggest that `start` would continue from the time when `stop` was called, or from 0 if `_stopped_total_time` is `None`, and reset would only reset `_stopeed_total_time` to `None` Better explained with code :smiley_cat: :

```python
def start(self):
    """
    Starts the timer. If the timer is already running, does nothing.
    If the timer was previously stopped, continues with the previous elapsed time
    """
    if self._stopped_total_time is None:
        if self._start_time is None:
            self._start_time = get_ticks_ms()
    else:
        self._start_time = get_ticks_ms() - self._stopped_total_time
        self._stopped_total_time = None
```
Also I would either remove the `start()` call from reset, or name it restart.
And maybe rename `is_started` to `is_running` and implement `is_started` as `_start_time is not None or _stopped_total_time is not None`.

However, all of these are just ideas what I would find more intuitive...

WasabiFan:
I agree on the stop/reset semantics — I did it this way because it was backwards-compatible with what was released before. That being said, it was only released in the beta release, so if we're going to break it we should do it now. I'll look back at it tomorrow and maybe change it to be as you described.

@dwalton76 what do you think?

ev3dev2/stopwatch.py
```diff
@@ -9,76 +9,114 @@
...
+    """
     def __init__(self, desc=None):
+        """
+        Initializes the StopWatch but does not start it.
+
+        desc:
+            A string description to print when stringifying.
+        """
         self.desc = desc
         self._value = 0
```
mvolfik:
```suggestion
```
Remove `_value` as it's not used anymore

ev3dev2/stopwatch.py
```diff
@@ -9,76 +9,113 @@
...
+        Resets the timer, and starts it again.
+        """
+        self._start_time = None
+        self._stopped_total_time = None
+        self.start()
+    
+    @property
+    def is_started(self):
+        """
+        True if the StopWatch has been started but not stoped (i.e., it's currently running), false otherwise.
```
dwalton76, line 84: `s/stoped/stopped/`

dwalton76: `s/false/False/`

dwalton76 (approved):
The rest looks good to me

WasabiFan:
@dwalton76 What do you think about the questions raised by @mvolfik (about start/reset behavior)?

I think it would make sense to change the interface a bit but I'd like your opinion too.

dwalton76:
I think it would be better to have a 2nd call to start raise an Exception, just because it feels like an easy mistake for the user to make. If they want to reset the timer I think they should have to explicitly call `reset()`

WasabiFan:
I guess the question of restart behavior comes down to use cases: do we expect there's a common case where someone wants to start+stop a timer then start it again from where they left off? Is that useful? I can see cases where that could be used but it seems like the majority of use cases are where you want to start the timer then wait for some amount of time to elapse while performing another action.

ev3dev2/stopwatch.py
```diff
@@ -9,76 +9,113 @@
...
 
     def reset(self):
-        self.stop()
-        self._value = 0
+        """
+        Resets the timer, and starts it again.
+        """
+        self._start_time = None
+        self._stopped_total_time = None
+        self.start()
```
dwalton76, line 79: Having the reset call start the time feels odd. How about we drop the start() here but then add a restart method that calls reset() followed by start()?

ev3dev2/stopwatch.py
```diff
@@ -9,76 +9,113 @@
...
-            delta = current_time - self.start_time
-        else:
-            delta = current_time - self.prev_update_time
-
-        self._value += delta
-        self.prev_update_time = current_time
+        """
+        Starts the timer. If the timer is already running, resets it.
+        """
+        self._stopped_total_time = None
```
dwalton76, line 54: If the timer is already running and they call start() I think we should raise an exception. That feels like a error scenario we should make the user aware of.

dwalton76 (changes requested):

ev3dev2/stopwatch.py
WasabiFan: "restart" does make sense to me. I think that's a better option than what's currently here.

dwalton76 (approved):
Looks good

WasabiFan:
This is a slight breaking change; this prompts a reasonable discussion. One option would be to increment the minor version (`2.1.0`) and bend semver, which seems attractive but I'm a bit worried about setting that precedent.

The other option would be to increment it properly and call it `3.0`. I am generally OK with incrementing major versions relatively often as long as we have at least a few breaking changes included.

dwalton76:
Do we have any other breaking changes? If so 3.0 sounds good, if not then 2.1.0.

WasabiFan:
So far, we don't. I'll merge this as-is and then I can change the planned version number as new changes come in.

PR #669, opened 2019-08-27
Play song rests (fixes #595)

### LINKED ISSUE #595: Add rest notes to sound.Sound.play_song()
It would be nice if sound.Sound.play_song() had rest notes as a possibility.  My student was looking through all sorts of music he wanted to make his robot play and to quote him, "All music has rests, well at least all good music."

It can be solved by breaking apart songs on the rest notes and adding a delay, but that is rather messy.

mvolfik:
Also fix ignoring note capitalization and add ignoring value capit.

EDIT: Sorry, I didn't know that # links don't work in title...
fixes #595

### DIFF
--- a/ev3dev2/sound.py
+++ b/ev3dev2/sound.py
@@ -44,7 +44,7 @@ def _make_scales(notes):
     for note, freq in notes:
         freq = round(freq)
         for n in note.split('/'):
-            res[n] = freq
+            res[n.upper()] = freq
     return res
 
 
@@ -475,6 +475,7 @@ class Sound(object):
 
         try:
             for (note, value) in song:
+                value = value.lower()
                 if '/' in value:
                     base, factor = value.split('/')
                     duration_ms = meas_duration_ms * self._NOTE_VALUES[base] / float(factor)
@@ -493,7 +494,7 @@ class Sound(object):
                 if note == "R":
                     sleep(duration_ms / 1000 + delay)
                 else:
-                    freq = self._NOTE_FREQUENCIES.get(note.upper(), self._NOTE_FREQUENCIES[note])
+                    freq = self._NOTE_FREQUENCIES[note.upper()]
                     self.beep('-f %d -l %d -D %d' % (freq, duration_ms, delay_ms))
         except KeyError as e:
             raise ValueError('invalid note (%s)' % e)
--- a/ev3dev2/sound.py
+++ b/ev3dev2/sound.py
@@ -31,6 +31,7 @@ if sys.version_info < (3, 4):
 from ev3dev2 import is_micropython
 import os
 import re
+from time import sleep
 
 if not is_micropython():
     import shlex
@@ -406,6 +407,7 @@ class Sound(object):
         and duration.
 
         It supports symbolic notes (e.g. ``A4``, ``D#3``, ``Gb5``) and durations (e.g. ``q``, ``h``).
+        You can also specify rests by using ``R`` instead of note pitch.
 
         For an exhaustive list of accepted note symbols and values, have a look at the ``_NOTE_FREQUENCIES``
         and ``_NOTE_VALUES`` private dictionaries in the source code.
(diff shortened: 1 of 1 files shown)


ev3dev2/sound.py
```diff
@@ -471,39 +473,29 @@ def play_song(self, song, tempo=120, delay=0.05):
...
+                    duration_ms = meas_duration_ms * self._NOTE_VALUES[base] * 2 / 3
+                else:
+                    duration_ms = meas_duration_ms * self._NOTE_VALUES[value]
+
+                if note == "R":
+                    sleep(duration_ms / 1000 + delay)
+                else:
+                    freq = self._NOTE_FREQUENCIES[note.upper()]
+                    self.beep('-f %d -l %d -D %d' % (freq, duration_ms, delay_ms))
         except KeyError as e:
```
WasabiFan, line 499: this `try...except` around the whole thing is scary. Can we move it to the smallest relevant statement, probably the `freq` lookup?

ev3dev2/sound.py
```diff
@@ -43,7 +44,7 @@ def _make_scales(notes):
     for note, freq in notes:
         freq = round(freq)
         for n in note.split('/'):
-            res[n] = freq
```
WasabiFan, line 46: Are we confident there are no negative side-effects of this? Based on a cursory look it doesn't seem like there are any clashes but I am not 100% sure.

Also, it's probably weird for, say, the "b" (which I assume is supposed to look like a "flat" symbol) to be capitalized... but at the same time, it's easier to type all-caps. Not sure what the right balance is there.

WasabiFan (commented):
Minor changes requested.

It would be really nice to have unit tests for this (inject a fake beep method which just logs the args, play a song, then assert the beep args match the expected. Could generate that list from the previous version of this code and compare.) but only do that if you're interested in it -- we didn't have tests before and I don't expect you to have to add them.

ev3dev2/sound.py
mvolfik: No, there should definitely be no conflicts, actually if there were any, there's something wrong with the input data.

> Also, it's probably weird for, say, the "b" (which I assume is supposed to look like a "flat" symbol) to be capitalized... but at the same time, it's easier to type all-caps. Not sure what the right balance is there.

Yeah, it is the flat symbol, but what you say is one of the things I'm addressing in this PR - previously, the note inputted by user was looked up .upper()ed and unmodified (l484). This caused that A1, a1, A#1, a#1 and Ab1 were all ok, but AB1, ab1 and aB1 were not found. Now, that the data we're looking in is all uppercase, we can just convert input to uppercase too and find all of these.

WasabiFan: I was suggesting that perhaps we _didn't_ want those allowed. Isn't it weird to see a capital "B" used to mean "flat"?

mvolfik: Oh, ok, sorry, I didn't understand. Well yeah, it's weird, but I think it doesn't matter. However, if you don't want to allow "B" as flat, I guess removing the .upper in dict creation and .capitalize()ing the note in lookup would work. Should I do it like that, or revert it? (The only difference would be not "ab1")

WasabiFan: OK, no problem; I figure we can allow the variable caps because I don't see much of a danger in it anyway.

WasabiFan: OK, no problem; I figure we can allow the variable caps because I don't see much of a danger in it anyway.

ev3dev2/sound.py
WasabiFan: If you get a chance to change this I'll merge the PR.

dwalton76:
We haven't heard from mvolfik for a few months so I grabbed his/her patch, addressed the code review comments and posted a new PR.  Will give them credit or the patch in the commit message.
https://github.com/ev3dev/ev3dev-lang-python/pull/701

PR #692, opened 2019-11-24
Fix gyro sensor reset

dlech:
Gyro sensor reset was failing because we were writing 17 bytes of 0
to the gyro sensor instead of the byte 17.

Also change it to a bytes literal for max efficiency.

### DIFF
--- a/ev3dev2/sensor/lego.py
+++ b/ev3dev2/sensor/lego.py
@@ -631,7 +631,7 @@ class GyroSensor(Sensor):
               PiStorms, or with any sensor multiplexors.
         """
         # 17 comes from inspecting the .vix file of the Gyro sensor block in EV3-G
-        self._direct = self.set_attr_raw(self._direct, 'direct', bytes(17,))
+        self._direct = self.set_attr_raw(self._direct, 'direct', b'\x11')
 
     def wait_until_angle_changed_by(self, delta, direction_sensitive=False):
         """


WasabiFan (approved):
That's... absurd 😁

dlech:
I'm guessing from the trailing comma, what I originally meant to do was `bytes((17,))` which would have been correct. Apparently a trailing comma doesn't work inside of arguments like `x = 17,` does.

ghost:
What release will this change be in?

dlech:
As a workaround, you can create a subclass of `GyroSensor` with the fix and use it instead.

```python
class GyroSensor2(GyroSensor):
    def reset(self):
        self._direct = self.set_attr_raw(self._direct, 'direct', b'\x11')
```

WasabiFan:
Fix released in v2.0.0. See [our FAQ](https://python-ev3dev.readthedocs.io/en/ev3dev-stretch/faq.html) for upgrade commands. Let me know if there are any issues.

ghost:
Thanks for the solution.

Right now the kids read the gyro at the start of their mission runs when
the robot faces North on the FLL table.

They pass that to all the move functions they've written as an argument
called Gyro_Offset.

Then they have a function called Gyro_Angle that reads the gyro and
subtracts the offset to get a "table relative" angle, every time they need
an angle.

If we have time before our competition, we'll implement your suggestion.

Cheers!

Issue #620, opened 2019-04-18
MoveDifferential: use gyro support for better accuracy


dwalton76:
odometry support went in via #235 but the initial commit did not include gyro support

dwalton76:
@WasabiFan in a nutshell do you just use the gryo angle instead of calculating the angle based on the wheel movements?

dwalton76:
ok I watched enough builderdude35 videos on Youtube last night that I think I have my head wrapped around what to do with the gyro in terms of odometry.  I think my gryo is busted though, I do things like use MoveDifferential turn_to_angle to turn 90 degrees and the robot looks like it turned exactly 90 degrees but the gyro says it turned 50 degrees.

WasabiFan:
The way I would imagine implementing it is to throw the gyro and wheel odometry angle measurements into a Kalman filter (to smooth out the gyro data, primarily) then use the output measurement for calculations. That being said, the noise of the gyro probably isn't huge and both of our measurements are continuously integrated rather than individually sampled, so a Kalman filter might be overkill.

What was the approach you saw?

dwalton76:
For rotating in place it just used move steering to start rotating and then stopped when the gyro got to the desired angle. They had to underestimate though (aim for 80 degrees to end up turning the desired 90 degrees) so it wasnt crazy impressive. What we have right now (without a gyro) is actually pretty good. I can tell it to torn 90 and it will sometimes hit it dead on but others it will only be off by a degree or two.

dwalton76:
For driving forward in a straight line they used a PID controller to adjust motors to keep the gyro at the original angle. This all looked very cool to me and worked really well in the video. He would have the robot driving straight and then would nudge it with his hand and the robot would straighten itself out

dwalton76:
Having lots of gyro issues right now, filed https://github.com/ev3dev/ev3dev/issues/1262

dwalton76:
Also hitting https://github.com/ev3dev/ev3dev/issues/1269

dwalton76:
We reverted the commit so am re-opening this one

PR #697, opened 2019-12-18
Adds Gyro Support to MoveTank

nshenoy:
Adds the following gyro supported functions to MoveTank:
- follow_gyro_angle: angle following function that uses the same PID algorithm as the line follower
 to have the robot drive straight at a specific angle.
- pivot_gyro: pivots the robot to a specific angle
- calibrate_gyro: flips the mode of the sensor to zero-out the gyro to reduce chance of drift

These changes may partially address #620 as it follows the intent of that issue in that it uses the gyro for accurate driving.

### DIFF
--- a/ev3dev2/motor.py
+++ b/ev3dev2/motor.py
@@ -1866,20 +1866,20 @@ class MoveTank(MotorSet):
     # color sensor used by follow_line()
     @property
     def cs(self):
-        return self.__cs
+        return self._cs
     
     @cs.setter
     def cs(self, cs):
-        self.__cs = cs
+        self._cs = cs
 
     # gyro sensor used by follow_gyro_angle()
     @property
     def gyro(self):
-        return self.__gyro
+        return self._gyro
     
     @gyro.setter
     def gyro(self, gyro):
-        self.__gyro = gyro
+        self._gyro = gyro
 
     def _unpack_speeds_to_native_units(self, left_speed, right_speed):
         left_speed = self.left_motor._speed_native_units(left_speed, "left_speed")
@@ -2070,10 +2070,10 @@ class MoveTank(MotorSet):
                 tank.stop()
                 raise
         """
-        assert self.__cs, "ColorSensor must be defined"
+        assert self._cs, "ColorSensor must be defined"
 
         if target_light_intensity is None:
-            target_light_intensity = self.__cs.reflected_light_intensity
+            target_light_intensity = self._cs.reflected_light_intensity
 
         integral = 0.0
         last_error = 0.0
@@ -2083,7 +2083,7 @@ class MoveTank(MotorSet):
         MAX_SPEED = SpeedNativeUnits(self.max_speed)
 
         while follow_for(self, **kwargs):
-            reflected_light_intensity = self.__cs.reflected_light_intensity
+            reflected_light_intensity = self._cs.reflected_light_intensity
             error = target_light_intensity - reflected_light_intensity
             integral = integral + error
             derivative = error - last_error
@@ -2129,11 +2129,11 @@ class MoveTank(MotorSet):
 
         NOTE: This takes 1sec to run
         """
-        assert self.__gyro, "GyroSensor must be defined"
+        assert self._gyro, "GyroSensor must be defined"
 
         for x in range(2):
-            self.__gyro.mode = 'GYRO-RATE'
-            self.__gyro.mode = 'GYRO-ANG'
+            self._gyro.mode = 'GYRO-RATE'
+            self._gyro.mode = 'GYRO-ANG'
             time.sleep(0.5)
 
     def follow_gyro_angle(self,
@@ -2160,7 +2160,7 @@ class MoveTank(MotorSet):
 
         ``follow_for`` is called to determine if we should keep following the
             desired angle or stop.  This function will be passed ``self`` (the current
-            ``MoveWithGyroTank`` object). Current supported options are:
+            ``MoveTank`` object). Current supported options are:
             - ``follow_for_forever``
             - ``follow_for_ms``
 
@@ -2195,7 +2195,7 @@ class MoveTank(MotorSet):
                 tank.stop()
                 raise
         """
-        assert self.__gyro, "GyroSensor must be defined"
+        assert self._gyro, "GyroSensor must be defined"
 
         integral = 0.0
         last_error = 0.0
@@ -2203,8 +2203,10 @@ class MoveTank(MotorSet):
         speed_native_units = speed.to_native_units(self.left_motor)
         MAX_SPEED = SpeedNativeUnits(self.max_speed)
 
+        assert speed_native_units <= MAX_SPEED, "Speed exceeds the max speed of the motors"
+
         while follow_for(self, **kwargs):
-            current_angle = self.__gyro.angle
+            current_angle = self._gyro.angle
             error = current_angle - target_angle
             integral = integral + error
             derivative = error - last_error
@@ -2214,14 +2216,14 @@ class MoveTank(MotorSet):
             left_speed = SpeedNativeUnits(speed_native_units - turn_native_units)
             right_speed = SpeedNativeUnits(speed_native_units + turn_native_units)
 
-            if left_speed > MAX_SPEED:
+            if abs(left_speed) > MAX_SPEED:
                 log.info("%s: left_speed %s is greater than MAX_SPEED %s" %
                         (self, left_speed, MAX_SPEED))
                 self.stop()
                 raise FollowGyroAngleErrorTooFast(
                     "The robot is moving too fast to follow the angle")
 
-            if right_speed > MAX_SPEED:
+            if abs(right_speed) > MAX_SPEED:
                 log.info("%s: right_speed %s is greater than MAX_SPEED %s" %
                         (self, right_speed, MAX_SPEED))
                 self.stop()
@@ -2238,6 +2240,7 @@ class MoveTank(MotorSet):
     def turn_to_angle_gyro(self,
             speed,
             target_angle=0,
+            wiggle_room=2,
             sleep_time=0.01
         ):
         """
@@ -2246,6 +2249,8 @@ class MoveTank(MotorSet):
         ``speed`` is the desired speed of the midpoint of the robot
 
         ``target_angle`` is the target angle we want to pivot to
+
+        ``wiggle_room`` is the +/- angle threshold to control how accurate the turn should be
 
         ``sleep_time`` is how many seconds we sleep on each pass through
             the loop.  This is to give the robot a chance to react
@@ -2274,14 +2279,14 @@ class MoveTank(MotorSet):
                 target_angle(30)
             )
         """
-        assert self.__gyro, "GyroSensor must be defined"
+        assert self._gyro, "GyroSensor must be defined"
 
         speed_native_units = speed.to_native_units(self.left_motor)
         target_reached = False
 
         while not target_reached:
-            current_angle = self.__gyro.angle
-            if (abs(current_angle - target_angle) <= 2):
+            current_angle = self._gyro.angle
+            if abs(current_angle - target_angle) <= wiggle_room:
                 target_reached = True
                 self.stop()
             elif (current_angle > target_angle):
--- a/ev3dev2/motor.py
+++ b/ev3dev2/motor.py
@@ -1863,11 +1863,23 @@ class MoveTank(MotorSet):
         self.right_motor = self.motors[right_motor_port]
         self.max_speed = self.left_motor.max_speed
 
-        # color sensor used by follow_line()
-        self.cs = None
+    # color sensor used by follow_line()
+    @property
+    def cs(self):
+        return self.__cs
+    
+    @cs.setter
+    def cs(self, cs):
+        self.__cs = cs
 
-        # gyro sensor used by follow_gyro_angle()
-        self.gyro = None
+    # gyro sensor used by follow_gyro_angle()
+    @property
+    def gyro(self):
+        return self.__gyro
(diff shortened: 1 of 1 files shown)


ev3dev2/motor.py
```diff
@@ -1857,6 +1866,9 @@ def __init__(self, left_motor_port, right_motor_port, desc=None, motor_class=Lar
         # color sensor used by follow_line()
         self.cs = None
 
+        # gyro sensor used by follow_gyro_angle()
+        self.gyro = None
```
WasabiFan: How does the gyro get initialized?

ev3dev2/motor.py
```diff
@@ -2099,6 +2111,169 @@ def follow_line(self,
...
+            **kwargs
+        ):
+        """
+        PID gyro angle follower
+
+        ``kp``, ``ki``, and ``kd`` are the PID constants.
+
+        ``speed`` is the desired speed of the midpoint of the robot
+
+        ``target_angle`` is the angle we want drive
```
WasabiFan: missing `to`

ev3dev2/motor.py
```diff
@@ -2099,6 +2111,169 @@ def follow_line(self,
...
+
+        ``sleep_time`` is how many seconds we sleep on each pass through
+            the loop.  This is to give the robot a chance to react
+            to the new motor settings. This should be something small such
+            as 0.01 (10ms).
```
WasabiFan, line 2159: How does it perform with a 10ms loop time? Does it actually run at that frequency or is it limited by the execution time?

ev3dev2/motor.py
```diff
@@ -2099,6 +2111,169 @@ def follow_line(self,
...
+            - ``follow_for_forever``
+            - ``follow_for_ms``
+
+        ``**kwargs`` will be passed to the ``follow_for`` function
+
+        Example:
+
+        .. code:: python
+
+            from ev3dev2.motor import OUTPUT_A, OUTPUT_B, MoveTank, SpeedPercent, follow_for_ms
```
WasabiFan, line 2173: where is `follow_for_ms` defined?

ev3dev2/motor.py
```diff
@@ -2099,6 +2111,169 @@ def follow_line(self,
...
+            try:
+                # Follow the line for 4500ms
+                tank.calibrate_gyro()
+                tank.follow_gyro_angle(
+                    kp=11.3, ki=0.05, kd=3.2,
+                    speed=SpeedPercent(30),
+                    follow_for=follow_for_ms,
+                    ms=4500
+                )
+            except Exception:
```
WasabiFan: Probably good to catch the `FollowGyroAngleErrorTooFast` error here explicitly

ev3dev2/motor.py
```diff
@@ -2099,6 +2111,169 @@ def follow_line(self,
...
+        .. code:: python
+
+            from ev3dev2.motor import OUTPUT_A, OUTPUT_B, MoveTank, SpeedPercent, follow_for_ms
+            from ev3dev2.sensor.lego import GyroSensor
+            tank = MoveTank(OUTPUT_A, OUTPUT_B)
+            tank.gyro = GyroSensor()
+            try:
+                # Follow the line for 4500ms
+                tank.calibrate_gyro()
+                tank.follow_gyro_angle(
```
WasabiFan, line 2187: Where is the target angle?

What's the expected use case for this? Is the idea that you'd first retrieve the current angle then follow that? I'm in particular wondering how you know what target angle to use.

ev3dev2/motor.py
```diff
@@ -2099,6 +2111,169 @@ def follow_line(self,
...
+            error = current_angle - target_angle
+            integral = integral + error
+            derivative = error - last_error
+            last_error = error
+            turn_native_units = (kp * error) + (ki * integral) + (kd * derivative)
+
+            left_speed = SpeedNativeUnits(speed_native_units - turn_native_units)
+            right_speed = SpeedNativeUnits(speed_native_units + turn_native_units)
+
+            if left_speed > MAX_SPEED:
```
WasabiFan: Looking at it... what's the motivation for failing if the speed is too high as opposed to just capping it at that max speed?

I think we did this for the line following too and not I'm not sure why.

ev3dev2/motor.py
```diff
@@ -2099,6 +2111,169 @@ def follow_line(self,
...
+            speed,
+            target_angle=0,
+            sleep_time=0.01
+        ):
+        """
+        Pivot Turn
+
+        ``speed`` is the desired speed of the midpoint of the robot
+
+        ``target_angle`` is the angle we want drive
```
WasabiFan: this is the angle we want to _rotate to_, right?

ev3dev2/motor.py
```diff
@@ -2099,6 +2111,169 @@ def follow_line(self,
...
+            try:
+                # Reset gyro sensor to zero
+                tank.calibrate_gyro()
+
+                # Pivot 30 degrees
+                tank.pivot_gyro(
+                    speed=SpeedPercent(5),
+                    target_angle(30)
+                )
+            except Exception:
```
WasabiFan: same note as above regarding exception type

ev3dev2/motor.py
WasabiFan: I see from the examples it seems the intent is for the user to set it themselves. While this might be fine in general, I worry it will cause problems because a) we don't do this elsewhere AFAIK and b) it probably won't show up well in the documentation. I'm thinking it might be better to have a setter method or a property for it.

WasabiFan (commented):
Thanks for the contribution! Some various minor comments/questions.

ev3dev2/motor.py
nshenoy: You're correct, the intent for the user to set it. I followed the same pattern used for the color sensor `cs` attribute. I can make `gyro` an explicit property, but will that make it inconsistent though?

ev3dev2/motor.py
nshenoy: Excellent question. My tests have primarily been to use the default of 10ms. When using the gyro in EV3-G, we've had to introduce some amount of small sleep for the gyro to chillax. This allows users to tweak this if their gyro needs more rest between samples.

ev3dev2/motor.py
nshenoy: That's defined on [line 1826](https://github.com/ev3dev/ev3dev-lang-python/blame/3d22941975ca2b017812198120868f2cd4333e3c/ev3dev2/motor.py#L1826)  introduced by #630 .

ev3dev2/motor.py
nshenoy: Typo in my example code. Good catch!

ev3dev2/motor.py
nshenoy: It's a good way for the user to know that they're potentially shooting themselves in the foot. Going over a max could result in the sensors and algorithm to not correct themselves within the feedback loop's quantum.

ev3dev2/motor.py
nshenoy: I'll remove. This method doesn't throw an exception unless someone forgot to initialize the gyro. But this sample properly initializes it so there will be zero chance of the AssertionError being raised.

ev3dev2/motor.py
WasabiFan: Hmmm... That seems odd to me. Relatedly, I'm a bit bugged by the fact that we're putting so many different optional things in the `MoveTank` class. This isn't an issue specific to your PR; I'm just thinking through how we really want to do it.

How about, at least for now, we make it a property (getter/setter pair with `_gyro` private member internally) with a docstring so that it shows up in the documentation, and also change the color sensor to match? I don't think it will break anyone's existing code.

nshenoy: That's a good idea. I'll work on making them both properties in a non-breaking way.

dwalton76: Setting the self.cs this way is my fault. We should use a setter/getter for that one too

ev3dev2/motor.py
```diff
@@ -2099,6 +2111,178 @@ def follow_line(self,
...
+                    "The robot is moving too fast to follow the angle")
+
+            if sleep_time:
+                time.sleep(sleep_time)
+
+            self.on(left_speed, right_speed)
+
+        self.stop()
+
+    def pivot_gyro(self,
```
dwalton76: How about `turn_to_angle_gryo`? That would be more consistent with `MoveDifferential.turn_to_angle()`

ev3dev2/motor.py
```diff
@@ -2099,6 +2111,178 @@ def follow_line(self,
...
+            )
+        """
+        assert self.gyro, "GyroSensor must be defined"
+
+        speed_native_units = speed.to_native_units(self.left_motor)
+        target_reached = False
+
+        while not target_reached:
+            current_angle = self.gyro.angle
+            if (current_angle >= target_angle-2 and current_angle <= target_angle+2):
```
dwalton76: could do `if abs(current_angle - target_angle) <= 2:` to get this down to a single check

dwalton76 (changes requested):

ev3dev2/motor.py
nshenoy: Good catch! I'll fix this.

ev3dev2/motor.py
nshenoy: Makes sense.

ev3dev2/motor.py
```diff
@@ -1854,8 +1863,23 @@ def __init__(self, left_motor_port, right_motor_port, desc=None, motor_class=Lar
         self.right_motor = self.motors[right_motor_port]
         self.max_speed = self.left_motor.max_speed
 
-        # color sensor used by follow_line()
-        self.cs = None
+    # color sensor used by follow_line()
+    @property
+    def cs(self):
+        return self.__cs
```
dwalton76: For cs and for gyro is the double leading underscore intentional or should this be the more standard single leading underscore?  I didn't realize double leading underscore had a special meaning but then read this
https://dbader.org/blog/meaning-of-underscores-in-python

I'm not sure we need the name mangling for these...I don't think we use this for any other private variables.

ev3dev2/motor.py
```diff
@@ -2099,6 +2123,178 @@ def follow_line(self,
...
+        ``target_angle`` is the angle we want to maintain
+
+        ``sleep_time`` is how many seconds we sleep on each pass through
+            the loop.  This is to give the robot a chance to react
+            to the new motor settings. This should be something small such
+            as 0.01 (10ms).
+
+        ``follow_for`` is called to determine if we should keep following the
+            desired angle or stop.  This function will be passed ``self`` (the current
+            ``MoveWithGyroTank`` object). Current supported options are:
```
dwalton76: `MoveWithGyroTank` -> `MoveTank` (there is no `MoveWithGyroTank` class)

ev3dev2/motor.py
```diff
@@ -2099,6 +2123,178 @@ def follow_line(self,
...
+            error = current_angle - target_angle
+            integral = integral + error
+            derivative = error - last_error
+            last_error = error
+            turn_native_units = (kp * error) + (ki * integral) + (kd * derivative)
+
+            left_speed = SpeedNativeUnits(speed_native_units - turn_native_units)
+            right_speed = SpeedNativeUnits(speed_native_units + turn_native_units)
+
+            if left_speed > MAX_SPEED:
```
dwalton76: I think this needs to be `if abs(left_speed)...` in case the motor is being asked to move backwards at an impossible speed.  That seems very unlikely but you never know.

Same for the right_speed check on 2224

ev3dev2/motor.py
```diff
@@ -2099,6 +2123,178 @@ def follow_line(self,
...
+            tank.calibrate_gyro()
+
+            # Pivot 30 degrees
+            tank.turn_to_angle_gyro(
+                speed=SpeedPercent(5),
+                target_angle(30)
+            )
+        """
+        assert self.__gyro, "GyroSensor must be defined"
+
```
dwalton76, line 2283: We should assert that ``speed`` is <= the max speed of the motors

ev3dev2/motor.py
```diff
@@ -2099,6 +2123,178 @@ def follow_line(self,
...
+        speed_native_units = speed.to_native_units(self.left_motor)
+        target_reached = False
+
+        while not target_reached:
+            current_angle = self.__gyro.angle
+            if (abs(current_angle - target_angle) <= 2):
```
dwalton76: Would pass the 2 into this method as an arg `wiggle_room=2` and then compare against `wiggle_room` here...that way the user can control how accurate they want it to be.

dwalton76 (changes requested):

ev3dev2/motor.py
nshenoy: I'd argue that the double underscore is the more pythonic way of keeping the private variable truly private and having accessor methods. But I agree that if this doesn't fit the conventions of the rest of the project, then we should make it consistent with a single underscore. Let me know what you prefer.

ev3dev2/motor.py
nshenoy: Good idea! I'll add that.

ev3dev2/motor.py
WasabiFan: Personally I'd say use a single underscore to be consistent within this codebase. I didn't know about the double-underscore's behavior -- Python always has some new implicit behavior to deal with! _shakes fist_

dwalton76: I like sticking with single underscore for consistency

dwalton76 (approved):

dwalton76:
Looks good to me...merging. 
Thanks for the code :)

PR #701, opened 2019-12-29
Add rest notes to sound.Sound.play_song()

dwalton76:
mat.volfik@gmail.com is the original author of this patch
resolves issue #595


WasabiFan:
If you're adopting the patch can you set the author on the commits to be @mvolfik so it's kept track of within Git? Force-pushing to this branch is fine.

dwalton76:
got the commit log edited to list mvolfik...merging this one

PR #705, opened 2020-01-04
MoveDifferential: use gyro support for better accuracy (reverted this commit)

### LINKED ISSUE #620: MoveDifferential: use gyro support for better accuracy

dwalton76:
- Resolves #620 
- unify (as much as possible) the *turn* API for MoveTank and MoveDifferential.  MoveTank recently got gryo support so we needed to cleanup the API between it and MoveDifferential.
- removed all trailing whitespaces in various files

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,9 +1,9 @@
 python-ev3dev2 (2.1.0) UNRELEASED; urgency=medium
 
   [Daniel Walton]
+  * RPyC update docs and make it easier to use
   * MoveDifferential use gyro for better accuracy. Make MoveTank and
     MoveDifferential "turn" APIs more consistent.
-  * RPyC update docs and make it easier to use
 
   [Matěj Volf]
   * LED animation fix duration None
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,5 +1,8 @@
 python-ev3dev2 (2.1.0) UNRELEASED; urgency=medium
 
+  [Daniel Walton]
+  * MoveDifferential._turn() use gyro for better accuracy
+
   [Matěj Volf]
   * LED animation fix duration None
   * Add rest notes to sound.Sound.play_song()
--- a/debian/changelog
+++ b/debian/changelog
@@ -4,7 +4,7 @@ python-ev3dev2 (2.1.0) UNRELEASED; urgency=medium
   * Add "value_secs" and "is_elapsed_*" methods to StopWatch
   * Rename "value_hms" to "hms_str". New "value_hms" returns original tuple.
   * Fix bug in Button.process() on Micropython
-  
+
   [Nithin Shenoy]
   * Add Gyro-based driving support to MoveTank
 
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,7 +1,8 @@
 python-ev3dev2 (2.1.0) UNRELEASED; urgency=medium
 
   [Daniel Walton]
-  * MoveDifferential._turn() use gyro for better accuracy
+  * MoveDifferential use gyro for better accuracy. Make MoveTank and
+    MoveDifferential "turn" APIs more consistent.
 
   [Matěj Volf]
   * LED animation fix duration None
--- a/ev3dev2/motor.py
+++ b/ev3dev2/motor.py
@@ -1867,7 +1867,7 @@ class MoveTank(MotorSet):
     @property
     def cs(self):
         return self._cs
-    
+
     @cs.setter
     def cs(self, cs):
         self._cs = cs
@@ -1876,7 +1876,7 @@ class MoveTank(MotorSet):
     @property
     def gyro(self):
         return self._gyro
-    
+
     @gyro.setter
     def gyro(self, gyro):
         self._gyro = gyro
@@ -2260,7 +2260,7 @@ class MoveTank(MotorSet):
         Example:
 
         .. code:: python
-        
+
             from ev3dev2.motor import OUTPUT_A, OUTPUT_B, MoveTank, SpeedPercent
             from ev3dev2.sensor.lego import GyroSensor
 
@@ -2482,9 +2482,8 @@ class MoveDifferential(MoveTank):
         self.x_pos_mm = 0.0  # robot X position in mm
         self.y_pos_mm = 0.0  # robot Y position in mm
         self.odometry_thread_run = False
-        self.odometry_thread_id = None
+        self.odometry_thread_running = False
         self.theta = 0.0
-        self.gyro = None
 
     def on_for_distance(self, speed, distance_mm, brake=True, block=True):
         """
@@ -2563,35 +2562,38 @@ class MoveDifferential(MoveTank):
         """
         self._on_arc(speed, radius_mm, distance_mm, brake, block, False)
 
-    def _turn(self, speed, degrees, brake=True, block=True):
+    def _turn(self, speed, degrees, brake=True, block=True, wiggle_room=2):
         """
         Rotate in place `degrees`. Both wheels must turn at the same speed for us
         to rotate in place.
         """
-        use_gyro = bool(block and brake and self.gyro)
 
-        if degrees > 0:
-            target_theta = self.theta - degrees
-        else:
-            target_theta = self.theta + degrees
+        def final_angle(init_angle, degrees):
+            result = init_angle - degrees
 
-        while target_theta < -360:
-            target_theta += 360
+            while result <= -360:
+                result += 360
 
-        while target_theta > 360:
-            target_theta -= 360
+            while result >= 360:
+                result -= 360
+
+            if result < 0:
+                result += 360
+
+            return result
+
+        # use the gyro to check that we turned the correct amount?
+        use_gyro = bool(block and brake and self._gyro)
 
         if use_gyro:
-            # If we are at 90 degrees and are rotating 5 degrees we will
-            # rotate clockwise and should end up at 85 degrees
-            gyro_init_angle = self.gyro.angle
-            gyro_target_angle = gyro_init_angle + degrees
-            log.info("%s: _turn() %d degrees from %s to %s, gyro from %s to %s, gyro offset %s" %
-                (self, degrees, self.theta, target_theta,
-                gyro_init_angle, gyro_target_angle, self.gyro_angle_offset))
+            angle_init_degrees = math.degrees(self.theta)
         else:
-            log.info("%s: _turn() %d degrees from %s to %s" %
-                (self, degrees, self.theta, target_theta))
+            angle_init_degrees = self._gyro.circle_angle()
+
+        angle_target_degrees = final_angle(angle_init_degrees, degrees)
+
+        log.info("%s: _turn() %d degrees from %s to %s" %
+            (self, degrees, angle_init_degrees, angle_target_degrees))
 
         # The distance each wheel needs to travel
         distance_mm = (abs(degrees) / 360) * self.circumference_mm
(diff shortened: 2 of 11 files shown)


ev3dev2/motor.py
```diff
@@ -1925,7 +1925,7 @@ def on_for_degrees(self, left_speed, right_speed, degrees, brake=True, block=Tru
         self.right_motor._set_rel_position_degrees_and_speed_sp(right_degrees, right_speed_native_units)
         self.right_motor._set_brake(brake)
 
-        log.debug("{}: on_for_degrees {}".format(self, degrees))
```
WasabiFan, line 1928: Is this intentional? Is it for perf?

debian/changelog
```diff
@@ -1,5 +1,8 @@
 python-ev3dev2 (2.1.0) UNRELEASED; urgency=medium
 
+  [Daniel Walton]
+  * MoveDifferential._turn() use gyro for better accuracy
```
WasabiFan: Can this include a description of the external effects the change has? `_turn` is a new internal API.

ev3dev2/motor.py
```diff
@@ -2237,30 +2224,103 @@ def follow_gyro_angle(self,
...
 
         ``sleep_time`` is how many seconds we sleep on each pass through
             the loop.  This is to give the robot a chance to react
             to the new motor settings. This should be something small such
             as 0.01 (10ms).
+        """
+
+        # MoveTank does not have information on wheel size and distance (that is
+        # MoveDifferential) so we must use a GyroSensor to control how far we rotate.
+        assert self._gyro, "GyroSensor must be defined"
```
WasabiFan: Two things:
- I don't think an `AssertionError` is the preferred experience for trying to use this without having a gyro configured. Could we throw a custom exception type for not having configured a gyro? And make sure it's consistent if we do this anywhere else.
- The message should say how to fix it, otherwise this will be confusing.

ev3dev2/motor.py
```diff
@@ -2237,30 +2224,103 @@ def follow_gyro_angle(self,
...
+
+        if not target_angle:
```
WasabiFan: This is somewhat inconsistent (`0.0` will no-op but `0.01` will run the below logic). I'd say we can remove this.

ev3dev2/motor.py
```diff
@@ -2237,30 +2224,103 @@ def follow_gyro_angle(self,
...
+            return
+
+        speed_native_units = speed.to_native_units(self.left_motor)
+        init_angle = self._gyro.angle
+        first = True
+
+        while True:
+            current_angle = self._gyro.angle
+            degrees_moved = abs(current_angle - init_angle)
+            delta = abs(abs(target_angle) - degrees_moved)
```
WasabiFan: What if we requested to rotate `+15` but have so far rotated `-15`? Won't this count it as a delta of `0`?

ev3dev2/motor.py
```diff
@@ -2237,30 +2224,103 @@ def follow_gyro_angle(self,
...
+
+            if delta <= wiggle_room:
+                self.stop(brake=brake)
+                break
+
+            # our goal is to rotate target_angle clockwise
+            if target_angle > 0:
+
+                # we have not gone far enough, rotate clockwise
+                if first or degrees_moved < target_angle:
```
WasabiFan: Why is this "first" needed? And why is this larger "if" two separate branches with two sub-branches? Can't it be simplified to _If we're too far to the left, turn right; otherwise, turn left_ or similar?

ev3dev2/motor.py
```diff
@@ -2237,30 +2224,103 @@ def follow_gyro_angle(self,
...
+            self.on(left_speed, right_speed)
+
+            if sleep_time:
+                time.sleep(sleep_time)
+
+    def turn_right(self, speed, degrees, brake=True, wiggle_room=2, sleep_time=0.01):
+        """
+        Rotate clockwise `degrees` in place
+        """
+        self._turn(speed, abs(degrees), brake, wiggle_room, sleep_time)
```
WasabiFan: I don't think the `abs` is needed... it would be weird if both `-15` and `15` did the same thing. I think it's fine to just invert it in the "left" case.

ev3dev2/motor.py
```diff
@@ -2271,35 +2331,19 @@ def turn_to_angle_gyro(self,
...
-                target_angle(30)
+                target_angle=30
             )
         """
-        assert self._gyro, "GyroSensor must be defined"
-
-        speed_native_units = speed.to_native_units(self.left_motor)
-        target_reached = False
+        if target_degrees > 0:
+            self.turn_right(speed, target_degrees, brake, wiggle_room, sleep_time)
```
WasabiFan: This function just does what `_turn` does. I don't think it needs to call through the "left" and "right" versions which in turn call `_turn`. In fact, it might make most sense to rename `_turn` back to `turn_degrees` or `turn_degrees_with_gyro`.

Also, this is a breaking change (the function name). Ditto to the calibration function. So this might warrant a `3.0.0` release.

ev3dev2/motor.py
```diff
@@ -2482,21 +2526,21 @@ def __init__(self, left_motor_port, right_motor_port,
...
-        log.debug("%s: on_for_rotations distance_mm %s, rotations %s, speed %s" % (self, distance_mm, rotations, speed))
+        log.debug("%s: on_for_rotations distance_mm %s, rotations %s, speed %s" %
+            (self, distance_mm, rotations, speed))
 
         MoveTank.on_for_rotations(self, speed, speed, rotations, brake, block)
 
     def _on_arc(self, speed, radius_mm, distance_mm, brake, block, arc_right):
         """
-        Drive in a circle with 'radius' for 'distance'
+        Drive in a circle with `radius` for `distance`
```
WasabiFan: This is rST, so I think all of these need to be double ticks.

ev3dev2/motor.py
```diff
@@ -2551,51 +2595,144 @@ def _on_arc(self, speed, radius_mm, distance_mm, brake, block, arc_right):
...
+            while result >= 360:
+                result -= 360
+
+            if result < 0:
+                result += 360
+
+            return result
+
+        # use the gyro to check that we turned the correct amount?
+        use_gyro = bool(block and brake and self._gyro)
```
WasabiFan: This behavior should be documented in the docstring.

WasabiFan: Also, I'm worried that this function implicitly changes its behavior if you define a gyro. What if you only want the gyro to be used for other functions?

Also, consider this:
- An FLL team designs their robot with a gyro. They do it to use one of the gyro-specific functions that helps them drive straighter, for example.
- They decide they no longer want to have the gyro in use for driving straighter and delete that code.
- Now, this function silently changes the way it behaves and likely no longer turns the same way.

One option would be to make gyro usage intentionally opt-in on each individual function. Another would be to have separate versions of any functions that use the gyro. There are also other possibilities; I'm not sure which is best.

ev3dev2/motor.py
```diff
@@ -2551,51 +2595,144 @@ def _on_arc(self, speed, radius_mm, distance_mm, brake, block, arc_right):
...
-            rotations = distance_mm / self.wheel.circumference_mm
             MoveTank.on_for_rotations(self, speed * -1, speed, rotations, brake, block)
 
-    def turn_right(self, speed, degrees, brake=True, block=True):
+        if use_gyro:
+            angle_current_degrees = self._gyro.circle_angle()
+
+            # This can happen if we are aiming for 2 degrees and overrotate to 358 degrees
+            # We need to rotate counter-clockwise
+            if 90 >= angle_target_degrees >= 0 and 270 <= angle_current_degrees <= 360:
```
WasabiFan, line 2621: What about the other direction? Aiming for 358 and turn past to 2?

ev3dev2/motor.py
```diff
@@ -2551,51 +2595,144 @@ def _on_arc(self, speed, radius_mm, distance_mm, brake, block, arc_right):
...
+            if abs(degrees_error) > wiggle_room:
+                self._turn(speed, degrees_error, brake, block)
+
+    def turn_right(self, speed, degrees, brake=True, block=True, wiggle_room=2):
         """
-        Rotate clockwise 'degrees' in place
+        Rotate clockwise `degrees` in place
         """
-        self._turn(speed, abs(degrees), brake, block)
+        self._turn(speed, abs(degrees), brake, block, wiggle_room)
```
WasabiFan: I think `wiggle_room` should be named something more conventional to be clearer. Perhaps `error_margin`, `deadzone`, `epsilon`, `completion_threshold`, `precision`, etc.

ev3dev2/motor.py
```diff
@@ -2551,51 +2595,144 @@ def _on_arc(self, speed, radius_mm, distance_mm, brake, block, arc_right):
...
+
+        if angle_delta > 180:
+            angle_delta = 360 - angle_delta
+            turn_right = not turn_right
+
+        log.debug("%s: turn_to_angle %s, current angle %s, delta %s, turn_right %s" %
+            (self, angle_target_degrees, angle_current_degrees, angle_delta, turn_right))
+        self.odometry_coordinates_log()
+
+        if turn_right:
```
WasabiFan, line 2688: Similar to above, I think we should avoid the special-casing, both for readability and potential minor performance gains.

ev3dev2/sensor/lego.py
```diff
@@ -622,6 +626,15 @@ def tilt_rate(self):
         self._ensure_mode(self.MODE_TILT_RATE)
         return self.value(0)
 
+    def calibrate(self):
```
WasabiFan, line 629: :+1:

ev3dev2/sensor/lego.py
```diff
@@ -660,6 +674,61 @@ def wait_until_angle_changed_by(self, delta, direction_sensitive=False):
...
+        The angles on a circle have the opposite behavior though, they start
+        at 0 and increase as you move counter-clockwise around the circle.
+
+        Convert the gyro angle to the angle on a circle. We consider the initial
+        position of the gyro to be at 90 degrees on the cirlce.
+        """
+        current_angle = self.angle
+        delta = abs(current_angle - self._init_angle) % 360
+
+        if delta == 0:
```
WasabiFan, line 692: This seems over-complicated. Shouldn't the transformation from gyro angle to appropriate angle always be `some_constant_offset - current_angle`?

WasabiFan (commented):
Sorry for taking so long to review this!

I like consolidating and making the various `Move*` classes consistent. My main worry is just making sure that we don't introduce confusion about when and how the gyro is used.

debian/changelog
dwalton76: Will drop the `_turn` part of the description, that doesn't need to be in there.  Will add some more text about the other changes.

ev3dev2/motor.py
dwalton76: perf...and it was a little chatty

ev3dev2/motor.py
dwalton76: If we are asked to move 0 degrees that should be a no-op though?

ev3dev2/motor.py
dwalton76: I don't think that is possible, if `target_angle` is positive we will rotate clockwise which will cause the gryo angle to increase.

ev3dev2/motor.py
dwalton76: The `first` is needed so that on our first pass through this loop we start the motors in the correct direction...although now that I type that and re-read this code it doesn't seem like it would be needed.  Will test with this removed.

dwalton76: `If we're too far to the left, turn right; otherwise, turn left or similar` it depends on if our target_angle is positive or negative...take a look at the speeds that are being set in the various blocks of this if/else chain.

ev3dev2/motor.py
dwalton76: It is just a safety check so that if the user does `turn_right(degrees=-15)` we still rotate in the correct direction.  We have these sorts of safety checks in other places so seems good to have them here for consistency.

ev3dev2/motor.py
dwalton76: `turn_to_angle_gyro` and `calibrate_gyro` have only been in the code base for a couple of weeks so they haven't gone out in a release...so these shouldn't be a breaking change.

yep I can have this call `_turn` directly...no need to call turn_left and turn_right here really.

How about `_turn` to `_turn_via_gyro`?

ev3dev2/motor.py
dwalton76: ah, I was trying to make them all consistent (we are all over the place with this in our docstrings).  Will change them all to double ticks.

ev3dev2/motor.py
dwalton76: good point, I'll add a `use_gyro` arg to this method and then we can do
```
use_gryo = use_gryo and block and brake
```

ev3dev2/motor.py
dwalton76: good catch...will test this (am pretty sure it will break)

ev3dev2/motor.py
dwalton76: `error_margin` sounds good to me

ev3dev2/sensor/lego.py
dwalton76: I thought so too, figured it would be relatively simple.  I spent a few hours coming up with this function though...maybe I am missing something.

ev3dev2/motor.py
dwalton76: will just delete this so we aren't leaving commented out code in the repo

ev3dev2/motor.py
dwalton76: Will fix here but am also going to file a bug for us to clean this up elsewhere...we have a lot of places we are doing asserts where it would be more clear if we were raising ValueError.

dwalton76: Filed #707

ev3dev2/motor.py
dwalton76: Actually yeah we can just delete this method and rename `_turn` to `turn_degrees`.

ev3dev2/motor.py
dwalton76: This is another one where we are all over the place with our use of single quotes, double quotes, single backticks and double backticks. Filed #708

dwalton76:
Still some more work to go on this, I should be able to get to the other comments this weekend.

ev3dev2/motor.py
dwalton76: confirmed that `first` is not needed

dwalton76: `If we're too far to the left, turn right; otherwise, turn left or similar` you are correct this code path can be simplified...have chopped a good bit here

WasabiFan:
I had been waiting to review this until I had more time... I still have some concerns which I'll comment on when I can work through it more thoroughly.

dwalton76:
My bad...I had resolved all of the comments so thought we were set.  We can revert the commit or I can tackle the new comments via a new PR.

WasabiFan:
In the same vein as we discussed in #707 and shown in #715, this PR makes breaking changes, and it will cause confusion. I think this should be reverted so we can fix the breaking changes.

dwalton76:
@WasabiFan ack let me revert this and open a new PR.  My bad for jumping the gun on the commit.

dwalton76:
Reverted this commit and opened https://github.com/ev3dev/ev3dev-lang-python/pull/718

Issue #707, opened 2020-01-11
replace asserts with ValueError or TypeError


dwalton76:
Although not common asserts can be disabled (`python -O run.py`). We have a lot of places in the code where we are using asserts but should really be raising `ValueError`.

PR #709, opened 2020-01-11
replace asserts with ValueError or TypeError

dwalton76:


dwalton76:
For #707

dlech:
Although this is "more correct", it is a breaking change for anyone catching the existing exceptions - something to consider.

WasabiFan:
On one hand, these are errors that _should_ be considered programmer error (i.e., a caller's bug) rather than a signal, and thus people _shouldn't_ be catching them. On the other hand, it _is_ a breaking change, and we've been having this discussion already with other slight breaking changes which we have not yet released.

I am leaning toward including this, calling the next release v3.0, and making a breaking changes page in the docs for v3. But I'm not yet entirely sure that's what I want to do. Certainly, it breaks the nice "v1.0 is Jessie, v2.0 is Stretch" rule we have so far, but I'm not sure that really matters.

dlech:
There is ev3dev-buster if you want to go crazy with breaking changes :wink:

I would be very cautious about releasing any breaking changes to ev3dev-stretch, no matter how small. My experience has always been that it breaks more than you would think. And there are lots of people using ev3dev-lang-python.

dwalton76:
I would be game for creating an ev3dev-buster branch and putting this there instead.

PR #710, opened 2020-01-11
use double backticks consistently in docstrings

dwalton76:


ev3dev2/motor.py
```diff
@@ -1007,7 +1007,7 @@ def on_to_position(self, speed, position, brake=True, block=True):
         """
         Rotate the motor at ``speed`` to ``position``
 
-        ``speed`` can be a percentage or a :class:`ev3dev2.motor.SpeedValue`
+        ``speed`` can be a percentage or a :class:``ev3dev2.motor.SpeedValue``
```
WasabiFan: Are you sure this is right? I think these might be functional links.

ev3dev2/sensor/lego.py
```diff
@@ -727,48 +727,48 @@ class InfraredSensor(Sensor, ButtonBase):
...
+    #: Handler for bottom-left button events on channel 4. See :meth:``InfraredSensor.process``.
     on_channel4_bottom_left = None
-    #: Handler for top-right button events on channel 4. See :meth:`InfraredSensor.process`.
+    #: Handler for top-right button events on channel 4. See :meth:``InfraredSensor.process``.
     on_channel4_top_right = None
-    #: Handler for bottom-right button events on channel 4. See :meth:`InfraredSensor.process`.
+    #: Handler for bottom-right button events on channel 4. See :meth:``InfraredSensor.process``.
     on_channel4_bottom_right = None
-    #: Handler for beacon button events on channel 4. See :meth:`InfraredSensor.process`.
+    #: Handler for beacon button events on channel 4. See :meth:``InfraredSensor.process``.
```
WasabiFan: Same as other comment (might be additional instances); I think it might be required as-is

WasabiFan (commented):
Looks good in general, but I think some of these will break existing links. Not 100% sure though.

ev3dev2/sensor/lego.py
dwalton76: I’m not sure on these two will revert these parts to play it safe and then merge.

PR #718, opened 2020-01-17
MoveDifferential: use gyro for better accuracy

dwalton76:


ev3dev2/motor.py
```diff
@@ -2070,7 +2054,8 @@ def follow_line(self,
                 tank.stop()
                 raise
         """
-        assert self._cs, "ColorSensor must be defined"
+        if not self._cs:
+            raise DeviceNotDefined("The 'cs' variable must be defined with a ColorSensor. Example: tank.cs = ColorSensor()")
```
dwalton76: Changing from an `assert` to raising `DeviceNotDefined` is a breaking change but the odds of someone getting burned by this are pretty low.  They would have to be using `MoveTank.cs` without one defined but catching the assert to ignore the error.

ev3dev2/motor.py
```diff
@@ -2123,19 +2108,6 @@ def follow_line(self,
...
-
-        NOTE: This takes 1sec to run
-        """
-        assert self._gyro, "GyroSensor must be defined"
-
-        for x in range(2):
-            self._gyro.mode = 'GYRO-RATE'
-            self._gyro.mode = 'GYRO-ANG'
-            time.sleep(0.5)
-
```
dwalton76, line 2133: This method was not present in 2.0.0 so we are ok here
https://github.com/ev3dev/ev3dev-lang-python/blob/ev3dev-stretch/2.0.0/ev3dev2/motor.py

ev3dev2/motor.py
```diff
@@ -2195,7 +2167,8 @@ def follow_gyro_angle(self,
                 tank.stop()
                 raise
         """
-        assert self._gyro, "GyroSensor must be defined"
+        if not self._gyro:
+            raise DeviceNotDefined("The 'gyro' variable must be defined with a GyroSensor. Example: tank.gyro = GyroSensor()")
```
dwalton76: `follow_gyro_angle` was not there in 2.0.0 so we are ok here

ev3dev2/motor.py
```diff
@@ -2237,30 +2210,36 @@ def follow_gyro_angle(self,
 
         self.stop()
 
-    def turn_to_angle_gyro(self,
+    def turn_degrees(
```
dwalton76: `turn_to_angle_gyro` was not there in 2.0.0 so we are ok here

ev3dev2/motor.py
```diff
@@ -2561,41 +2563,134 @@ def on_arc_left(self, speed, radius_mm, distance_mm, brake=True, block=True):
         """
         self._on_arc(speed, radius_mm, distance_mm, brake, block, False)
 
-    def _turn(self, speed, degrees, brake=True, block=True):
+    def turn_degrees(self, speed, degrees, brake=True, block=True, error_margin=2, use_gyro=False):
```
dwalton76, line 2557: `_turn` was there in 2.0.0 but this is private so I don't think this would count as a breaking change since it is not part of the public API

ev3dev2/motor.py
```diff
@@ -2561,41 +2563,134 @@ def on_arc_left(self, speed, radius_mm, distance_mm, brake=True, block=True):
...
+    def turn_right(self, speed, degrees, brake=True, block=True, error_margin=2, use_gyro=False):
         """
-        Rotate clockwise 'degrees' in place
+        Rotate clockwise ``degrees`` in place
         """
-        self._turn(speed, abs(degrees), brake, block)
+        self.turn_degrees(speed, abs(degrees), brake, block, error_margin, use_gyro)
 
-    def turn_left(self, speed, degrees, brake=True, block=True):
+    def turn_left(self, speed, degrees, brake=True, block=True, error_margin=2, use_gyro=False):
```
dwalton76, line 2642: `turn_right`, `turn_left` and `turn_to_angle` had args added at the end so we should be ok here

ev3dev2/motor.py
```diff
@@ -2670,66 +2761,26 @@ def _odometry_monitor():
                 if sleep_time:
                     time.sleep(sleep_time)
 
-            self.odometry_thread_id = None
+        _thread.start_new_thread(_odometry_monitor, ())
 
-        self.odometry_thread_run = True
-        self.odometry_thread_id = _thread.start_new_thread(_odometry_monitor, ())
```
dwalton76, line 2662: The removal of `odometry_thread_id` could be breaking.  I modified this because `_thread.start_new_thread` in micropython returns None so the behavior was inconsistent.  I could put this back the way it was but just not use `odometry_thread_id` anywhere.

ev3dev2/motor.py
```diff
@@ -2670,66 +2761,26 @@ def _odometry_monitor():
...
-
-        self.odometry_coordinates_log()
-
     def on_to_coordinates(self, speed, x_target_mm, y_target_mm, brake=True, block=True):
         """
         Drive to (``x_target_mm``, ``y_target_mm``) coordinates at ``speed``
         """
-        assert self.odometry_thread_id, "odometry_start() must be called to track robot coordinates"
+        if not self.odometry_thread_run:
+            raise ThreadNotRunning("odometry_start() must be called to track robot coordinates")
```
dwalton76, line 2772: This could be breaking but seems unlikely someone would be catching this assert when trying to use odometry.

dwalton76:
@WasabiFan any comments on this one

ev3dev2/motor.py
```diff
@@ -2551,41 +2552,133 @@ def on_arc_left(self, speed, radius_mm, distance_mm, brake=True, block=True):
...
             MoveTank.on_for_rotations(self, speed, speed * -1, rotations, brake, block)
 
         # If degrees is negative rotate counter-clockwise
         else:
-            rotations = distance_mm / self.wheel.circumference_mm
             MoveTank.on_for_rotations(self, speed * -1, speed, rotations, brake, block)
 
-    def turn_right(self, speed, degrees, brake=True, block=True):
+        if use_gyro:
+            angle_current_degrees = self._gyro.circle_angle()
```
WasabiFan, line 2610: Throw error if no gyro?

ev3dev2/sensor/lego.py
```diff
@@ -661,6 +675,61 @@ def wait_until_angle_changed_by(self, delta, direction_sensitive=False):
...
+    def circle_angle(self):
+        """
+        As the gryo rotates clockwise the angle increases, it will increase
+        by 360 for each full rotation. As the gyro rotates counter-clockwise
+        the gyro angle will decrease.
+
+        The angles on a circle have the opposite behavior though, they start
+        at 0 and increase as you move counter-clockwise around the circle.
+
+        Convert the gyro angle to the angle on a circle. We consider the initial
```
WasabiFan, line 687: I don't understand this... what circle is involved here?

WasabiFan (approved):
Sorry! I did a bad job of keeping up on GitHub threads. Minor questions (don't need to be changed necessarily) but otherwise looks good!

ev3dev2/sensor/lego.py
dwalton76: A circle that represents the robot’s heading

Issue #714, opened 2020-01-12
flake8


dwalton76:
For grins...using this config file
```
dwalton@laptop:ev3dev-lang-python$ cat .flake8.cfg 
[flake8]
max-line-length = 120 
max-complexity = 40
ignore =
    E203
exclude =
    .git
    ./ev3dev2/auto.py
    ./tests/fake-sys/
dwalton@laptop:ev3dev-lang-python$
```

here are the issues flake8 says we should clean up
```
dwalton@laptop:ev3dev-lang-python$ python3 -m flake8 --config=.flake8.cfg .
./git_version.py:2:10: E401 multiple imports on one line
./git_version.py:15:1: E302 expected 2 blank lines, found 1
./git_version.py:22:5: E722 do not use bare 'except'
./git_version.py:31:5: E722 do not use bare 'except'
./ev3dev2/__init__.py:50:1: E402 module level import not at top of file
./ev3dev2/__init__.py:51:1: E402 module level import not at top of file
./ev3dev2/__init__.py:52:1: E402 module level import not at top of file
./ev3dev2/__init__.py:53:1: E402 module level import not at top of file
./ev3dev2/__init__.py:54:1: E402 module level import not at top of file
./ev3dev2/__init__.py:55:1: E402 module level import not at top of file
./ev3dev2/__init__.py:56:1: E402 module level import not at top of file
./ev3dev2/__init__.py:128:9: E722 do not use bare 'except'
./ev3dev2/power.py:31:1: E402 module level import not at top of file
./ev3dev2/sound.py:31:1: E402 module level import not at top of file
./ev3dev2/sound.py:32:1: E402 module level import not at top of file
./ev3dev2/sound.py:33:1: E402 module level import not at top of file
./ev3dev2/sound.py:34:1: E402 module level import not at top of file
./ev3dev2/sound.py:138:43: F841 local variable 'n' is assigned to but never used
./ev3dev2/sound.py:169:121: E501 line too long (178 > 120 characters)
./ev3dev2/sound.py:209:121: E501 line too long (269 > 120 characters)
./ev3dev2/sound.py:214:121: E501 line too long (178 > 120 characters)
./ev3dev2/sound.py:226:121: E501 line too long (178 > 120 characters)
./ev3dev2/sound.py:261:121: E501 line too long (158 > 120 characters)
./ev3dev2/sound.py:291:121: E501 line too long (158 > 120 characters)
[...]

PR #723, opened 2020-02-09
Flake8

dwalton76:
This corrects all warning/errors found by flake8

### DIFF
--- a/.flake8.cfg
+++ b/.flake8.cfg
@@ -7,3 +7,4 @@ exclude =
     .git
     ./ev3dev2/auto.py
     ./tests/fake-sys/
+    ./venv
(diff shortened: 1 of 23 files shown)


dwalton76:
For issue #714

WasabiFan:
This seems generally OK, but I think a lot of this functionality should be tested if we're making these changes, because some of it could have introduced bugs.

dwalton76:
@wasabifan I’ll do some testing tonight and then merge. I looked through the patch again and all of the changes look pretty trivial so I think after some basic testing we should be good.

dwalton76:
I did some testing...all seems fine.

PR #731, opened 2020-03-23
Create "common issues" page and include BrickPi info

### LINKED ISSUE #728: BrickPi3 example
- **ev3dev version:** PASTE THE OUTPUT OF `uname -r` HERE
- **ev3dev-lang-python version:** INSERT ALL VERSIONS GIVEN BY `dpkg-query -l {python3,micropython}-ev3dev*` HERE

We get a fairly steady stream of people trying to use BrickPi3 and failing because it cannot automatically detect sensors and motors. It would be super-helpful if the docs here had an example that people could copy and paste to get started.

Despite the constant questions, I still cannot find any known working example in any of the ev3dev GithHub issues.

WasabiFan:
I figured the FAQ was growing to include two different types of content so I split out the errors in a place that we can include lots of searchable error messages and such.

Fixes #728


docs/common-issues.rst
```diff
@@ -0,0 +1,31 @@
...
+`here <https://www.jetbrains.com/help/pycharm/2016.2/configuring-line-separators.html>`_.
+In Visual Studio Code, there is an option in the lower-right corner.
+Most other editors have similar options; there may be an option for it in the
+status bar at the bottom of the window or in the menu bar at the top.
+
+
+``Exception: Unsupported platform 'None'``
+------------------------------------------
+
+You probably forgot to `update config.txt`_.
```
dlech: It would be good to mention that this only applies to Raspberry Pi.

dlech: On BrickrPi3, there is also the common issue of "my sensors don't work". Solution is to copy https://github.com/ev3dev/ev3dev-lang-python-demo/blob/stretch/platform/brickpi3-motor-and-sensor.py

docs/common-issues.rst
```diff
@@ -0,0 +1,31 @@
+Common Issues with ev3dev-lang-python
+=====================================
+
+``/usr/bin/env: 'python3\r': No such file or directory``
```
dlech, line 15: Another issue (possibly more common than this one) is "no module named ev3dev2". This is usually caused by users running the program on their computer instead of the EV3. 

e.g. https://github.com/ev3dev/ev3dev-lang-python/issues/539#issuecomment-435684710

docs/common-issues.rst
```diff
@@ -0,0 +1,31 @@
...
+
+.. code:: shell
+
+    sed -i 's/\r//g' <file>
+
+This will fix it for the copy of the file on the brick, but if you plan to edit
+it again from Windows, you should configure your editor to use Unix-style
+line endings (LF--line-feed). For PyCharm, you can find a guide on doing this
+`here <https://www.jetbrains.com/help/pycharm/2016.2/configuring-line-separators.html>`_.
+In Visual Studio Code, there is an option in the lower-right corner.
```
dlech: I think most users are using VS Code these days, so maybe list this one first with a screenshot?

WasabiFan:
Updated. Preview: https://ev3dev-lang.readthedocs.io/projects/python-ev3dev/en/common-issues/common-issues.html

WasabiFan:
Note that my instructions for debugging in vscode somewhat depend on https://github.com/ev3dev/vscode-ev3dev-browser/commit/ee1975179db2d94ecc67ea32968dff25ab110322.

PR #732, opened 2020-03-26
Property access benchmarking: timeit-based shell script

WasabiFan:
Here's my take on a reproducible benchmark for property accesses. What do others think about this? I picked a few representative properties (mostly arbitrarily).

My hope is that we can then implement changes to the property access code and make reasonable decisions regarding perf regressions.

### DIFF
--- a/utils/benchmarking/benchmark-results.md
+++ b/utils/benchmarking/benchmark-results.md
@@ -1,0 +1,45 @@
+# Historical I/O benchmark results
+
+Procedure: Use an EV3. Disconnect all peripherals, including Wi-Fi. Connect one large EV3 motor. Cold-boot the EV3 and
+leave it for around five minutes to make sure all services have loaded. Execute `io-benchmark.sh` from the Brickman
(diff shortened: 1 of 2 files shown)


utils/benchmarking/benchmark-results.md
```diff
@@ -0,0 +1,45 @@
...
+Using interpreter micropython; found ev3dev2 version 2.1.0 from /usr/lib/micropython/ev3dev2
+Motor read address:
+1000 loops, best of 3: 1.21 msec per loop
+Motor read speed_sp:
+1000 loops, best of 3: 1.17 msec per loop
+Motor read count_per_rot (cached):
+10000 loops, best of 3: 144 usec per loop
+Motor write speed_sp:
+1000 loops, best of 3: 1.08 msec per loop
+```
```
dwalton76, line 45: This looks pretty cool.  I think some newlines in the output would help make it a little more readable but that is minor.

LGTM

